PR #40, opened 2026-04-23
Add Codex CLI observability

anshss:
## Summary

- Adds a `/v1/logs` OTLP log ingest endpoint so Codex CLI's built-in OTel exporter sends telemetry directly to OCW
- Converts five Codex event types (`codex.sse_event`, `codex.user_prompt`, `codex.tool_decision`, `codex.tool_result`, `codex.api_request`) into normalized spans for cost tracking, drift detection, and alerting
- Adds `CodexEvents` semconv class with all Codex attribute names
- `ocw onboard --codex` now writes the ingest secret to the correct config file (whichever OCW server is already using), preventing secret mismatches
- `ocw serve` writes `~/.local/share/ocw/server.state` at startup so `onboard --codex` finds the right config from any project directory
- Adds 14 new unit tests covering all Codex converters and edge-case parse behavior

## Bugs fixed

| # | Root cause | Fix |
|---|-----------|-----|
| 1 | Codex puts event name in `attrs["event.name"]`, not the log body — all events were silently skipped | Fall back to `attrs.get("event.name")` when body is empty |
| 2 | SSE event kind filter used `"completion"` but Codex 0.123.0 uses `"response.completed"` — no token spans emitted | Updated filter string |
| 3 | `ocw onboard --codex` always wrote to `~/.config/ocw/config.toml` regardless of which config the running server loaded — caused persistent 401s | `ocw serve` now writes a state file; `onboard --codex` reads it first |
| 4 | Codex sets `timeUnixNano=0` for all records; real timestamp is in `attrs["event.timestamp"]` — spans stored with 1970-01-01, invisible to date filters | Parse ISO-8601 fallback when `timeUnixNano == 0` |
| 5 | Unused `import re` + f-string-without-placeholder in `cmd_onboard.py` | Removed both |
| 6 | Non-log OTLP signals (`resourceSpans`, `resourceMetrics`) posted to `/v1/logs` by Codex's SDK returned 400 — noisy logs | Return 200 with 0 ingested instead |

## Known limitation: all Codex sessions appear under `codex_exec`

Codex CLI hardcodes all OTel resource attributes in the binary (`service.name: codex_exec`, `service.version`, `telemetry.sdk.language: rust`, etc.) and ignores the `[otel.resource]` section in `~/.codex/config.toml`. There is no project path or directory identifier in the telemetry. Each Codex run creates a distinct trace (unique `conversation.id`), but all traces show `codex_exec` as the agent regardless of which project directory they came from.

## Test plan

- [x] `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` — 372 passed
- [x] `ruff check ocw/` — clean
- [x] `mypy ocw/` — clean
- [x] End-to-end: `env -u OTEL_EXPORTER_OTLP_HEADERS codex exec "echo hello" -s danger-full-access` → `ocw traces` shows new trace with 6 spans, correct provider/model/token counts
- [x] Multi-project: `ocw onboard --codex --force` from a different project directory correctly syncs the running server's secret via state file

## Usage

```bash
# Point Codex at a running OCW server (run once per machine, or after server restart)
ocw onboard --codex --force

# Codex telemetry now flows into OCW from any project
codex "write a hello world script"
ocw traces
ocw cost
```


anilmurty (approved):
Solid PR - thanks for working on this. We'll just have to keep an eye out and make sure we adjust the regex if OpenAI changes things. LGTM otherwise!

PR #79, opened 2026-06-08
pricing: support a user override file (TJ_PRICING_FILE / ~/.config/tj/pricing.toml)

kelter-antunes:
## Summary

Today `pricing/models.toml` is the only source of rates, and it ships **inside the installed package**. So when a model is missing or a rate is stale, the only fix for an end user is to edit the file under `site-packages/tokenjam/pricing/models.toml` — which `pip install --upgrade tokenjam` then silently overwrites. Contributing the fix upstream (PR #78 does exactly that for the Opus line) is the right long-term answer, but users still need a local, upgrade-safe escape hatch in the meantime.

This adds an optional **user override file** that is merged over the packaged table, resolved highest-priority-first:

1. the path in the **`TJ_PRICING_FILE`** env var, if set
2. **`~/.config/tj/pricing.toml`**, if it exists

The merge is per `provider`/`model`, so an override can both **correct** a packaged rate and **add** a model the package doesn't ship. Same TOML schema as `models.toml`.

```toml
# ~/.config/tj/pricing.toml
[anthropic.claude-opus-4-8]
input_per_mtok = 5.00
output_per_mtok = 25.00
cache_read_per_mtok = 0.50
cache_write_per_mtok = 6.25
```

## Behavior / safety

- A missing `TJ_PRICING_FILE` path or a malformed override **logs a warning and falls back** to packaged rates — it never breaks cost calculation.
- `Path.home()` is resolved at call time (not import) so the lookup honors the current environment.
- Pure stdlib; lives in `tokenjam/core/` (no CLI/HTTP imports per `AGENTS.md`); keeps the existing `load_pricing_table()` signature and `lru_cache`. The `get_rates` date-suffix fallback is unaffected.

## Test plan

- [x] Standalone harness exercising all paths passes: no-override → packaged; env override add/override/untouched; default `~/.config/tj/pricing.toml`; missing-env → warn+fallback; malformed → warn+fallback
- [x] `tokenjam/core/pricing.py` and the new test compile (`py_compile`)
- [x] `pytest tests/unit/test_pricing_override.py` — 7 passed (verified locally, Python 3.13; see verification comment)
- [x] `ruff check tokenjam/` ✅ / `mypy tokenjam/` ✅ (verified locally)

Adds `tests/unit/test_pricing_override.py`. Independent of #78 (different files), so the two can merge in either order.

Note: this is a small change to `core/` rather than a framework integration, so per `CONTRIBUTING.md` I went straight to a PR — happy to convert to an issue-first discussion if you'd prefer to align on the approach (env var name, default path) first.

### DIFF
--- a/tests/unit/test_pricing_override.py
+++ b/tests/unit/test_pricing_override.py
@@ -1,0 +1,114 @@
+"""Unit tests for the user pricing override file in tokenjam.core.pricing.
+
+Covers the TJ_PRICING_FILE env var, the default ~/.config/tj/pricing.toml
+path, merge semantics over the packaged table, and graceful fallback when
+the override is missing or malformed.
+"""
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+
+import pytest
+
+from tokenjam.core import pricing
+
+
+@pytest.fixture(autouse=True)
+def _isolate_pricing(monkeypatch, tmp_path):
+    """Clear the lru_cache around every test and point HOME at an empty dir
+    so a real ~/.config/tj/pricing.toml on the dev machine can't leak in."""
+    monkeypatch.delenv(pricing.USER_PRICING_ENV, raising=False)
+    monkeypatch.setenv("HOME", str(tmp_path))          # POSIX
+    monkeypatch.setenv("USERPROFILE", str(tmp_path))   # Windows Path.home()
+    pricing.load_pricing_table.cache_clear()
+    yield
+    pricing.load_pricing_table.cache_clear()
+
+
+def _write(path: Path, body: str) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(body, encoding="utf-8")
+
+
+def test_no_override_uses_packaged_rates():
+    rates = pricing.get_rates("anthropic", "claude-haiku-4-5")
+    assert rates is not None
+    assert rates.input_per_mtok == 0.80
+
+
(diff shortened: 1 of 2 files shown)


kelter-antunes:
**Local CI verification (ruff + mypy + `pytest tests/unit tests/synthetic tests/agents tests/integration`), Python 3.13.6, editable `pip install -e ".[dev,mcp]"`:**

Green as-is, no changes needed.

- `ruff check tokenjam/` ✅
- `mypy tokenjam/` ✅ (no issues in 104 files)
- `pytest tests/unit/test_pricing_override.py -v` ✅ **7 passed** (add / override / untouched merge, default `~/.config/tj/pricing.toml` path, missing-file warn+fallback, malformed-file warn+fallback)
- full CI test set ✅ **582 passed**
- Local merge of this PR + #78 over `main`: clean (disjoint files), ruff ✅, mypy ✅, **584 passed** ✅

`tomllib.TOMLDecodeError` resolves fine on the import path (`tomllib` on 3.11+, `tomli` alias on 3.10); mypy is clean. The autouse fixture's `cache_clear()` + `HOME`/`USERPROFILE` reset correctly isolates each case.

anilmurty:
Thanks so much for the contribution and for the careful CONTRIBUTING.md note about going PR-first vs. issue-first; the change is small and self-contained enough that PR-first was the right call here.

anilmurty (approved):

PR #90, opened 2026-06-14
cost: charge cache-only spans instead of dropping them as no-ops

sjhddh:
## What

A prompt-cache *hit* — a span with 0 new input/output tokens but non-zero cache-read tokens — bills the cache-read rate, so it costs real money. But both `calculate_cost()` and `CostEngine.process_span()` short-circuited on input/output counts alone:

```python
if input_tokens == 0 and output_tokens == 0:
    return 0.0   # ...and in process_span: return
```

So a cache-only span was dropped as a no-op and costed at $0, quietly under-reporting spend. That's a bit of an own-goal for a cost-observability tool — the more effectively you cache, the more cost goes missing.

## Fix

Widen both early-return guards to fire only when *all* token counts are zero. In `process_span` the `cache_read_tokens` read is hoisted above the guard so it can participate in the check. `calculate_cost`'s arithmetic already summed all four token types correctly — only the guard was wrong, so no math changes.

The pure all-zero no-op (genuinely empty spans) is preserved.

## Tests

- `test_calculate_cost_cache_read_only` — pure cache-read span is priced (0.08 for 1M @ claude-haiku-4-5).
- `test_calculate_cost_with_cache_write_tokens` — updated: this previously asserted `0.0` (codifying the bug); now asserts `1.0`, comment corrected.
- `test_cost_engine_costs_cache_only_span` — `process_span` records the span cost.
- `test_cost_engine_cache_only_span_updates_session_total` — and accumulates it into the session total (the under-reporting was at the session level too).

Full suite passes locally (`pytest tests/unit tests/synthetic tests/agents tests/integration`); `ruff check` and `mypy` clean on the changed files. CHANGELOG updated under Unreleased → Fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,9 @@ This release pivots TokenJam toward cost-optimization for autonomous agents. Fou
 - **`tj optimize` plan-tier-aware rendering.** When every session in the window has `plan_tier = 'unknown'`, dollar figures are suppressed and a header note explains why. Mixed / partial-unknown windows render normally with an advisory note.
 - **MCP `get_optimize_report` tool.** Now accepts `findings: list[str]` (was `only: str`). Docstring surfaces for both API-billing and subscription-plan-efficiency phrasings.
 
+### Fixed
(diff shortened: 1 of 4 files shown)


anilmurty (approved):
Real cost-correctness bug, well-isolated fix, and the right call to update the test that was codifying the bug rather than silently overwriting it.

anilmurty:
Merging this now - thanks so much for finding this and fixing it @sjhddh !

anilmurty:
One follow-up: I'm filing a separate issue (no rush): the process_span tests cover cache-read but not cache-write. Both paths are now correct, but adding a cache-write process_span test would be great for completeness.

PR #92, opened 2026-06-14
fix(otlp): cost cache-creation tokens on the live ingest path

sjhddh:
## What

Follow-up to #90. In that PR's review, the request was:

> the `process_span` tests cover cache-read but not cache-write. Both should be covered.

Digging into that turned up a real gap behind the missing test: **cache-creation (cache-write) tokens are dropped on the live ingest path and never costed.**

The plumbing was half-wired:
- the SDK integrations emit `gen_ai.usage.cache_creation_tokens`,
- the pricing table carries a `cache_write_per_mtok` rate, and
- `calculate_cost` already accepts and prices `cache_write_tokens`,

but the two live-path readers — `parse_otlp_span` (OTLP file/HTTP ingest) and `convert_otel_span` (in-process OTel exporter) — only read cache-**read** tokens. So cache-creation tokens never reached `CostEngine.process_span`, and their (higher-rate) cost was silently never charged. Only the `backfill` path costed cache-write; the live path under-reported spend.

## Change

- Add `cache_write_tokens` to `NormalizedSpan` (kept separate from `cache_tokens`/cache-read since the two bill at different rates).
- Populate it in both live parsers (`parse_otlp_span`, `convert_otel_span`) from `CACHE_CREATE_TOKENS`.
- Read it in `process_span` and pass it to `calculate_cost`; widen the no-op early-return guard so a cache-write-only span is no longer dropped.

## Tests

- `process_span` cache-write coverage (the originally-requested gap): span cost **and** session total, cache-write-only and cache-read+write together, at the haiku rates (`cache_read=0.08`, `cache_write=1.00` per MTok).
- A guard test ensuring a pre-priced span (cost already set upstream) is recomputed but **not** double-counted into the session total.
- Parser-level tests asserting both readers populate `cache_write_tokens`.

Full suite green locally (584 passed); `ruff` clean.

## Out of scope (noted as follow-ups)

- The Claude Code logs adapter (`api/routes/logs.py`) and the Langfuse/Helicone backfill adapters are unchanged here — they already persist `cost_usd` directly, so they don't under-report, but threading the raw `cache_write_tokens` count through them for consistency is a separate cleanup.
- `cache_write_tokens` is used for costing but not yet persisted as its own DB column (only the resulting `cost_usd` is). Adding the column is a follow-up if per-token-class reporting is wanted.


anilmurty (approved):

anilmurty:
merging and noticed a couple more things while reviewing. filed #93 and #94

Issue #106, opened 2026-06-16
v0.3.5 pre-release playbook findings


anilmurty:
Bugs and rough edges surfaced while running `tests/manual-pre-release-testing.md` against the v0.3.5 candidate on `main`. Bundling so we can close them in one or two PRs before cutting the release.

## 1. Stale version strings (UI + package) — **blocker for v0.3.5**

Two hardcoded version strings have drifted from `pyproject.toml`:

- `tokenjam/ui/index.html:532` → `v0.1.4` (rendered in the sidebar footer of the web UI)
- `tokenjam/__init__.py:1` → `__version__ = \"0.1.9\"`

Fix: source from `importlib.metadata.version(\"tokenjam\")` in `__init__.py`, expose via a small `/api/v1/version` endpoint, fetch from the UI on init, render in footer. Same-origin so the UI stays fully offline-capable (Critical Rule 18).

## 2. `/health` endpoint missing

`curl http://127.0.0.1:7391/health` returns 404. The real liveness probe is `/api/v1/status`, which works, but `/health` is the conventional name for a process-liveness check and is what most uptime-monitoring tools probe by default. Add as a thin alias returning `{\"status\": \"ok\"}`.

## 3. Tokenmaxx multiplier doesn't render from project-local cwd

When run from a directory with a project-local `.tj/config.toml` that has no `[budget.<provider>]` section, `tj tokenmaxx` silently falls back to api-pricing framing even if the **global** `~/.config/tj/config.toml` has `plan = \"max_5x\"` configured. Running from `\$HOME` (where global wins) renders the multiplier line correctly (\"33.9× your Max 5x plan cost\").

Two possible fixes:
- Merge plan-tier from global config into project-local config-as-overlay during discovery.
- Print a one-line hint when project-local config exists but has no `[budget]` section: \"No plan configured at this scope — run `tj onboard --reconfigure` here or `cd ~` to see plan-aware framing.\"

The first is more correct; the second is one line of code.

## 4. Cosmetic Rich markup bug — already filed

#105 covers the `[capture]` markup-escape drop in `tj report --trim` hint. Listed here for completeness, not re-opening.

---

Found during the v0.3.5 pre-release playbook run on 2026-06-15. #1 must be fixed before cutting v0.3.5; #2 and #3 are low-risk follow-ups that can ship with the release or in v0.3.6.

PR #117, opened 2026-06-17
#110: Extract plan-tier framing into core/framing.py

### LINKED ISSUE #110: [feature] Extract plan-tier-aware rendering into core/framing.py
**Type:** Refactor / foundation. No user-visible change. Parallelizable with B and C.

## Background

`tj cost`, `tj optimize`, and `tj tokenmaxx` each apply plan-tier-aware rules when rendering dollar figures (subscription users get token-share framing; api users get raw dollars; unknown plans get a qualifier hint). Today this logic is duplicated across `cmd_cost.py`, `cmd_optimize.py`, and `cmd_tokenmaxx.py`, and the REST API doesn't apply it at all — it returns raw numbers and the CLI re-derives the framing.

The local web UI needs to render the same numbers as the CLI without re-implementing the rules in JavaScript. The cleanest path: extract the framing logic into a single shared module that both the CLI commands and the REST API endpoints consume.

## Acceptance criteria

1. New module `tokenjam/core/framing.py` exports:
   - `compute_framing(config, window_summary, by_provider_breakdown) -> Framing` — returns a `Framing` dataclass with `pricing_mode`, `plan_tier`, `plan_label`, `plan_monthly_usd`, `subscription_share_pct`, `api_share_pct`, `display_rule`, `qualifier_text`.
   - `render_dollar(value, framing) -> str` — returns the appropriately framed dollar string for the given context (`"$148"`, `"12.4% of cycle"`, `"—"`).
   - `render_savings(value_usd, value_tokens, framing) -> str` — same shape, for savings tiles.
2. `cmd_cost.py`, `cmd_optimize.py`, `cmd_tokenmaxx.py` import from `core/framing.py` and remove their local copies. CLI output must be byte-identical before vs after.
3. `/api/v1/cost`, `/api/v1/cost_compare`, `/api/v1/optimize`, `/api/v1/budget` responses include a top-level `framing` object with the same fields as `Framing`. Existing callers that don't read the `framing` field continue to work.
4. New unit tests in `tests/unit/test_framing.py` cover: each `pricing_mode` value, mixed subscription+api window, unknown plan, empty data.
5. Snapshot or golden tests confirming CLI output didn't change.
6. `mypy` and `ruff` clean.

## Scope guardrails

- Do not change `SessionRecord.plan_tier` or `pricing_mode` derivation rules. The module reads them, it doesn't define them.
- Do not change CLI behavior. Output must be identical.
- Do not introduce a new config field. Plan-tier still comes from `[budget.<provider>] plan = "..."` only.

## Files likely touched

- `tokenjam/core/framing.py` (new)
- `tokenjam/cli/cmd_cost.py`, `cmd_optimize.py`, `cmd_tokenmaxx.py` (consume the module)
- `tokenjam/api/routes/cost.py`, `cost_compare.py`, `optimize.py`, `budget.py` (emit the framing block)
- `tests/unit/test_framing.py` (new)

anilmurty:
Closes #110.

Foundation for the Lens UI work: a single source of truth for plan-tier-aware
rendering, consumed by both the CLI and the REST API so the web UI can render
identical dollar/token figures without re-deriving the rules in JavaScript.

## What changed

**New module `tokenjam/core/framing.py`**
- `Framing` dataclass (`pricing_mode`, `plan_tier`, `plan_label`,
  `plan_monthly_usd`, `subscription_share_pct`, `api_share_pct`,
  `display_rule`, `qualifier_text`, plus window totals).
- `compute_framing(config, window_summary, by_provider_breakdown)`.
- `render_dollar(value, framing)` and `render_savings(value_usd, value_tokens, framing)`.
- Shared helpers `pricing_mode_for` / `dominant_plan` / `config_declared_plan`
  (with the #106 global-config fallback) / `plan_tier_mix`.

**CLI migration (output unchanged)**
- `cmd_optimize.py` and `cmd_tokenmaxx.py` import the helpers from the shared
  module and drop their local copies. `cmd_cost.py` did not apply framing
  before and still doesn't (its rendering is untouched).

**API**
- `/api/v1/cost`, `/api/v1/cost/compare`, `/api/v1/optimize`, `/api/v1/budget`
  now return a top-level `framing` block. Callers that ignore it keep working.

## Tests
- `tests/unit/test_framing.py` — one path per `pricing_mode`
  (api / subscription / local / unknown), mixed windows, empty data, and both
  render helpers.
- `tests/integration/test_api.py` — framing block present on the endpoints and
  reflects a configured subscription plan.
- Full suite: 665 passed; `mypy` and `ruff` clean.

## Scope guardrails honoured
- No change to `SessionRecord.plan_tier` / `pricing_mode` derivation.
- No new config field.
- CLI behaviour unchanged.

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -238,6 +238,54 @@ async def test_get_cost_returns_aggregated_rows(client):
     assert "total_cost_usd" in data
 
 
+_FRAMING_KEYS = {
+    "pricing_mode", "plan_tier", "plan_label", "plan_monthly_usd",
+    "subscription_share_pct", "api_share_pct", "display_rule", "qualifier_text",
+}
+
+
+async def test_cost_response_includes_framing_block(client):
+    await _ingest_sample_span(client)
+    resp = await client.get("/api/v1/cost")
+    assert resp.status_code == 200
+    framing = resp.json()["framing"]
+    assert _FRAMING_KEYS <= set(framing)
+
+
+async def test_optimize_response_includes_framing_block(client):
+    await _ingest_sample_span(client)
+    resp = await client.get("/api/v1/optimize?since=30d")
+    assert resp.status_code == 200
+    data = resp.json()
+    if data.get("error") == "no_data":
+        pytest.skip("no spans landed for optimize in this fixture")
+    assert _FRAMING_KEYS <= set(data["framing"])
+
+
+async def test_budget_framing_reflects_configured_subscription_plan(db):
+    """The budget surface has no window, so framing falls back to the
+    declared plan in config (#110)."""
+    from tokenjam.core.config import ProviderBudget
+
+    cfg = TjConfig(
+        version="1",
+        security=SecurityConfig(ingest_secret=INGEST_SECRET),
+        api=ApiConfig(auth=ApiAuthConfig(enabled=False)),
+    )
+    cfg.budgets["anthropic"] = ProviderBudget(plan="max_5x")
+    pipeline = IngestPipeline(db=db, config=cfg)
+    app = create_app(config=cfg, db=db, ingest_pipeline=pipeline)
+    transport = httpx.ASGITransport(app=app)
+    async with httpx.AsyncClient(transport=transport, base_url="http://test") as c:
+        resp = await c.get("/api/v1/budget")
+    assert resp.status_code == 200
+    framing = resp.json()["framing"]
+    assert framing["pricing_mode"] == "subscription"
+    assert framing["plan_tier"] == "max_5x"
+    assert framing["plan_label"] == "Max 5x plan"
+    assert framing["display_rule"] == "suppress_dollars_for_subscription_share"
+
+
 async def test_get_alerts_returns_list(client):
     resp = await client.get("/api/v1/alerts")
     assert resp.status_code == 200
--- a/tests/unit/test_cmd_tokenmaxx.py
+++ b/tests/unit/test_cmd_tokenmaxx.py
@@ -1,14 +1,33 @@
 """Unit tests for `tj tokenmaxx`."""
 from __future__ import annotations
 
+import pytest
+
 from tokenjam.cli.cmd_tokenmaxx import (
     Tier,
     _TIERS,
     _classify,
-    _config_declared_plan,
-    _plan_label_and_fee,
 )
 from tokenjam.core.config import ProviderBudget, TjConfig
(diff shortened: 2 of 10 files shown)


anilmurty (commented):

**Acceptance criteria — all met:**

1. ✓ `core/framing.py` exports `compute_framing`, `render_dollar`, `render_savings` with the spec'd signatures
2. ✓ CLI byte-identical — `cmd_optimize` / `cmd_tokenmaxx` use the new module as a pure drop-in; helpers (including the #106 global-config fallback) are unchanged
3. ✓ All 4 endpoints emit the `framing` block with the exact field names from the spec
4. ✓ Tests cover api / subscription / local / unknown, mixed window, empty data, and both render helpers
5. ✓ No changes to `models.py` / `semconv.py` (pricing-mode derivation untouched)
6. ✓ No new config keys

**Nits / discussion points:**

- **F-string in SQL** at `tokenjam/core/framing.py:325` — `f"SELECT … FROM sessions{where}"`. Values are properly parameterized via `$N`; only the WHERE skeleton is interpolated from a fixed set of hardcoded clauses, so behavior is correct and it matches the original `_plan_tier_mix` shape from `cmd_optimize.py`. Technically still touches CLAUDE.md Critical Rule 7 ("Parameterised SQL only — never use f-string SQL"). Minor — would be nicer as a constant lookup or two-branch conditional. Not a blocker.

- **`cmd_cost.py` follow-up worth filing.** The PR correctly identifies that `cmd_cost.py` had no framing logic to extract pre-PR (only a deferral comment at `cmd_cost.py:151-156`). Legitimate scope cut. But now that `core/framing.py` exists, `tj cost`'s `_render_diff` still renders raw dollars regardless of plan tier — subscription users see dollar figures they shouldn't. Worth a follow-up issue rather than expanding this one.

**No blockers.** Dependency rule clean (no `cli`/`api` imports from `core`), no `datetime.now()` usage, no debug prints or TODOs, module well-organized (constants → helpers → dataclasses → core function → renderers → DB helper).

CI green on 3.10 / 3.11 / 3.12 + TS. Good to merge.

anilmurty:
Thanks for the review! Both nits addressed:

1. **F-string SQL (Rule 7)** — fixed in 24c9795. `plan_tier_mix` now builds the query by concatenating static fragments; every value still binds through a `$N` parameter, nothing interpolated. Rebased onto the updated branch.
2. **`cmd_cost` framing gap** — filed as #120 (follow-up), keeping this PR a pure refactor as you suggested.

CI should re-run on the new commit.

PR #119, opened 2026-06-17
#111: Add estimated_recoverable_usd to every savings analyzer

### LINKED ISSUE #111: [feature] Add estimated_recoverable_usd to every optimize analyzer
**Type:** Feature / backend. Parallelizable with #110 and C.

## Background

Today only `DowngradeFinding` carries a dollar field (`monthly_savings_usd`). The other analyzers — `cache`, `script`, `trim`, `cache-recommend` — return clusters, ratios, and significance scores but no consistent dollar figure. We want every analyzer that produces a usable recommendation to also produce an estimated-recoverable amount, so it can be summarized in one place.

The estimate is **heuristic** and must be framed as such. The existing honesty discipline carries forward — every number is "estimated recoverable," never "saves you," and every analyzer keeps its existing caveat language.

## Contract

Every analyzer's result dataclass gains:

```python
estimated_recoverable_usd: float | None = None     # api-billed framing
estimated_recoverable_tokens: int | None = None    # subscription / local framing
estimate_basis: str = ""                            # one-sentence heuristic explanation
confidence: Literal["heuristic"] = "heuristic"      # constant for now
```

`None` means "no estimate available for this finding state" (e.g. Trim with capture off).

## Per-analyzer heuristic

- **`downsize`** (`model_downgrade.py`) — already has `monthly_savings_usd`. Alias to `estimated_recoverable_usd`. Populate `estimated_recoverable_tokens` from the existing `candidate_tokens` field. `estimate_basis = "candidate sessions routed to a cheaper model — structural fit only, no quality validation"`.

- **`cache`** (`cache_efficacy.py`) — for each `(provider, model)` row, gap to an 80% efficacy ceiling × cacheable input tokens × (input rate − cache-read rate). Sum across rows. Skip rows with no `cache_read_per_mtok` in `pricing/models.toml` (those models have no caching dimension). `estimate_basis = "gap between current cache-read efficacy and 80% ceiling at the input-vs-cache rate delta"`.

- **`script`** (`workflow_restructure.py`) — for each `(tool_name, arg_shape)` cluster, sum `session.total_cost_usd` across sessions in the cluster. `estimate_basis = "total cost of sessions matching a deterministic call-pattern — replacing the cluster with a script eliminates the LLM call entirely"`.

- **`trim`** (`prompt_bloat.py`) — sum of low-significance tokens × window-average input rate. When `[capture] prompts = false` or `llmlingua` missing: `estimated_recoverable_usd = None`. `estimate_basis = "low-significance tokens predicted by LLMLingua-2 × input rate — review before editing prompts"`.

- **`cache-recommend`** and **`budget-projection`** — do **not** emit `estimated_recoverable_usd`. Both keep their fields as-is. They are not savings analyzers.

## Acceptance criteria

1. Each analyzer result dataclass listed above gains the new fields.
2. Existing CLI rendering output is unchanged (the new fields are read by the API/UI, not by the existing CLI rendering paths).
3. `/api/v1/optimize` response includes the new fields for each finding.
4. Per-analyzer unit tests assert: realistic input produces a reasonable USD estimate (within a hand-calculated bound); `None` is returned in not-ready states; `estimate_basis` is non-empty.
5. No analyzer's existing caveat language is removed or weakened.

## Scope guardrails

- Do not add a new analyzer.
- Do not change which analyzers run by default.
- Do not change the registry pattern. Self-registration via `@register("...")` stays.
- Do not change CLI output format.

## Files likely touched

- `tokenjam/core/optimize/types.py` (new fields on dataclasses)
- `tokenjam/core/optimize/analyzers/model_downgrade.py`
- `tokenjam/core/optimize/analyzers/cache_efficacy.py`
- `tokenjam/core/optimize/analyzers/workflow_restructure.py`
- `tokenjam/core/optimize/analyzers/prompt_bloat.py`
- `tokenjam/api/routes/optimize.py` (response serialization)
- `tests/unit/test_optimize_*.py` (new test cases per analyzer)

anilmurty:
Closes #111.

Every analyzer that produces a usable recommendation now also produces a
consistent, **heuristic** estimated-recoverable figure, so the Lens Overview can
summarize "recoverable waste" in one band. The honesty discipline carries
forward — every number is "estimated recoverable", never "saves you", and each
analyzer keeps its existing caveat language.

## Contract (added to each savings analyzer's result dataclass)
- `estimated_recoverable_usd: float | None` — api-billed framing
- `estimated_recoverable_tokens: int | None` — subscription / local framing
- `estimate_basis: str` — one-line heuristic explanation (the "estimated" tag's hover text)
- `estimate_confidence: str = "heuristic"` — named distinctly from the
  pre-existing structural `confidence` field on the wave-2 findings to avoid a
  semantic clash.

`None` means "no estimate for this finding state" (e.g. trim with capture off).

## Per-analyzer heuristic
- **downsize** — aliases the existing `monthly_savings_usd`; tokens from `candidate_tokens`.
- **cache** — gap to an 80% efficacy ceiling × input tokens × (input − cache-read rate delta), summed across rows. `None` for models with no cache-read rate.
- **script** — full cost (and tokens) of the sessions in each deterministic cluster.
- **trim** — low-significance tokens × window-average input rate. `None` when capture is off or `llmlingua` is missing.
- **cache-recommend / budget-projection** — intentionally emit no figure (not savings analyzers).

## Plumbing
- Fields serialize through `report_to_dict` (so `/api/v1/optimize` carries them) and round-trip through `report_from_dict`.
- CLI rendering is unchanged — the new fields are read by the API/UI, not the existing render paths.

## Tests
`tests/unit/test_optimize_recoverable.py`: per-analyzer USD within a
hand-calculated bound, `None` in not-ready states, non-empty `estimate_basis`,
and a `report_to_dict` / `report_from_dict` round-trip. Full suite green; `mypy`
and `ruff` clean.

## Scope guardrails honoured
No new analyzer; no change to which analyzers run by default; registry pattern
unchanged; CLI output format unchanged.

### DIFF
--- a/tests/unit/test_optimize_recoverable.py
+++ b/tests/unit/test_optimize_recoverable.py
@@ -192,4 +192,5 @@ def test_trim_recoverable_none_when_capture_off(db):
     finding = report.findings["trim"]
     assert finding.enabled is False
     assert finding.estimated_recoverable_usd is None
-    assert finding.estimate_basis == "" or finding.estimate_basis is not None
+    # Not-ready path doesn't set a basis (default empty string).
+    assert finding.estimate_basis == ""
--- a/tests/unit/test_optimize_recoverable.py
+++ b/tests/unit/test_optimize_recoverable.py
@@ -1,0 +1,195 @@
+"""Per-analyzer estimated-recoverable-USD tests (issue #111)."""
+from __future__ import annotations
+
+from datetime import timedelta
+
+import pytest
+
+from tokenjam.core.config import CaptureConfig, TjConfig
+from tokenjam.core.db import InMemoryBackend
+from tokenjam.core.optimize import analyze_model_downgrade, build_report
+from tokenjam.core.optimize.analyzers.cache_efficacy import (
+    CacheEfficacyRow,
+    estimate_cache_recoverable,
+)
+from tokenjam.core.optimize.analyzers.prompt_bloat import estimate_trim_recoverable
+from tokenjam.utils.time_parse import utcnow
+from tests.factories import make_llm_span, make_session, make_tool_span
+
+
+@pytest.fixture
+def db():
+    backend = InMemoryBackend()
+    yield backend
+    backend.close()
+
+
+def _window():
+    return utcnow() - timedelta(days=30), utcnow() + timedelta(hours=1)
+
+
+# --------------------------------------------------------------------------- #
+# downsize — aliases the existing monthly projection
+# --------------------------------------------------------------------------- #
+def _insert_small_opus_session(db, session_id="a"):
+    start = utcnow() - timedelta(days=2)
+    llm = make_llm_span(
+        agent_id="claude-code-x", model="claude-opus-4-7", provider="anthropic",
+        input_tokens=1000, output_tokens=200, cost_usd=0.030,
+        session_id=session_id, start_time=start,
+    )
+    db.insert_span(llm)
+    for _ in range(2):
+        tool = make_tool_span(agent_id="claude-code-x", tool_name="Read",
+                              trace_id=llm.trace_id)
+        tool.session_id = session_id
+        tool.start_time = start
+        db.insert_span(tool)
+
+
+def test_downsize_recoverable_aliases_monthly_savings(db):
+    _insert_small_opus_session(db)
+    since, until = _window()
+    finding = analyze_model_downgrade(db.conn, since, until, None, 30.0)
+    assert finding is not None
+    assert finding.estimated_recoverable_usd == finding.monthly_savings_usd
+    assert finding.estimated_recoverable_usd > 0
+    assert finding.estimated_recoverable_tokens == finding.candidate_tokens
+    assert finding.estimate_basis
+    assert finding.estimate_confidence == "heuristic"
+
+
+# --------------------------------------------------------------------------- #
+# cache — efficacy-gap heuristic (pure helper)
+# --------------------------------------------------------------------------- #
+def test_cache_recoverable_matches_hand_calculation():
+    # opus-4-7: input 5.00/MTok, cache_read 0.50/MTok → delta 4.50.
+    # 1M input, 0 cache → efficacy 0 → gap 0.80 → 800K recoverable tokens.
+    # usd = 800_000/1e6 * 4.50 = 3.60
+    row = CacheEfficacyRow(
+        provider="anthropic", model="claude-opus-4-7",
+        input_tokens=1_000_000, cache_tokens=0, efficacy=0.0,
+        support="full", flagged=True,
+    )
+    usd, tokens = estimate_cache_recoverable([row])
+    assert tokens == 800_000
+    assert usd == pytest.approx(3.60, abs=0.01)
+
+
+def test_cache_recoverable_none_when_no_caching_dimension():
+    # An unknown model falls back to default rates with cache_read = 0 → no
+    # caching dimension → no estimate.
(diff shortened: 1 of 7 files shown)


anilmurty (commented):

**Acceptance criteria — all met:**

1. ✓ `confidence` → `estimate_confidence` rename is justified. Wave-2 findings (`cache_efficacy.py:71`, `prompt_bloat.py:102`, `workflow_restructure.py:102`, `cache_recommend.py:53`) already use `confidence: "structural"` for a different axis (structural vs replay_validated). The new `estimate_confidence: "heuristic"` describes a different axis. Collapsing would have been semantic overloading.
2. ✓ `model_downgrade.py:206` is a literal alias — `estimated_recoverable_usd` and `monthly_savings_usd` share the same computed value.
3. ✓ `cache_efficacy.py:80-115` formula matches spec: `gap = max(0, 0.80 − efficacy) × cacheable input tokens × (input rate − cache-read rate)`. `EFFICACY_CEILING = 0.80` is a module constant (line 63). Skips rows when `cache_read_per_mtok <= 0` (line 98) and when `rate_delta <= 0` (line 101).
4. ✓ `prompt_bloat.py` returns `None` in both not-ready states (capture off at `:287-294`, llmlingua missing at `:301-306`). `_window_avg_input_rate` is token-weighted across the actual model mix — better than spec.
5. ✓ `cache-recommend` and `budget-projection` untouched (correct — not savings analyzers).
6. ✓ CLI render paths don't read the new fields. Round-trip works via `runner.py:250-380` `asdict`/constructor paths.
7. ✓ Tests cover USD bounds, `None` in not-ready states, non-empty basis, round-trip.

**Nits / follow-ups (non-blocking):**

- **`runner.py:42`** — `_utcnow()` uses `datetime.now(tz=timezone.utc)` rather than `tokenjam.utils.time_parse.utcnow()` (CLAUDE.md rule #9). Pre-existing, not introduced by this PR, but the file is touched here — easy to clean up while we're at it.
- **`tests/unit/test_optimize_recoverable.py:195`** — `assert finding.estimate_basis == "" or finding.estimate_basis is not None` is a tautology (`""` is not `None`, so the right operand is always true). Likely meant `== ""` for the not-ready case.
- **`prompt_bloat.py:396`** — `low_sig_tokens = total_bloat / CHARS_PER_TOKEN` where `CHARS_PER_TOKEN = 4`. Reasonable approximation for English prose, but the `estimate_basis` string doesn't disclose the conversion. Worth surfacing in the UI hover text once Lens lands.
- **`workflow_restructure.py:194`** — `cluster_cost` is the *full* historical cost of clustered sessions, not window-normalized. Basis text is honest, but the figure isn't directly comparable to monthly-projected numbers on other tiles. Follow-up to either window-normalize or annotate the time basis explicitly.

CI green on 3.10 / 3.11 / 3.12 + TS. No blockers. Good to merge.

anilmurty:
Thanks for the thorough review! Addressed in 04f3330:

1. **`runner.py` Rule 9** — `_utcnow()` and the `report_from_dict` fallbacks now use `tokenjam.utils.time_parse.utcnow()` instead of `datetime.now(tz=timezone.utc)`.
2. **Tautological assert** — fixed; the not-ready trim path asserts `estimate_basis == ""`.

For the two follow-ups (script estimate is full-historical vs monthly-projected on other tiles + `CHARS_PER_TOKEN` disclosure in hover text), filed #122 so they're tracked for when the Overview band lands — kept out of this PR to stay focused.

Issue #112, opened 2026-06-17
[feature] Vendor uPlot for offline charting + move filter state into the URL


anilmurty:
**Type:** Feature / UI platform. Parallelizable with #110 and #111.

## Background

Two foundational changes needed before any new dashboard surfaces can be built.

**1. Charting library.** The UI is all tables today. We need a small, vendorable chart library to render spend-over-time and other time-series visualizations. The dashboard is offline-first (enforced by `tests/unit/test_ui_offline.py`), so the library must ship in the wheel — no CDN loads at render time.

**2. URL-driven filters.** Today filter state lives in Preact component state. The hash route only carries the screen name (`#/cost`, `#/status`, etc.). To enable cross-screen drill-through ("click an item on screen A, land on screen B already filtered"), filters need to live in the URL via query params.

## Acceptance criteria — uPlot

1. uPlot (https://github.com/leeoniya/uPlot) vendored under `tokenjam/ui/vendor/uplot.js` and `tokenjam/ui/vendor/uplot.css`. Pinned to a specific release; version noted in a comment at the top of the vendored file.
2. Both files included in the wheel build automatically (the existing `tokenjam` package include should cover it; verify).
3. Wired via the existing import flow in `tokenjam/ui/index.html` so charts can be instantiated from the inline `<script>` block.
4. `tests/unit/test_ui_offline.py` continues to pass with no modifications. Specifically: no new external URLs introduced.
5. A small smoke chart rendered somewhere in the existing UI to prove the wiring works (can be removed in the Cost screen work later). Suggest: tiny sparkline at the top of the existing Cost screen showing total cost per day.

## Acceptance criteria — URL state

1. Each existing screen reads its filter state from the hash route + query params, falling back to defaults when params are missing or malformed.
2. Time-window vocabulary matches the CLI: `1h`, `24h`, `7d`, `30d`, `90d`, and explicit ranges `YYYY-MM-DD:YYYY-MM-DD`.
3. Default values are *omitted* from the URL (don't write `since=24h` when 24h is the default).
4. Changing a filter updates the URL via `history.pushState` so the browser back/forward buttons work.
5. A reload preserves filter state.
6. Bad / unparseable values fall back to defaults with a `console.warn`, never throw.
7. The following filters work on the listed screens:
   - **Status**: `agent_id` (single)
   - **Traces**: `agent_id`, `result`, `since`
   - **Cost**: `since`, `group_by`, `agent_id`
[...]

PR #121, opened 2026-06-17
#112: Vendor uPlot for offline charting + move filter state into the URL

anilmurty:
Closes #112.

Two foundations the dashboard surfaces need before #113 / #114.

## Charting (uPlot)
- Vendored **uPlot v1.6.32** — `tokenjam/ui/vendor/uplot.js` (IIFE) + `uplot.css`, pinned with a header comment. Loaded via a plain `<script>` / `<link>` (no CDN, no importmap entry needed since it sets the `uPlot` global).
- Verified both files ship in the wheel (built and inspected `tokenjam/ui/vendor/*` in the `.whl`).
- New `LineChart` component over the global; reads CSS custom properties so it matches the active light/dark theme.
- **Smoke chart**: a "Spend per day" sparkline at the top of the Cost screen, proving the wiring (will be folded into the real Cost chart in #113).

## URL-driven filters
- `getRoute()` parses the hash into `{view, param, params}` where `params` is the query string. Filter state now lives in the URL — the source of truth — so Overview→detail drill-through will round-trip through reload + back/forward.
- `navigate()` writes filters back, **omitting defaults** (`#/cost`, not `#/cost?since=7d`) and pushing a history entry. `readParam()` validates and falls back to defaults with a `console.warn`, never throws.
- Wired per screen:
  - **Status**: `agent_id`
  - **Traces**: `agent_id`, `result` (→ API `status`), `since`
  - **Cost**: `since`, `group_by`, `agent_id`
  - **Alerts**: `severity`, `unread`, `agent_id`, `since`
  - **Drift**: `agent_id`
  - **Budget**: `provider` (preserved + clearable chip; provider-scoped budget content arrives with budget-projection surfacing in #114)
- Window vocab matches the CLI: `1h` / `24h` / `7d` / `30d` / `90d` + `YYYY-MM-DD:YYYY-MM-DD` ranges.

## Tests
- `tests/unit/test_ui_offline.py` still green; vendored files asserted present **and** shipped in the wheel; new check that vendored CSS has no external `url()`.
- Module JS syntax validated with `node --check`; ASGI smoke test confirms `/`, `/ui/vendor/uplot.js`, `/ui/vendor/uplot.css` all serve 200.

## Scope guardrails honoured
No client-side router, no state-management lib, no build step, no TypeScript — built on the existing hash-route handler + importmap.

PR #123, opened 2026-06-17
#113: Cost screen — replace table-only view with a real chart

### LINKED ISSUE #113: [feature] Cost screen — replace the table-only view with a real chart
**Type:** Feature / UI.

**Depends on:** #110 (framing module), #112 (uPlot + URL state).

## Background

The Cost screen today is a single table of daily rows. The data behind it — `tj cost --since 7d` and `tj cost --since 7d --compare previous` — already supports rich period-over-period comparison, but the UI never surfaces it. With uPlot vendored and URL state in place, we can lead with the chart and keep the table below for precise lookup.

## Acceptance criteria

1. Cost screen renders, top to bottom:
   - A spend-over-time line/area chart using uPlot, default window 7d.
   - Window toggle: 24h / 7d / 30d / 90d (URL state via `?since=`).
   - Group-by toggle: by model / by agent / total (URL state via `?group_by=`).
   - A trend chip showing delta vs the previous equivalent period (▲/▼ with %, sourced from `/api/v1/cost_compare`).
   - The existing precise daily table below the chart, unchanged.
2. When `framing.display_rule = "suppress_dollars_for_subscription_share"`, the chart axis labels and totals switch to token-share framing (the framing module from #110 handles the rendering decision).
3. When the window contains a mix of subscription and api-billed activity, the `qualifier_text` from the framing block is shown above the chart.
4. Chart adapts to light / dark theme via CSS custom properties.
5. Empty state: "No spend in this window" rendered cleanly when the API returns zero data.
6. Loading state: low-key shimmer, no full-page loader.
7. Error state: inline error message with a retry button.

## Scope guardrails

- Do not add a new endpoint. Use the existing `/api/v1/cost` and `/api/v1/cost_compare`. If the response shape needs an extra field for the chart (e.g. a daily-bucketed series), extend the existing response, do not create a new route.
- Do not add forecasting beyond a single linear run-rate continuation line on the chart, sourced from spend over the visible window. No moving averages, no seasonality, no anomaly bands.
- Do not regress the existing daily table.

## Files likely touched

- `tokenjam/ui/index.html` (Cost view component)
- `tokenjam/api/routes/cost.py` (possibly extend response with a daily-bucketed series)
- Tests as appropriate

anilmurty:
Closes #113.

**Stacked PR.** Base is `feature/112` (uPlot + URL state); also merges in
`feature/110` (framing) since the Cost chart needs both. The diff therefore
includes the framing module (#110) until #117/#121 merge — review the
`#113`-titled commit for the new work, or re-review after the parents land. I'll
retarget the base to `main` as the foundations merge.

## What changed
- **Chart leads, table follows.** uPlot spend-over-time chart (default 7d), with
  the precise daily table preserved below.
- **Toggles, URL-synced** (#112): window `24h/7d/30d/90d` (`?since=`), group-by
  `total / model / agent` (`?group_by=`). Splits keep the top-5 keys + "other".
- **Trend chip** ▲/▼ % vs the previous equivalent period, from
  `/api/v1/cost/compare`.
- **Run-rate**: a single dashed mean-per-day line + a "~$X over 30 days"
  projection, explicitly labelled *not a forecast* (no moving averages /
  seasonality — within #113 scope).
- **Plan-tier framing** (#110): subscription / local switch the axis + totals to
  token framing and surface the framing qualifier above the chart; api / unknown
  show dollars. The JS formatter mirrors `core/framing.render_dollar` — it reads
  the server's decision, doesn't re-derive it.
- **States**: low-key loading shimmer, clean "No spend in this window" empty
  state, inline error + retry.
- **Endpoint extension (no new route)**: `/api/v1/cost` now returns a daily
  `series` (per date+agent+model) so the chart can split by model/agent — the
  grouped `rows` collapse that dimension. Dates use `AT TIME ZONE 'UTC'` (Rule 1).

## Tests
New assertion that `/api/v1/cost` returns the daily series; offline test green;
module JS validated with `node --check`; ASGI smoke confirms the series + the
by-model split. Full suite: 578 passed; `ruff`/`mypy` clean.

## Note
Per repo convention (architecture doc: no Playwright/Cypress for v1) this is
verified via the offline + API tests; a visual pass via `tj serve` is
recommended before merge.

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -238,6 +238,20 @@ async def test_get_cost_returns_aggregated_rows(client):
     assert "total_cost_usd" in data
 
 
+async def test_cost_includes_daily_series_for_chart(client):
+    """/api/v1/cost carries a daily-bucketed series (per date+agent+model) the
+    Cost chart consumes — finer than the grouped rows (#113)."""
+    await _ingest_sample_span(client)
+    resp = await client.get("/api/v1/cost", params={"since": "7d", "group_by": "day"})
+    assert resp.status_code == 200
+    data = resp.json()
+    assert "series" in data and isinstance(data["series"], list)
+    if data["series"]:
+        item = data["series"][0]
+        assert {"date", "agent_id", "model", "cost_usd",
+                "input_tokens", "output_tokens"} <= set(item)
+
+
 _FRAMING_KEYS = {
     "pricing_mode", "plan_tier", "plan_label", "plan_monthly_usd",
     "subscription_share_pct", "api_share_pct", "display_rule", "qualifier_text",
--- a/tokenjam/api/routes/cost.py
+++ b/tokenjam/api/routes/cost.py
@@ -11,6 +11,47 @@ from tokenjam.utils.time_parse import parse_since
 router = APIRouter(dependencies=[Depends(require_api_key)])
 
 
+def _daily_series(conn, agent_id, since_dt, until_dt) -> list[dict]:
+    """Daily-bucketed cost/tokens per (date, agent, model) for charting (#113).
+
+    The grouped `rows` collapse the agent/model dimension for day grouping, so
+    the time-series chart can't split by model/agent from them. This returns the
+    finer breakdown the chart needs without a new endpoint. Dates use
+    AT TIME ZONE 'UTC' (CLAUDE.md Rule 1) so they match Python's UTC dates.
+    """
+    if conn is None:
+        return []
+    clauses = ["model IS NOT NULL"]
+    params: list = []
+    if agent_id:
+        params.append(agent_id)
+        clauses.append("agent_id = $" + str(len(params)))
+    if since_dt is not None:
+        params.append(since_dt)
+        clauses.append("start_time >= $" + str(len(params)))
+    if until_dt is not None:
+        params.append(until_dt)
+        clauses.append("start_time <= $" + str(len(params)))
+    where = " AND ".join(clauses)
+    sql = (
+        "SELECT CAST(start_time AT TIME ZONE 'UTC' AS DATE) AS d, agent_id, model, "
+        "COALESCE(SUM(cost_usd), 0.0), "
+        "COALESCE(SUM(input_tokens), 0), "
+        "COALESCE(SUM(output_tokens), 0) "
+        "FROM spans WHERE " + where + " "
+        "GROUP BY d, agent_id, model ORDER BY d"
+    )
+    rows = conn.execute(sql, params).fetchall()
+    return [
+        {
+            "date": str(r[0]), "agent_id": r[1], "model": r[2],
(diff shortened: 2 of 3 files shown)


anilmurty (commented):

Reviewed the #113-specific commit `9964b9d` only — the rest of the PR diff is #110 + #112 work I already approved on #117 / #119 / #121.

**Acceptance criteria — all 10 met:**

1. ✓ Chart leads (uPlot, 7d default), precise daily table preserved below.
2. ✓ Window toggle (`?since=`) and group-by toggle (`?group_by=`) URL-synced via #112's machinery.
3. ✓ Trend chip ▲/▼ % from `/api/v1/cost/compare`.
4. ✓ Linear run-rate only: dashed mean-per-day line + "~$X over 30 days" with "linear run-rate, not a forecast" caveat. No moving averages.
5. ✓ Plan-tier framing — JS reads the server's `pricing_mode` rather than re-deriving the rules.
6. ✓ Loading shimmer, "No spend in this window" empty state, inline error+retry all present.
7. ✓ Theme via CSS custom properties; `useEffect` cleanup correct (`plot.destroy()` + `removeEventListener`).
8. ✓ No new endpoint — extends `/api/v1/cost` with a daily `series` field.
9. ✓ Daily table not regressed.
10. ✓ Rule 1 honored (`AT TIME ZONE 'UTC'`). No new f-string SQL, no `datetime.now()`, no CDN URLs.

**Nits / discussion points (non-blocking):**

- **`cost.py:26-30`** — `_daily_series` always computes the `(date, agent, model)` breakdown regardless of the table's `group_by`. Documented choice, but adds one extra GROUP BY scan per `/cost` call. For 90d on a busy DB this is the most expensive part of the response. Consider `?series=true` gating later if it shows up in profiling.
- **Payload scaling**: top-5 + "other" bucketing happens **client-side** in `buildCostSeries`. Full series cardinality crosses the wire — ~225KB at 90d × 5 agents × 5 models. Defensible for v1; server-side top-5 would shrink it.
- **Plan-tier framing**: JS derives `useTokens` from `framing.pricing_mode` rather than the explicit `framing.display_rule` field. Both work today since they're consistent, but the spec wanted `display_rule` as the switch so the server stays the single source of truth if the rules ever diverge from raw `pricing_mode`. Worth a tiny follow-up.
- **`test_cost_includes_daily_series_for_chart`** uses `if data["series"]:` — passes silently if empty. Asserting non-empty after `_ingest_sample_span` would be stronger.
- **Polling**: cadence lowered from 10s → 30s. Sensible for the heavier payload; worth calling out in the PR description.

**Note on merge order**: this PR's base is `feature/112-uplot-url-state`, so it can't merge until #117 + #121 land first. Plan to retarget to `main` once they do, per PR body. Tracking that.

Good work — chart looks faithful to the spec. Visual pass via `tj serve` recommended before merge as the PR body suggests.

Issue #129, opened 2026-06-18
[bug] Linear run-rate caption overstates: uses data-range denominator instead of window length


anilmurty:
**Type:** Bug / honesty.

## Symptom

The Overview and Cost screens render a "linear run-rate" projection like \`At this pace, ~\$622.6965 over 30 days\`. With a 7d window and \$62.27 of spend concentrated on 3 of those days, the projection should be \`\$62.27 / 7 × 30 = \$266.87\`. The displayed value (\`\$622.7\`) only matches \`\$62.27 / 3 × 30\` — i.e. the denominator is the range of days that *have* data, not the window length.

This systematically **overstates** the run-rate for any user whose activity is sparse within the window. In this case it's a 2.3× overstatement. For a marketing-honest projection caption, that's a substantial miss.

## Spec violation

\`.claude/lens-architecture.md\` §10 (Forecasting Boundary) explicitly defines what's allowed:

> Linear extrapolation: \`(spend over window) / (days in window) × (days remaining in cycle)\`

The implementation uses \`(spend over window) / (data range in days)\` instead. \`data range\` ≤ \`window\`, so the run-rate is always ≥ the spec'd value.

## Two related polish items to fix in the same PR

1. **Caption inconsistency.** Cost screen says "\`(linear run-rate, not a forecast)\`". Overview says just "\`(linear run-rate)\`". The "not a forecast" qualifier is part of the honesty discipline — add it to Overview for parity.

2. **Y-axis labels lack the \$ prefix on the spend chart.** Ticks show "\`50.0000\`", "\`25.0000\`", "\`\$0.0000\`" — only the bottom gridline has the \`\$\`. The header total renders correctly ("\`\$62.2697\`"), so the formatter is available — just not threaded into the axis tick formatter. Either prefix every tick with \`\$\` or use a more compact \`\$50\` / \`\$25\` format (no trailing zeros) to match the chart's intent.

## Acceptance criteria

1. Run-rate uses window length as the denominator: \`(spend) / max(window_days, 1) × 30\`.
2. The "not a forecast" qualifier appears on the Overview run-rate caption.
3. Y-axis labels carry the \`\$\` prefix (or appropriate token-share symbol when in subscription mode) and use compact formatting (no superfluous trailing zeros for round dollar amounts).
4. New unit test on the run-rate calculation confirms the formula matches the spec for a sparse-data window.

## Files likely touched

[...]

Issue #128, opened 2026-06-18
[bug] Spend chart: missing hover tooltip + chart-as-button drill-through is hostile


anilmurty:
**Type:** Bug / UX.

## Symptom

The Overview spend chart (and the Cost screen chart) has no hover tooltip — hovering over a data point shows nothing. Worse, the chart container is a click-target that drills to the Cost tab on any click, so the user cannot click on a data point to inspect it. The combined effect: there is no way to read the value at a point on the chart.

## Root cause

uPlot supports a crosshair + tooltip via its \`cursor\` config (a few lines of JSON: \`cursor: { drag: { x: false, y: false }, focus: { prox: 16 } }\` plus a hook that renders the values). The Overview implementation didn't wire it up, presumably because the chart is wrapped in a click-handler that drills to the Cost screen.

## Acceptance criteria

1. Hovering the chart shows a tooltip with the date and the value at the nearest data point, in the framed unit (USD or token-share, per \`framing.pricing_mode\`).
2. The tooltip respects the active theme (light/dark).
3. Drill-through to the Cost screen is preserved but moves off the chart container — implement it via a small explicit "View Cost details →" link/button near the chart header, not by making the chart itself a click-target.
4. The same chart on the Cost screen gets the same tooltip behavior (consistency).
5. Keyboard accessibility: the drill link is a focusable \`<a>\` or \`<button>\`, not a \`<div onclick>\`.

## Out of scope

- Multi-series tooltip when group-by ≠ total (nice-to-have follow-up).
- Pinned tooltips on click.

## Files likely touched

- \`tokenjam/ui/index.html\` — \`LineChart\` component (~\`:722-757\`), Overview Cost band, Cost view.

anilmurty:
Shipped in PR #130 (commit `69b572a`, merged 2026-06-18). Verified against the live UI: `classifyFinding` returns 4 distinct states; chart now has crosshair tooltips with explicit drill-link; run-rate uses window-length denominator with the 'not a forecast' qualifier and $-prefixed compact y-axis labels. GitHub auto-closed #126 only — closing this manually since the comma-list form didn't catch.

Issue #127, opened 2026-06-18
[bug] Overview recoverable tiles: 'not ready' state conflates three different analyzer outcomes


anilmurty:
**Type:** Bug / UX-honesty.

## Symptom

The Overview's recoverable band uses "— not ready" as a catch-all whenever an analyzer's \`estimated_recoverable_usd\` is \`None\` or \`0\`. But that condition fires for three structurally different reasons, and the tile's hint copy is wrong in two of them.

### Observed on a typical install

| Analyzer | Actual state (per Optimize tab) | Tile renders | Why it's wrong |
|---|---|---|---|
| Cache | 100% efficacy on opus-4-7 — already at the ceiling | "— not ready / Raise cache-read efficacy toward the 80% ceiling" | The hint contradicts reality. Efficacy is already past the target. |
| Script | "No clusters above threshold (≥20 identical signatures)" | "— not ready / Replace a deterministic call-pattern with a script" | The hint implies an actionable pattern exists. Analyzer says none does. |
| Trim | Capture off (\`[capture] prompts = false\`) | "— not ready / Not run on Overview — open the Optimize tab" | This one is fine — same shape, but accurate copy. |

So users at peak efficiency see "not ready" warnings that say to fix things that are already fixed.

## Three distinct states, three distinct tiles

The recoverable-band tile rendering needs three branches, not one:

1. **"Already optimized"** — analyzer ran, found no waste because the user is at the ceiling. Tile shows the achieved metric ("100% cache efficacy") and a positive framing. No "not ready" word.
2. **"No findings in this window"** — analyzer ran, found nothing actionable. Tile says "No candidates in this window" with the window length, and possibly suggests widening the window.
3. **"Not run / disabled"** — analyzer can't run because of capture flag / missing data / skipped on fast path. This is the *only* state that should use "— not ready" and an enablement hint.

## Acceptance criteria

1. Each analyzer's tile rendering reads a discriminator from the finding (e.g. an enum \`finding.state ∈ {actionable, at_ceiling, no_findings, not_ready}\`) rather than checking \`estimated_recoverable_usd in (None, 0)\` to decide tile shape.
2. The "at-ceiling" state is positive in tone, not warning-styled.
3. The "no-findings" state surfaces the window length so the user understands the scope.
4. Honesty rule: no tile says "raise X toward Y" when X is already at or past Y.
5. Trim's current copy ("Not run on Overview — open the Optimize tab") is the gold-standard pattern — keep it.

[...]

PR #130, opened 2026-06-18
Fix 4 Lens UI bugs (#126–#129)

### LINKED ISSUE #126: [bug] Overview/Optimize: Downsize analyzer not rendered (typed slot ignored)
**Type:** Bug (Lens v1 regression).

## Symptom

The new Optimize tab shows **Cache, Cache recommend, Script, Trim** — but **no Downsize section**, even when the analyzer runs. The Overview's recoverable band shows "Downsize — not ready" in the same situation.

## Root cause

\`OptimizeReport.downgrade\` is a typed slot on the report dataclass (carried over for backwards-compat with the original analyzer registry rework). The wave-2 analyzers (\`cache\`, \`cache-recommend\`, \`script\`, \`trim\`) attach into the generic \`findings\` dict. The Optimize tab UI iterates over \`report.findings\` and forgets to render the \`report.downgrade\` typed slot. Same root cause for the Overview tile.

## Acceptance criteria

1. Optimize tab renders a "Downsize" section when \`report.downgrade\` is non-null, alongside the other analyzers. Section content matches what \`tj optimize downsize\` shows (candidate sessions, examples, framed dollar/token recoverable, honesty caveat).
2. Overview recoverable band's Downsize tile reads from \`report.downgrade.estimated_recoverable_usd\` (already populated per #111).
3. Section ordering on Optimize: Downsize first (highest-impact analyzer historically), then the wave-2 set.
4. \`?finding=downsize\` deep-link from the Overview tile scrolls/highlights the Downsize section correctly.
5. New integration test asserts the response shape carries \`downgrade\` and the rendered HTML includes a Downsize section header.

## Scope

UI-only fix. No analyzer code changes. No new endpoint.

anilmurty:
Closes #126, #127, #128, #129. One PR — all four are in `tokenjam/ui/index.html`.

### #126 — Downsize (typed slot) was silently dropped
`report.downgrade` is a typed slot, not in `findings`. When it's `null` (no
candidates — common for heavy users whose sessions don't match the small-shape
heuristic) the Optimize tab rendered nothing and the Overview said "not ready".
- Optimize tab now renders a **Downsize section first**, with a "no candidates
  in this window" state when the slot is null.
- Overview always includes a Downsize tile.

### #127 — "not ready" conflated three outcomes
New `classifyFinding` returns one of four honest states, read from richer
finding signals (not `usd in (None,0)`):
- **actionable** — recoverable waste found (carries the number)
- **at_ceiling** — already optimized (e.g. cache ≥80% efficacy) → positive, ✓,
  never a warning, never "raise X toward Y"
- **no_findings** — ran, nothing actionable → shows the window length
- **not_ready** — disabled / missing dep / skipped on the fast path (the only
  state that keeps the "not ready" wording)

### #128 — chart had no tooltip and was a hostile click-target
- uPlot cursor **tooltip** (date + framed value, themed) on the Overview hero
  and the Cost screen.
- The chart is no longer wrapped in a click handler; drill-through is an
  explicit focusable **"View Cost details →"** link in the header.

### #129 — run-rate overstated + axis polish
- Run-rate denominator is now the **window length** (`spend / window_days × 30`)
  per `lens-architecture` §10, not the data range (which overstated sparse
  windows ~2.3×). New pure `windowDays` / `runRateProjection` helpers.
- "**not a forecast**" qualifier added to the Overview caption (parity with Cost).
- Y-axis uses a compact `$` formatter (`$0` / `$25` / `$1.5k`), not `$0.0000`.

## Verified live
`tj serve` against seeded data (sparse 7d window + downsize candidates + a
high-efficacy cache model): the run-rate reads `~$21.26` (= 4.96/7×30, was
~$50), the Downsize section renders first and highlights on the
`?finding=downsize` deep-link, Script shows "No candidates / nothing actionable
in the last 7d", and the y-axis shows compact `$`.

## Tests
- Integration: `/api/v1/optimize` always carries the `downgrade` key.
- `tests/unit/test_lens_ui_regression.py` static-guards each fix in the served
  source (no JS test runner in the Python CI job) — e.g. the buggy
  `/ ys.length` run-rate denominator and the `band-hero` click-target are
  asserted gone.
- Full suite green (689); `ruff` clean; offline test still passing.

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -276,6 +276,17 @@ async def test_optimize_response_includes_framing_block(client):
     assert _FRAMING_KEYS <= set(data["framing"])
 
 
+async def test_optimize_response_always_carries_downgrade_key(client):
+    """The downsize typed slot must always be present (null when no candidates)
+    so the UI can always render a Downsize section (#126)."""
+    await _ingest_sample_span(client)
+    resp = await client.get("/api/v1/optimize?since=30d")
+    data = resp.json()
+    if data.get("error") == "no_data":
+        pytest.skip("no spans landed for optimize in this fixture")
+    assert "downgrade" in data  # present even when null
+
+
 async def test_root_serves_lens_title(client):
     """Brand pass (#114): the served dashboard <title> is 'TokenJam Lens'."""
     resp = await client.get("/")
--- a/tests/unit/test_lens_ui_regression.py
+++ b/tests/unit/test_lens_ui_regression.py
@@ -1,0 +1,79 @@
+"""Static regression guards for the Lens UI bug fixes (#126–#129).
+
+The dashboard is a single-file Preact SPA with no JS test runner in the Python
+CI job, so these assert the *served source* contains the corrected logic and no
+longer contains the buggy patterns. They're intentionally narrow — each pins one
+bug's fix so a future edit that reintroduces it fails here.
+"""
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+_UI = Path(__file__).parent.parent.parent / "tokenjam" / "ui" / "index.html"
+
+
+@pytest.fixture(scope="module")
+def html() -> str:
+    return _UI.read_text(encoding="utf-8")
+
+
+# --- #126: Downsize typed slot always rendered ----------------------------- #
+def test_downsize_section_always_renders(html):
+    # The no-candidates branch renders a literal Downsize section id instead of
+    # returning null, so the section is never silently dropped.
+    assert 'id="opt-downsize"' in html
+    assert "No downsize candidates in this window" in html
+
+
+def test_downsize_is_first_in_optimize_order(html):
+    assert "const order = ['downsize', 'cache', 'cache-recommend', 'script', 'trim']" in html
+
+
+# --- #127: four distinct recoverable-tile states --------------------------- #
+def test_recoverable_band_has_four_states(html):
+    assert "function classifyFinding" in html
+    for state in ("'actionable'", "'at_ceiling'", "'no_findings'", "'not_ready'"):
+        assert state in html, f"missing tile state {state}"
+    # at-ceiling must not reuse the "raise toward ceiling" hint.
+    assert "Already optimized" in html
+
+
+def test_recoverable_band_not_a_single_not_ready_catchall(html):
+    # The old crude check ("ready = fd && usd != null" → "— not ready" for
+    # everything else) must be gone.
+    assert "const ready = fd && usd != null" not in html
+
+
+# --- #128: chart tooltip + non-button drill -------------------------------- #
+def test_chart_has_hover_tooltip(html):
+    assert "function chartTooltipPlugin" in html
+    assert "plugins: [chartTooltipPlugin(" in html
+
+
+def test_overview_chart_is_not_a_click_target(html):
+    # The cost hero must no longer be wrapped in an <a class="band-hero">; drill
+    # is an explicit link.
+    assert 'class="chart-card band-hero"' not in html
+    assert 'class="drill-link"' in html
+    assert "View Cost details" in html
+
+
+# --- #129: run-rate denominator + caption + $ axis ------------------------- #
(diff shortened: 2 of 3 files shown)


anilmurty (commented):
## Review: ✅ Approve

All four bugs cleanly addressed in one well-scoped PR. CI green on 3.10/3.11/3.12 + TS.

**Per-bug verification:**

**#126 — Downsize rendering ✓**
- Recoverable tile reads from `opt.downgrade` typed slot (`index.html:1936`), not `findings`.
- Optimize section order pinned: `order = ['downsize', 'cache', ...]` (`:2096`). Null-downgrade branch still renders an honest "no candidates in this window" section.
- `?finding=downsize` deep-link scrolls + highlights correctly (`:2089-2092`).
- Integration test asserts `"downgrade" in data`.

**#127 — Tile state distinction ✓**
- New `classifyFinding()` (`:1782`) returns 4 distinct states: `actionable` / `at_ceiling` / `no_findings` / `not_ready`.
- At-ceiling renders positive: `class="rec-tile ok"`, green `--success`, `✓ {metric}`, "Already optimized — nothing to recover" (`:1968`).
- `no_findings` surfaces window length: "Nothing actionable in the last ${since}" (`:1975`).
- At-ceiling never says "raise toward Y" anymore (only reached when `efficacy >= 0.80`).
- Trim's gold-standard copy preserved.

**#128 — Chart tooltip + drill ✓**
- New `chartTooltipPlugin()` (`:845`) renders themed crosshair tooltip with date + value per series via uPlot `setCursor` hook; skips the dashed run-rate line. Themed via CSS vars.
- Overview hero changed from `<a class="band-hero">` → `<div class="chart-card">` (`:1937`); drill moved to an explicit `<a class="drill-link" href="#/cost?...">View Cost details →</a>` (`:1944`) — keyboard-accessible.
- Tooltip applied inside `SpendChart` itself (`:910`) so the Cost screen chart inherits identical behavior.

**#129 — Run-rate + polish ✓**
- `windowDays(since)` parses `7d` / `24h` / date-range (`:838`). `runRateProjection = (windowTotal / max(windowDays, 1/24)) * 30` (`:850`). Both Cost (`:1316`) and Overview (`:1916`) call it. The buggy `ys.reduce(...) / ys.length` is gone in both spots.
- Overview caption now reads `(linear run-rate, not a forecast)` (`:1958`), matching Cost.
- `fmtAxisUsd()` (`:856`) emits `$0` / `$25` / `$1.5k` with no trailing zeros. Plumbed through `SpendChart` via `axisFmtY`.

**Test quality — `test_lens_ui_regression.py`:** Tight, not loose. Asserts the literal buggy substrings are *gone* (`"ys.reduce((a, b) => a + b, 0) / ys.length" not in html`, `'class="chart-card band-hero"' not in html`, `"(linear run-rate)<" not in html`) AND that the new function names are present (`classifyFinding`, `chartTooltipPlugin`, `windowDays`, `runRateProjection`, `fmtAxisUsd`). Section order pinned. Smart workaround for the CI-without-JS-runner constraint — pins the fix without over-pinning impl details.

**Quality checks:** Rule 18 preserved (no new external HTTP loads). No SQL or Python timestamp changes. Adjacent screens (Status/Traces/Alerts/Drift/Budget) untouched — diff is concentrated on `SpendChart`, the Overview hero band, recoverable tiles, and the Optimize downsize branch. No collateral damage.

**Nits (non-blocking):**
- `runRateProjection` recomputes `windowDays(since)` while callers also compute it inline — minor duplication.
- `chartTooltipPlugin` skips the run-rate series by string-matching `label === 'run-rate'`; brittle if that label string ever changes. Local-only, harmless today.

Ship it.

anilmurty:
Thanks! Both nits addressed in the latest commit:

1. **windowDays duplication** — callers now derive the per-day rate via `runRateProjection(windowTotal, since, 1)` (then `× 30` for the projection), so `windowDays` is computed in exactly one place. Math is unchanged.
2. **Brittle run-rate skip** — the run-rate series now carries an `_isRunRate: true` flag and the tooltip skips on that, not on the `'run-rate'` label string.

Issue #136, opened 2026-06-18
[bug] Chart x-axis date labels are inconsistent — '6/15/26' mixed with '6/16' and '6/17'


anilmurty:
## Symptom

The spend chart's x-axis dates render inconsistently:

- First date: `6/15/26` (M/D/YY — with year suffix)
- Subsequent dates: `6/16`, `6/17` (M/D — no year)

The mix is visually jarring and the `/26` reads awkwardly — at a glance it could be misread as a strange day-of-month suffix rather than the year `2026`. Pure formatting issue, not a data bug.

## Suggested fix

Two clean options — pick whichever fits the existing design tone - preference is for option A

**Option A — Abbreviated month/day on every tick.** Most readable.

```
Jun 15        Jun 16        Jun 17
```

**Option B — ISO format.**

```
2026-06-15    2026-06-16    2026-06-17
```

Both are unambiguous. The current `6/15/26` mixed with `6/16` is neither.

If a year-suffix on the first tick is genuinely useful (e.g. when the window crosses a year boundary), use the full year `2026` rather than `26`, and either show it on every tick or only when the year actually changes.

## Acceptance criteria

1. All x-axis date labels use the same format.
2. The format is unambiguous (no `/26` two-digit-year suffix that could be misread).
3. Behavior holds across all chart locations (Overview hero, Cost screen) and all window sizes (24h shows hours, 7d+ shows dates).

## Files likely touched

- `tokenjam/ui/index.html` — uPlot axis values formatter for the time axis on `SpendChart`.

anilmurty:
Shipped in PR #137 (merged 2026-06-18). Verified against live UI: window-bucketed series with zero-fill across the full window; `daily_rate × days_remaining_in_cycle` projection with 'by end of {month}' caption; cache at-ceiling renders the positive ✓ green tile; single date-format per axis. GitHub auto-closed only #132 from the comma-list — closing this manually.

Issue #135, opened 2026-06-18
[bug] Cache 'at_ceiling' state not detected on Overview — 100% efficacy shows as 'No candidates' instead of 'Already optimized'


anilmurty:
## Symptom

The Optimize tab reports **100% cache efficacy** for `anthropic/claude-opus-4-7` (9.4k input / 77.6M cache tokens). The Overview's recoverable-band Cache tile should render the positive `at_ceiling` state per #127's design:

> "✓ Already optimized — 100% efficacy" in green / positive styling

But it actually renders the muted `no_findings` state instead:

> "Cache — No candidates — Nothing actionable in the last 7d."

The user has done the cache-optimization work; the dashboard should celebrate that, not shrug it off as "nothing actionable."

## Root cause

`classifyFinding()` (introduced in #130 to fix #127) returns four distinct states: `actionable` / `at_ceiling` / `no_findings` / `not_ready`. The function correctly distinguishes these for the analyzers in general, but the Cache branch is misclassifying 100% efficacy as `no_findings` rather than `at_ceiling`.

Likely culprit: the classifier checks `flagged.length === 0` to mean "no findings" — but `flagged` is the list of rows *below* the 80% ceiling. When all rows are at or above the ceiling, `flagged` is empty, which the classifier interprets as "no findings" rather than "everything's at the ceiling."

## Fix sketch

The Cache branch needs to distinguish:
- `rows.length === 0` → no data → `no_findings` (no usage in window)
- `rows.length > 0` AND every row's efficacy ≥ 80% (ceiling) → `at_ceiling`
- `flagged.length > 0` AND `estimated_recoverable_usd > 0` → `actionable`
- otherwise → `no_findings`

The 80% number is the existing constant `EFFICACY_CEILING` in `cache_efficacy.py` — UI should ideally read it from the response rather than hardcode it.

## Acceptance criteria

1. When all Cache rows in the window are at or above the 80% ceiling, the Overview tile renders the `at_ceiling` state with positive styling (green check, "Already optimized — 100% efficacy" or similar).
2. The displayed efficacy number comes from the data (could be 87% or 100%, not always exactly 100).
3. Other analyzers that don't have an `at_ceiling` concept (Downsize, Script) keep their current `no_findings` rendering — this fix is specifically the Cache classifier branch.

## Verification

With current install state (one user, 100% cache efficacy), reload Overview → Cache tile should be green-styled with the "Already optimized" copy, not the muted "No candidates" copy.

[...]

Issue #134, opened 2026-06-18
[bug] Run-rate projection formula is circular for 30d window and undershoots for 90d


anilmurty:
## Symptom

The "linear run-rate, not a forecast" caption uses the formula `(total / window_days) × 30`. This produces an honest projection for short windows but degenerate or misleading values for longer ones:

| Window | Total | Formula | Caption | Issue |
|---|---|---|---|---|
| 7d  | \$62.27 | 62.27/7×30   | \"~\$266.87 over 30 days\" | OK — projects last week's pace forward |
| 30d | \$62.27 | 62.27/30×30  | \"~\$62.27 over 30 days\" | **Circular** — the projection equals the input |
| 90d | \$62.27 | 62.27/90×30  | \"~\$20.76 over 30 days\" | **Undershoots** recent activity — projection is *lower* than what user actually spent in the last 7 days |

The 30d window's projection is uninformative ("at this 30-day pace, you'll spend the same in 30 days"). The 90d window's projection actively misleads — it understates current usage because it averages in old quiet periods.

## Why this matters

A user widening the window to 90d would expect more confidence, not a *smaller* projected number than the 30d view. The projection should be roughly stable across window choices (give or take recent trend), not inversely correlated with window length.

## Two cleaner framings

**Option A (simplest):** Show daily run-rate instead of monthly projection. \"\$8.90/day\" is stable across window choice (depends only on activity level, not on window). Users can multiply in their heads if they want a monthly figure.

**Option B (closer to the architecture-doc spec):** Use `daily_rate × days_remaining_in_calendar_cycle`. The `.claude/lens-architecture.md` §10 (Forecasting Boundary) actually defines this:

> `Linear extrapolation: (spend over window) / (days in window) × (days remaining in cycle)`

`days remaining in cycle` (not a fixed `× 30`) makes the projection cycle-relative and intuitive: \"at this pace, ~\$X by end of June.\"

Either option is fine; pick one. **Option B** matches the spec verbatim and aligns with how `tj optimize budget-projection` already frames things.

## Acceptance criteria

1. Run-rate caption produces a meaningful number for every window selection (no circular values, no inverse-correlation with window length).
2. Caption keeps the "not a forecast" qualifier and the "linear run-rate" framing.
3. Formula matches the architecture doc's spec (Option B), or a clearly-labeled simpler alternative (Option A).

## Files likely touched

[...]

Issue #133, opened 2026-06-18
[bug] Spend chart x-axis spans only the data range, not the selected window (7d/30d/90d look identical)


anilmurty:
## Symptom

The window toggle (24h / 7d / 30d / 90d) changes the data being queried, but the **chart looks visually identical** across selections because:

1. The x-axis only spans the range where spending occurred — not the full window length.
2. The tick granularity is auto-picked by uPlot based on the (cramped) visible range — so all the non-24h windows show **hourly** ticks regardless of selection.

The 24h window is the only one that currently renders sensibly because its data range and its desired granularity happen to align with the auto-pick.

### Concretely

With \$62.27 spread across two days (6/15 and 6/17), all four windows render an x-axis from ~6/15-6pm to ~6/17-3pm with hourly ticks (`6pm, 9pm, 12am, 3am, ...`):

| Window | Total | Visible x-axis | Ticks | Run-rate caption |
|---|---|---|---|---|
| 24h | \$58.31 | last 24h | hourly | "~\$1,749 over 30 days" |
| 7d | \$62.27 | ~2 days | hourly | "~\$266 over 30 days" |
| 30d | \$62.27 | ~2 days | hourly | "~\$62 over 30 days" |
| 90d | \$62.27 | ~2 days | hourly | "~\$21 over 30 days" |

Same chart, same total, same hourly ticks. Only the dashed mean-per-day line and the run-rate caption shift.

## Why this matters

"Spend over time" should let a user see *patterns* — bursty days, idle weeks, ramping usage. The current rendering hides all of that. A user widening from 7d to 30d gets no new visual information.

## Acceptance criteria

1. **X-axis spans the full selected window**, regardless of where data falls within it:
   - `24h` → last 24 hours
   - `7d`  → last 7 days
   - `30d` → last 30 days
   - `90d` → last 90 days
2. **Days with no spending render as \$0** on the chart (continuous series with zero values), not as missing points.
3. **Tick granularity matches the window**:
   - `24h` → hourly ticks (~6 labels)
   - `7d`  → daily ticks (~7 labels)
   - `30d` → daily or every-3-days ticks (~10 labels)
   - `90d` → weekly ticks (~12 labels)
   The exact pick is judgment-call; the spirit is "the unit you'd describe the window in" (hours / days / weeks). Don't leave it to uPlot defaults which picked hourly across the board.
4. The total (top-right of the chart card) is unchanged — already correct.

## Out of scope

[...]

PR #137, opened 2026-06-18
Fix 5 Lens UI bugs (#132–#136)

### LINKED ISSUE #132: [bug] First-load lands on /#/overview URL but sidebar highlights Status and content is blank
## Symptom

Opening `http://127.0.0.1:7391/` (bare URL, no hash) lands at `/#/overview` in the URL bar, **but**:

- The sidebar highlights **Status**, not Overview.
- The main content area is **blank**.

Hard-refreshing (Cmd+Shift+R) fixes it: sidebar highlights Overview, the content populates with the spend chart + tiles.

## Root cause (hypothesis)

The default-route handler at `tokenjam/ui/index.html:528-530` does:

```js
if (!location.hash || location.hash === '#/') {
  location.hash = '#/overview';
}
```

This pushes a new hash but the initial Preact render has already started with `getRoute()` returning the default fallback (`'status'`, from `:767`/`:770`). The resulting `hashchange` event apparently doesn't trigger a re-render of the active view + sidebar highlight on first load.

## Acceptance criteria

1. Opening `http://127.0.0.1:7391/` with no hash lands on Overview — both URL bar shows `/#/overview` **and** the sidebar highlights Overview and the content renders Overview's three bands.
2. No race condition between the redirect and the initial render.
3. Hard-refresh and back-button navigation behave consistently with first-load.

## Suggested approach

Two cleanest fixes (pick whichever fits the component model better):
- Render Overview directly when the hash is empty, instead of redirecting through `location.hash = '#/overview'`. Skip the hashchange round-trip entirely.
- After pushing the redirect hash, explicitly call the route-state setter (`setRoute(getRoute())` or equivalent) so the Preact tree re-renders with the new state.

## Verification

Open the dashboard in an incognito window. Without hard-refreshing, the URL bar should show `/#/overview` AND the Overview content should render with the sidebar highlighting Overview.

## Files likely touched

- `tokenjam/ui/index.html` — default-route handler near `:528-530` and the initial state computation in `App`/router.

anilmurty:
Closes #132, #133, #134, #135, #136. Verified live across all four windows + first-load.

### #132 — First-load lands on Overview (was blank + Status highlighted)
`getRoute()` now defaults to `overview` (the single source of the default), and
the render-time `location.hash = '#/overview'` redirect — which raced the
initial render and left the content blank while the sidebar highlighted Status —
is replaced by a cosmetic `history.replaceState` in an effect. Bare URL now
renders Overview with the sidebar highlighting it, no hard-refresh needed.

### #133 + #136 — Spend chart spans the full window, with consistent labels
- `/api/v1/cost` returns a **window-bucketed** series — hourly for ≤2-day
  windows, daily otherwise — as epoch buckets, plus `series_bucket` and
  `window_start`/`window_end`.
- The chart builds a **continuous, zero-filled grid across the full window** and
  pins the x scale to it, so **7d/30d/90d render distinctly** and idle periods
  show as `$0`.
- Tick increments are constrained to the bucket unit (no more duplicate
  "Jun 11 / Jun 11" labels), and labels use **one consistent format** —
  "6 PM" (hourly) / "Jun 15" (daily) — never `6/15/26` mixed with `6/16`.

### #134 — Run-rate is cycle-relative
Replaced the fixed `×30` (circular at a 30d window, undershooting at 90d) with
`daily_rate × days-remaining-in-calendar-cycle`, captioned **"~$X by end of
{month}"** per `lens-architecture` §10. Keeps "linear run-rate, not a forecast".

### #135 — Cache "at ceiling" now detected
`classifyFinding` no longer gates `at_ceiling` on input volume — a fully-cached
model has tiny *uncached* input, so 100% efficacy was misclassified as
"No candidates". It now renders the positive **"✓ Already optimized — 100% (or
86%…) cache efficacy"** state. The efficacy ceiling is exposed on the cache
finding so the UI reads it instead of hardcoding `0.80`.

## Verified live (`tj serve`, seeded data)
- Bare URL → Overview (sidebar highlighted, content populated).
- 24h → hourly ticks; 7d → 7 daily ticks; 30d → weekly over the month; 90d →
  weekly over the quarter — all distinct, full-window, zero-filled.
- Run-rate caption "~$X by end of June".
- Cache tile green "✓ 86% cache efficacy / Already optimized".

## Tests
- Integration: window-series shape (`bucket` + `series_bucket` + window bounds);
  hourly bucketing for short windows.
- `tests/unit/test_lens_ui_regression.py` guards for each fix (overview default,
  full-window range, time-axis formatter, cycle-relative caption, cache ceiling
  not volume-gated).
- Full suite green (695); `ruff check tokenjam/` + `mypy` clean.

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -238,18 +238,30 @@ async def test_get_cost_returns_aggregated_rows(client):
     assert "total_cost_usd" in data
 
 
-async def test_cost_includes_daily_series_for_chart(client):
-    """/api/v1/cost carries a daily-bucketed series (per date+agent+model) the
-    Cost chart consumes — finer than the grouped rows (#113)."""
+async def test_cost_includes_window_series_for_chart(client):
+    """/api/v1/cost carries a window-bucketed series (per bucket+agent+model)
+    plus the bucket size and window bounds so the chart can span the full
+    selected window with zero-fill (#113/#133)."""
     await _ingest_sample_span(client)
     resp = await client.get("/api/v1/cost", params={"since": "7d", "group_by": "day"})
     assert resp.status_code == 200
     data = resp.json()
     assert "series" in data and isinstance(data["series"], list)
+    assert data["series_bucket"] in ("hour", "day")
+    assert isinstance(data["window_start"], int) and isinstance(data["window_end"], int)
+    assert data["window_end"] >= data["window_start"]
     if data["series"]:
         item = data["series"][0]
-        assert {"date", "agent_id", "model", "cost_usd",
+        assert {"bucket", "agent_id", "model", "cost_usd",
                 "input_tokens", "output_tokens"} <= set(item)
+        assert isinstance(item["bucket"], int)  # epoch seconds
+
+
+async def test_cost_series_buckets_hourly_for_short_window(client):
+    """A ≤2-day window buckets hourly so 24h renders with hourly ticks (#133)."""
+    await _ingest_sample_span(client)
+    resp = await client.get("/api/v1/cost", params={"since": "24h"})
+    assert resp.json()["series_bucket"] == "hour"
 
 
 _FRAMING_KEYS = {
--- a/tests/unit/test_lens_ui_regression.py
+++ b/tests/unit/test_lens_ui_regression.py
@@ -77,3 +77,40 @@ def test_overview_caption_says_not_a_forecast(html):
 def test_axis_uses_compact_dollar_formatter(html):
     assert "function fmtAxisUsd" in html
     assert "axisFmtY=" in html
+
+
+# --- #132: first-load lands on Overview (no redirect race) ----------------- #
+def test_first_load_defaults_to_overview(html):
+    # getRoute defaults to overview; the render-time hash redirect is gone.
+    assert "|| 'overview'" in html
+    assert "location.hash = '#/overview'" not in html
+    assert "history.replaceState(null, '', '#/overview')" in html
+
+
+# --- #133/#136: chart spans full window + consistent date labels ----------- #
+def test_chart_spans_full_window_with_buckets(html):
+    assert "function windowDays" in html
+    assert "series_bucket" in html and "window_start" in html
+    # x scale pinned to the window range, not the data range.
+    assert "range: [data[0][0]" in html
+
+
+def test_axis_time_labels_consistent(html):
+    assert "function fmtAxisTime" in html
+    # daily labels use abbreviated month/day ("Jun 15"), one format per axis.
+    assert "month: 'short', day: 'numeric'" in html
+
+
+# --- #134: run-rate is cycle-relative, not a fixed ×30 --------------------- #
+def test_run_rate_is_cycle_relative(html):
+    assert "function cycleRemaining" in html
+    assert "by end of ${cyc.label}" in html
+    assert "over 30 days" not in html  # the circular/undershooting framing is gone
+
+
+# --- #135: cache at_ceiling not gated on input volume --------------------- #
+def test_cache_at_ceiling_not_volume_gated(html):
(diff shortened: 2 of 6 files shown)


anilmurty (commented):
## Review: ✅ Approve

All five bugs addressed cleanly. Tight diff (+200/-71). CI green on 3.10/3.11/3.12 + TS.

**Per-bug verification:**

- **#132 ✓** `getRoute()` defaults to `'overview'` (`index.html:783, 786`); render-time hash redirect replaced with cosmetic `history.replaceState` inside `useEffect` (`:2207-2210`). No hashchange re-fire, no race.

- **#133 ✓** `cost.py:_window_series` returns hourly buckets (≤2d windows) or daily buckets (longer), plus `series_bucket` + `window_start/end` metadata. SQL uses parameterized `date_trunc($1, ...)` (Rule 7 honored — comment even calls this out). UI `buildCostSeries` zero-fills the full grid (`:1289-1308`); `SpendChart` pins `scales.x.range` to the window bounds (`:961`); tick `incrs` constrain spacing to bucket units (`:967-970`).

- **#134 ✓** New `cycleRemaining()` (`:855-861`) computes days to end-of-calendar-month. Projection: `runRate × cyc.days`. Caption: `"~$X by end of ${cyc.label}"`. Both `* 30` and `"over 30 days"` removed (regression-test grep enforces).

- **#135 ✓** `classifyFinding` Cache branch (`index.html:1843-1850`) removes the `CACHE_MIN_INPUT` volume gate. Ceiling read from `fd.efficacy_ceiling` with 0.80 constant as safe fallback. Backend exposes `efficacy_ceiling` on `CacheEfficacyFinding` (`cache_efficacy.py:74`); `report_to_dict` serializes via `asdict`. Round-trip in `runner.py` reads with default.

- **#136 ✓** `fmtAxisTime(epoch, bucket)` (`:879-883`) returns `"6 PM"` (hourly) or `"Jun 15"` (daily) using `toLocaleTimeString`/`toLocaleDateString` with `timeZone: 'UTC'` matching the UTC-aligned buckets. Single format per axis.

**Quality checks clean:** no f-string SQL (bucket bound as `$1`); `utcnow()` used in `cost.py` (Rule 9); no external HTTP loads (Rule 18); the `+3` lines in `cache_efficacy.py` and `+1` in `runner.py` are purely additive. Tests look good — `test_lens_ui_regression.py` has static-grep guards for each fix, `test_api.py` asserts the new response shape (`series_bucket`, `window_start/end`).

**On your flagged judgment call (90d projects smaller than 7d):** Concur. Window-average is what the spec says; the new "~$X by end of June" framing is interpretable at every window selection without the circular 30d artifact. Stable-across-window projections would be a deliberate spec change — worth a separate issue, not a blocker here.

**Two non-blocking follow-ups worth filing:**

1. **`cycleRemaining()` uses the calendar month**, not the user's configured `[budget.<provider>] cycle_start_day`. With `cycle_start_day = 15`, the "by end of June" caption would be off by half a month. The existing `budget-projection` analyzer honors per-provider cycle days — symmetric treatment would be nicer for honesty parity.

2. **`xs.length > 5000` guard in `buildCostSeries`** is good defense but at hourly bucketing a 200d window would silently hit it (returns `null` → empty state). Fine today since the window picker tops out at 90d; worth noting if/when longer windows or finer bucketing ship.

Ship it.

anilmurty:
Thanks! Both follow-ups filed (keeping this PR as the approved bug-fix scope):

- **#138** — run-rate cycle should honor `[budget.<provider>] cycle_start_day` instead of the calendar month, for parity with `budget-projection`.
- **#139** — the `xs.length > 5000` guard would silently empty a very long hourly window; coarsen the bucket / surface a note instead, if longer windows or finer bucketing ever ship.

And agreed on the 90d-vs-7d projection: window-average per the spec; a recency-weighted variant would be a deliberate §10 change, separate issue.

Issue #116, opened 2026-06-17
[feature] tj report --reuse — generate HTML report and Markdown skeleton bundle


anilmurty:
**Type:** Feature / CLI + export pipeline.

**Depends on:** #115 (the Reuse analyzer).

## Background

The Reuse analyzer (#115) produces clusters of sessions with repeated planning. To make the findings actionable, users need a way to **see** the skeleton (with variable slots highlighted) and **copy** the skeleton into their tooling — Claude Code slash commands, saved prompts, deterministic scripts, etc.

The existing `tj report --trim` flag writes an HTML report to `~/.cache/tokenjam/reports/` and opens it in the user's browser. `tj report --reuse` mirrors that pattern, plus emits a per-cluster Markdown sidecar that's directly copy-paste-usable.

## Acceptance criteria

1. **New CLI flag** `tj report --reuse [<agent_id>] [--since 30d]` in `tokenjam/cli/cmd_report.py`, mirroring the existing `--trim` flag in shape.

2. **Single combined output** written to `~/.cache/tokenjam/reports/` (overridable via `TOKENJAM_REPORT_DIR`):
   - `reuse-<timestamp>.html` — main HTML report, opened in the default browser via `webbrowser.open()`.
   - `reuse-<cluster-id>-<timestamp>.md` — one per cluster, the Markdown skeleton.

3. **HTML report content**:
   - Header: window range, total cluster count, total recoverable (both numbers).
   - One section per cluster, ranked by `cache_reuse_recoverable_usd` descending:
     - Cluster signature (tool sequence as a pretty-printed list).
     - Repetition count + average planning cost.
     - Both savings numbers (cache-reuse and script-replacement), framed per pricing mode via `core/framing.py`.
     - The skeleton rendered from the cluster's `skeleton_session_id` with variable slots highlighted as `{{slot_1}}`, `{{slot_2}}`, etc.
     - Top-3 example session IDs with deep links to `tj traces` output (anchor or local URL).
     - A relative link to the per-cluster Markdown sidecar.
     - Standard honesty caveat: "Structural analysis only — review the templates before reusing them."

4. **Skeleton rendering** lives in a new module `tokenjam/core/export/reuse_skeleton.py`:
   - `render_skeleton(plan_text: str, example_texts: list[str]) -> tuple[str, dict]` returning `(skeleton_with_slots, slot_map)`.
[...]

Issue #115, opened 2026-06-17
[feature] Add Reuse analyzer — detect repeated planning across sessions


anilmurty:
**Type:** Feature / backend. Fifth analyzer in the `tj optimize` family.

**Depends on:** #111 (the estimated-recoverable contract). Can land independently and retrofit if #111 hasn't merged yet.

## Background

Agents re-plan the same work constantly. A developer who runs "patch release" thirty times this month paid for thirty plans whose tool sequences and structure were identical — only the version string and date varied. Today no `tj optimize` analyzer names this specific waste pattern. Reuse does.

The analyzer detects clusters of sessions that share a **planning skeleton**: the structural shape of the model's first planning output. It quantifies the cost of re-buying that plan and produces an estimated-recoverable amount so the finding fits the existing analyzer contract.

This is the fifth analyzer, joining `downsize`, `cache`, `script`, `trim`. It self-registers under the existing registry pattern.

## Definitions

**Planning portion** of a session: the first LLM call in the session that precedes any tool call. Mechanically:
- Order spans within a session by `start_time`.
- The most recent LLM completion span before the first tool span is the planning call.
- If the session has no tool calls, the first LLM span is the planning call.
- If the session has no LLM spans, the session is skipped.

**Tool-sequence signature** of a session: the ordered tuple of tool names following the planning call.

**Prompt-prefix hash** (when `[capture] prompts = true`): SHA-256 (first 16 hex chars) of a variable-stripped prefix of the planning call's prompt content. The strip pass replaces digits, ISO dates, and posix paths with placeholders so "release v0.3.4 on 2026-06-15" and "release v0.3.5 on 2026-06-17" hash identically.

**Cluster**: group of sessions sharing the same `(tool_signature, prompt_prefix_hash)` pair. When prompt-prefix isn't available, the cluster key is the tool signature alone.

## Acceptance criteria

1. New file `tokenjam/core/optimize/analyzers/plan_reuse.py` registered via `@register("reuse")`. File name and registry name decoupled per CLAUDE.md Critical Rule 19.

2. New result dataclasses in `tokenjam/core/optimize/types.py`:
[...]

PR #143, opened 2026-06-18
Add Reuse analyzer (#115)

anilmurty:
Implements the **Reuse** analyzer — the fifth in the `tj optimize` family — per issue #115 and `.claude/reuse-{prd,architecture,savings-contract}.md`.

## What it does
Detects clusters of sessions that share a **planning skeleton** (the structural shape of the model's first planning call) and quantifies what reusing that skeleton would recover. Read-only against existing spans; no ingestion changes, no new deps, no new tables.

## Changes
- **`analyzers/plan_reuse.py`** → `@register("reuse")`. Planning call = first LLM span before any tool call. Single windowed SQL query + O(N) per-session walk. Tiered detection: tool-sequence signature always (Mode 1); prompt-prefix hash when `[capture] prompts = true` (Mode 2). Clustering behind a single `_cluster_sessions` seam for forward-compat.
- **`types.py`** — `ReuseCluster` / `ReuseFinding` + `REUSE_HONESTY_CAVEAT` / `REUSE_ESTIMATE_BASIS`. Two recoverable framings per cluster (cache-reuse `cost×(reps−1)`, script-replacement `cost×reps`); aggregate uses the conservative cache-reuse number.
- **Thresholds** (module-level constants): ≥3 reps, ≥200 avg planning tokens, ≥\$0.01.
- **Wiring** — `ANALYZER_ORDER` after `script`; `report_from_dict` round-trip constructor (`/api/v1/optimize` serializes generically); optimize package exports; plan-tier-aware `_render_reuse` in `cmd_optimize`.
- **Tests** — 17 unit tests in `test_optimize_reuse.py`.

## Note on the architecture doc
The doc's planning-call identification keyed on an `operation_name` column, which **doesn't exist** in this schema. Adapted to the codebase's actual LLM-vs-tool discriminator (`model IS NOT NULL` vs `tool_name IS NOT NULL`), matching `summarize_window` and the script analyzer.

## Acceptance criteria (#115)
All met: self-registration, dataclasses, ordering, single query, tiered detection, thresholds, savings math, `estimate_basis` containing "review", honesty language, CLI rendering, `findings["reuse"]` over the API, and the full unit-test matrix.

## Verification
- `ruff check` ✅  ·  `mypy` ✅  ·  `pytest tests/unit tests/synthetic tests/agents tests/integration` → green ✅
- Manual smoke: build_report → report_to_dict → report_from_dict round-trips; `_render_reuse` renders cleanly across api/subscription/local/unknown modes.

Part 1 of 2 for **TokenJam Reuse**. Part 2 (`tj report --reuse`, #116) stacks on this branch.


anilmurty (commented):

Implementation tracks all three spec docs (`.claude/reuse-prd.md`, `reuse-architecture.md`, `reuse-savings-contract.md`) faithfully. CI green on 3.10/3.11/3.12 + TS.

**Architecture conformance — all pass:**
- `analyzers/plan_reuse.py` registered via `@register("reuse")`; file/registry decoupled per Rule 19.
- `ReuseCluster` / `ReuseFinding` dataclasses match the spec field list exactly.
- `ANALYZER_ORDER` inserts `"reuse"` after `"script"` (`runner.py:36`).
- Single windowed SQL with `ORDER BY session_id, start_time` then O(N) Python walk — equivalent to the spec's `ROW_NUMBER()` and simpler.
- `_identify_planning_call`, `_tool_signature`, `_prompt_prefix_hash`, `_strip_variables`, `_cluster_key`, `_cluster_sessions` all match the spec.
- Thresholds at `plan_reuse.py:48-50` as module-level constants.

**Savings contract — all pass:**
- `cache_reuse = avg × (reps-1)` (`:252`), `script = avg × reps` (`:255`).
- Aggregate `estimated_recoverable_usd` uses the cache-reuse number (`:283`) — the conservative one per the contract.
- `estimate_basis = REUSE_ESTIMATE_BASIS` contains "review" (`types.py:28`).
- `confidence: Literal["heuristic"]` (`types.py:140`).
- All None cases covered (sub-threshold clusters, no LLM, capture off).

**Tiered detection — pass.** `capture_mode` set from `config.capture.prompts`; `_MODE1_HINT` populated only when off.

**CLI integration — pass.** `_render_reuse` uses `format_cost`/`format_tokens` from `core.framing`; pricing-mode branching matches the contract; honesty caveat printed verbatim from `c.caveat`.

**Tests — all pass.** 14 tests covering all required cases: 4 planning-ID variants, clustering thresholds, prompt-prefix invariance, savings math ($0.80/$1.00 verbatim), capture-off/on modes, empty window, basis contains "review".

**CLAUDE.md compliance — pass.** SQL uses `$N` positional params (parameterized); `utcnow()` via `runner._utcnow()`; no new deps; tests use factories (`make_llm_span`, `make_session`, `make_tool_span`); no embeddings; honesty language consistent.

**Out-of-scope confirmed absent:** no live cache, no learned matcher, no HTML/Markdown report (that's PR 2), no migrations.

### Nits (non-blocking)

1. **`plan_reuse.py:183`** — SQL builds `f"agent_id = ${len(params) + 1}"`. Index interpolation (not user data), and matches the existing pattern in `runner.py:summarize_window`. Safe, but a one-line comment explaining the positional-index pattern would help future readers.
2. **`_PATH_RE` is greedy** — could swallow protocol fragments like `https://...` partially. Low impact (200-char prefix hash) but worth a comment.
3. **`example_session_ids` ordering** relies on `plan_start_time` being comparable — DuckDB returns datetime, so fine; a type comment would help.
4. **`tool_signature` round-trip via `asdict()`** is correctly handled in `runner.py:_build_finding_constructors` (list→tuple restoration is explicit). Good catch by the author — no change needed.
5. **No test for `cluster_id` determinism across runs.** PR 2 needs this property to keep Markdown sidecar filenames stable, so worth adding either in this PR or as the first test in PR 2.

Ship it.

anilmurty:
Thanks for the review! Addressed the nits in d162a0b (no behavior change):

1. **Positional-`$N` SQL pattern** — added a comment clarifying the f-string interpolates only the placeholder *index*, never user data (parameterized, matches `summarize_window`).
2. **`_PATH_RE` greediness** — commented that it's intentional; both sides of a comparison normalize identically and it only feeds the coarse 200-char prefix hash.
3. **`plan_start_time` type** — noted it's a tz-aware DuckDB `TIMESTAMPTZ` datetime, hence directly comparable for recency ordering.
4. — no change needed (as you noted).
5. **`cluster_id` determinism** — added `test_cluster_id_is_deterministic_across_runs`: asserts stability across re-runs and equality with the pure `_cluster_key`. This locks the property PR 2's idempotent Markdown sidecar filenames depend on.

ruff + mypy clean; reuse suite now 18 tests.

Issue #105, opened 2026-06-16
tj report --trim: Rich consumes [capture] in the disabled-analyzer hint


anilmurty:
When `capture.prompts` is false and you run `tj report --trim`, the hint string in `cmd_report.py` contains `[capture] prompts = true` but Rich consumes `[capture]` as a style tag during rendering. Output looks like:

```
Trim analyzer not ready:
Enable ` prompts = true` in tj.toml and let the daemon collect a window of data before re-running this analyzer. Trim needs captured prompt text to score.
```

Note the missing `[capture]` between the opening backtick and 'prompts'.

## Same fix pattern we already use elsewhere

`tj optimize trim` renders the same hint correctly because `cmd_optimize.py` runs it through `rich.markup.escape()` (via `_rich_escape`). The `cmd_report.py` path doesn't.

## Repro

```bash
# capture.prompts must be false in config (default)
tj report --trim --no-open
# Observe: hint string missing '[capture]'
```

## Fix

In `tokenjam/cli/cmd_report.py`, escape the hint string before passing it to `console.print()`, same way `cmd_optimize.py` does. Roughly:

```python
from rich.markup import escape as _rich_escape
console.print(f"Trim analyzer not ready:\n{_rich_escape(finding.hint)}")
```

## Impact

Cosmetic — the hint still points at the right config field via its prose context. But it confuses anyone copy-pasting the literal command. Worth fixing before the launch tweet drives traffic into the docs.

## Out of scope

A broader audit of Rich markup escape coverage across the CLI. The pattern of 'prose hint strings that contain TOML section names' is reused in several places and could grow — worth a single helper function someday.
EOF
)

PR #160, opened 2026-06-19
fix: escape [budget.<provider>] in onboard plan message (#157)

### LINKED ISSUE #157: tj onboard: plan-written message drops the section name (unescaped Rich markup)
## Summary

In 0.4.0, the success message `tj onboard` prints after writing a plan tier shows an empty
parenthetical instead of the config section name:

```
✓ Plan tier: api (written to )
```

It should read `✓ Plan tier: api (written to [budget.anthropic])`. The plan **is** written correctly —
the config contains `[budget.anthropic]` with `plan = "api"`. Only the confirmation message is
missing the bracketed section name. **Cosmetic only.**

## Root cause

`tokenjam/cli/cmd_onboard.py:160-161`:

```python
console.print(f"[green]✓[/green] Plan tier: "
              f"[bold]{plan_tier}[/bold] (written to [budget.{plan_provider}])")
```

`console.print` renders Rich markup, and Rich treats `[budget.anthropic]` as a style tag, so it's
consumed/stripped → `(written to )`. This is exactly the case CLAUDE.md warns about: *"Rich
source-section strings (`[budget.anthropic]`, `[[alerts.channels]]`) must be passed through
`rich.markup.escape()`."* The same file handles this elsewhere (comments at lines 656 and 892), but
this line — new code that arrived with the #4 fix — missed it.

## Steps to reproduce

1. `tj onboard --plan api --no-daemon` (answer the daily-budget prompt with `0`)
2. Observe: `✓ Plan tier: api (written to )` — empty parens

## Fix

Escape the section name before printing:

```python
from rich.markup import escape
...
console.print(f"[green]✓[/green] Plan tier: [bold]{plan_tier}[/bold] "
              f"(written to {escape(f'[budget.{plan_provider}]')})")
```

## Environment
- OS: macOS 14.4
- Python: 3.14
- tj version (`tj --version`): 0.4.0

ashwmu:
Fixes #157.

## Problem
`tj onboard` printed `✓ Plan tier: api (written to )` — the `[budget.<provider>]`
section header was missing because `console.print` (Rich) treated it as a markup
tag and stripped it.

## Fix
Escape the section header with `rich.markup.escape()` before printing, so it
renders literally as `(written to [budget.anthropic])`. The plan was always
written correctly to the config; only the confirmation message was affected.

Regression introduced by the #4 fix (plain `tj onboard` honoring `--plan`).

## Tests
- New regression test `test_plan_message_shows_section_name` — asserts the
  confirmation keeps the section header (verified to fail on the buggy code and
  pass on the fix).
- ruff ✅ · mypy ✅ · onboard suite (25 tests) ✅

### DIFF
--- a/tests/unit/test_onboard_plan.py
+++ b/tests/unit/test_onboard_plan.py
@@ -53,6 +53,16 @@ def test_plain_onboard_no_plan_noninteractive_writes_no_plan(tmp_path):
     assert "plan =" not in cfg
 
 
+def test_plan_message_shows_section_name(tmp_path):
+    # Regression for #157: the "written to [budget.<provider>]" confirmation must
+    # NOT have its TOML section header eaten by Rich markup parsing (which left
+    # the message reading "(written to )").
+    res, _ = _onboard(tmp_path, "--plan", "api")
+    assert res.exit_code == 0, res.output
+    assert "[budget.anthropic]" in res.output, res.output
+    assert "(written to )" not in res.output
+
+
 def test_plain_onboard_writes_valid_toml(tmp_path):
     import sys
     if sys.version_info >= (3, 11):
--- a/tokenjam/cli/cmd_onboard.py
+++ b/tokenjam/cli/cmd_onboard.py
@@ -9,6 +9,7 @@ import sys
 from pathlib import Path
 
 import click
+from rich.markup import escape
 
 from tokenjam.core.config import find_config_file
 from tokenjam.utils.formatting import console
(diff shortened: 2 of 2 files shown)


anilmurty (commented):
## Review: ✅ Approve

Tight, surgical fix. CI green on 3.10/3.11/3.12 + TS.

**Root-cause analysis is exactly right.** Rich's `console.print` consumes `[budget.<provider>]` as a markup tag, so the surrounding parens render with nothing inside. CLAUDE.md prescribes `rich.markup.escape()` for TOML section headers, and the same file already does this at `cmd_onboard.py:656` and `:892` — this was a missed spot from when PR #148 introduced the plain-path `--plan` confirmation line.

**The fix is the right size:**
- Single new import (`from rich.markup import escape`)
- Escape just the bracketed section header, not the whole f-string (so the `[green]` / `[bold]` markup still renders)
- Inline comment cites #157 for future readers
- Net +5/-1 in `cmd_onboard.py`

**Test is well-built.** `test_plan_message_shows_section_name` uses the existing `_onboard()` helper to stay consistent with the file's idioms, and asserts both:
- ✅ `[budget.anthropic]` is **in** the output (positive — fix worked)
- ✅ `(written to )` is **not** in the output (negative — buggy state gone)

That negative assertion is the load-bearing piece — it locks in the regression guard against the exact symptom.

**Other quality details I noticed:**
- `—` (em-dash) used in the source-code comment instead of a literal `—` — clean Unicode hygiene, matches the file's existing style.
- The escape is scoped to just the `plan_section` variable, not slapped over the whole `console.print` call. Surgical.

Thanks for the careful repro chain — issue #157's reproduction steps and the call-out of the prior `escape()` precedents elsewhere in `cmd_onboard.py` made this a 30-second decision. Reading it back, the fix is structurally identical to PR #107 (which closed #105, a similar Rich-escape miss in `tj report --trim`) — this is exactly the pattern we want to lock in.

Good to merge.

Issue #114, opened 2026-06-17
[feature] New Overview dashboard, Optimize detail tab, and TokenJam Lens brand pass


anilmurty:
**Type:** Feature / UI. The marquee.

**Depends on:** #110, #111, #112, #113.

## Background

The local web UI today is six screens of telemetry tables. Each is good at its job individually, but a user opening the dashboard has no triage surface — no answer to "what should I pay attention to?" We're adding two new screens and renaming the UI:

- **Overview** — a new front door that surfaces the three things a user cares about (current spend, what looks recoverable, what's wrong) with click-through into the existing screens.
- **Optimize** — a detail screen for analyzer findings, paired with Overview's "recoverable" band. Today `tj optimize` output has no UI surface.
- **TokenJam Lens** — the local UI gets a name so it's recognizable on the marketing site.

Overview and Status answer different questions and stay separate. Status keeps its job (every agent at the same altitude); Overview is the opinionated triage view that highlights the agents and findings that matter.

## Acceptance criteria — Overview screen

1. New screen at `#/overview`, becomes the default landing route (`/` → Overview, not Status).
2. Three bands, top to bottom:
   - **Cost (hero)**: total spend for the selected window, trend chip vs previous period, the spend-over-time chart from #113 embedded compactly, simple linear run-rate projection. Window toggle: 24h / 7d / 30d / 90d.
   - **Recoverable (the analyzer rollup)**: one tile per analyzer that returns `estimated_recoverable_usd > 0` for the window, sourced from `/api/v1/optimize`. Tile shows: analyzer name, estimated recoverable amount (framed per pricing mode), one-line "what to do" hint, a visible "estimated" tag with hover surfacing `estimate_basis`. Analyzers in a "not-ready" state (capture off, no data) render in a muted tile with an enablement hint instead of a number. `cache-recommend` and `budget-projection` are excluded by virtue of not emitting `estimated_recoverable_usd`.
   - **Health (at a glance)**: four small tiles — Alerts (count by severity), Drift (count of agents with signal), Budget (count over 80% of ceiling or projected to exceed), Activity (last few sessions with success/failure indicator).
3. Every tile is a clickable link that deep-links into a detail screen with relevant filters pre-applied via URL:
   - Cost hero → `#/cost?since=<window>`
   - Recoverable tile → `#/optimize?finding=<name>&since=<window>`
   - Alerts tile → `#/alerts?unread=true`
   - Drift tile → `#/drift`
[...]

Issue #139, opened 2026-06-18
[follow-up] buildCostSeries 5000-bucket guard silently empties very long hourly windows


anilmurty:
Follow-up from PR #137 review (#133).

`buildCostSeries` in `tokenjam/ui/index.html` guards against pathological grids
with `if (xs.length > 5000) return null`. That's correct defense today (the
window picker tops out at 90d, and 90d daily = 90 buckets), but if longer
windows or finer (hourly) bucketing for multi-day ranges ever ship, a long
hourly window (~208d) would silently cross 5000 buckets and render the empty
state instead of a chart.

### Acceptance (when longer windows/finer bucketing land)
- Long windows degrade gracefully (coarsen the bucket, e.g. day→week) rather
  than silently returning null.
- If a cap is genuinely hit, surface a note ("showing daily buckets for this
  range") rather than an empty chart — no silent truncation (CLAUDE.md spirit).

Not reachable with the current 24h/7d/30d/90d picker; filing so it isn't lost.

PR #161, opened 2026-06-19
fix(#124): make the daemon DB layer safe for concurrent API requests

### LINKED ISSUE #124: [follow-up] Make the daemon DB layer safe for concurrent API requests
Surfaced while building the Lens Overview (#114).

The daemon serves the REST API over a single DuckDB connection. The async
routes run on the event loop (serialized), but several read routes are sync
`def` (e.g. `/optimize`, `/cost/compare`) and run in Starlette's threadpool —
so concurrent requests can hit the same DuckDB connection from multiple threads
and 500 (DuckDB connections aren't safe for concurrent use).

This is latent (any two overlapping requests — UI poll + CLI, or two tabs — can
trigger it), but the Overview made it easy to hit because it fetched ~6
endpoints at once. #114 worked around it by having the Overview fetch
**sequentially**; the underlying daemon limitation remains.

### Options
- Give each request its own `conn.cursor()` (DuckDB cursors are safe for
  concurrent use from different threads), or
- Serialize DB access behind a lock, or
- Make the read routes `async def` so they share the event loop (simplest;
  acceptable for a local single-user daemon, but blocks the loop during
  `build_report`).

### Acceptance
Firing the Overview's endpoint set concurrently against `tj serve` returns 200
for all, repeatably.

anilmurty:
Closes #124 (follow-up surfaced while building the Lens Overview, #114).

## Problem
The daemon serves the REST API over a **single** DuckDB connection. The sync (`def`) read routes — `/optimize` and `/cost/compare` — run in Starlette's threadpool, so two overlapping requests (the Overview's fan-out, a UI poll + CLI, or two browser tabs) could drive the same DuckDB connection from multiple threads. A single DuckDB connection object is **not** safe for concurrent use: overlapping execute/fetch corrupts the shared result set (a reader gets `None` rows) and can abort the process (SIGABRT). #114 worked around it by making the Overview fetch sequentially; this fixes the root cause.

## Fix
`DuckDBBackend.conn` is now a **per-thread cursor** over one shared database:

- DuckDB cursors created via `connect().cursor()` are independent connections that **are** safe to use concurrently from different threads (the DuckDB-recommended multi-threading pattern), while still sharing one database — so a write on one thread is visible to reads on another.
- Each thread lazily gets and reuses its own cursor via `threading.local`. Single-threaded callers (tests, the CLI) always see one cursor, so behavior is unchanged.
- **No call-site changes** — every existing `db.conn.execute(...)` becomes thread-safe transparently. `InMemoryBackend` gets the same treatment (in-memory cursors share the same DB, so cross-thread visibility holds for the threadpool-backed integration tests).

I verified the underlying behavior empirically before implementing: a single shared connection hammered from 8+ threads SIGABRTs; per-thread cursors run clean (0 errors) with cross-cursor visibility and concurrent read/write.

## Test
`test_overview_endpoint_set_survives_concurrent_requests` fires the Overview endpoint set (including both sync routes) concurrently over several rounds and asserts every response is 200 — the issue's acceptance criterion. **Confirmed it fails on the pre-fix single-connection code** (`TypeError: 'NoneType' object is not subscriptable` from a corrupted mid-fetch result) and passes with the per-thread cursor.

Full CI set green locally (762 passed), mypy + ruff clean on `db.py`.

> Built in an isolated `git worktree` off `main`.

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -1,6 +1,8 @@
 """Integration tests for the REST API using httpx.AsyncClient + ASGITransport."""
 from __future__ import annotations
 
+import asyncio
+
 import pytest
 import httpx
 
@@ -331,6 +333,27 @@ async def test_cost_cycle_block_honors_provider_cycle_start_day(db):
     async with httpx.AsyncClient(transport=transport, base_url="http://test") as c:
         cycle = (await c.get("/api/v1/cost")).json()["cycle"]
     assert cycle["start_day"] == 15
+
+
+async def test_overview_endpoint_set_survives_concurrent_requests(client):
+    """#124: the daemon's sync (`def`) read routes (`/optimize`,
+    `/cost/compare`) run in Starlette's threadpool, so concurrent requests reach
+    DuckDB from several threads at once. A single shared connection aborts the
+    process under that load; per-thread cursors (DuckDBBackend.conn) make it
+    safe. Fire the Overview's endpoint set concurrently, repeatedly, and assert
+    every response is 200 — the issue's acceptance criterion."""
+    await _ingest_sample_span(client)
+    endpoints = [
+        "/api/v1/cost?since=7d",
+        "/api/v1/optimize?fast=true&since=30d",
+        "/api/v1/cost/compare?since=7d&compare=previous",
+        "/api/v1/traces",
+        "/api/v1/budget",
+    ]
+    for _ in range(5):  # several rounds so threadpool overlap is likely
+        results = await asyncio.gather(*(client.get(u) for u in endpoints * 3))
+        bad = [(str(r.url), r.status_code) for r in results if r.status_code != 200]
+        assert not bad, bad
 
 
 async def test_optimize_response_includes_framing_block(client):
--- a/tokenjam/core/db.py
+++ b/tokenjam/core/db.py
@@ -5,6 +5,7 @@ and migration runner. DuckDB only — never import sqlite3.
 from __future__ import annotations
 
 import json
+import threading
 from datetime import date, datetime
 from pathlib import Path
 from typing import Protocol, runtime_checkable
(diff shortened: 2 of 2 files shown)


anilmurty (commented):
## Review: ✅ Approve

CI green on 3.10/3.11/3.12 + TS. **Smaller, cleaner solution than the design comment prescribed.** Worth saying so explicitly.

## On the architectural deviation from #124's design comment

The design comment specified **Option A** — `threading.Lock()` around writes + a `read_only_connection()` context manager for reads + auditing ~10 route files to switch callers over. This PR uses a different pattern: per-thread DuckDB cursors via `threading.local`. **It's better than what I directed.** Reasoning:

1. **Transparent to all callers.** `db.conn` becomes a property that lazily allocates a cursor per thread. Every `db.conn.execute(...)` call site keeps working with no edits. Zero route-file diff vs. the ~10-file audit my prescription would have required.

2. **DuckDB's canonical thread-safety pattern.** Cursors from `connection.cursor()` are independent connection-like objects over the same database — explicitly documented as the way to do multi-threaded DuckDB.

3. **Better cross-thread consistency.** All cursors see the same database state instantly; the read-only-file-connection approach in my prescription would have had a (small but real) read-after-write lag.

4. **`InMemoryBackend` inherits the fix.** Cursors of an in-memory connection share the in-memory database, so cross-thread visibility works for tests too — the agent's comment calls this out and the test confirms it.

5. **Single-threaded callers behaviorally identical to pre-PR.** `getattr(self._local, "cursor", None)` returns the same cursor every time when only one thread asks — so tests and the CLI see no change. Important property.

Net diff: **+59/−6 across 2 files** vs. ~150 LOC across ~12 files for the prescribed approach. Same end-state, much smaller blast radius. Good judgment in evaluating the prescription and choosing the cleaner pattern.

## Verification quality

`test_overview_endpoint_set_survives_concurrent_requests` is well-built — fires **15 endpoints × 5 rounds = 75 concurrent operations** through `asyncio.gather` against the Overview's actual endpoint set, asserts every response is 200. The pre-fix behavior (SIGABRT under 8-thread contention) is empirically confirmed in the PR description.

The test correctly targets the *sync* (`def`) read routes (`/optimize`, `/cost/compare`) that get scheduled into Starlette's threadpool — those were the actual contention path. Calling out the threadpool mechanics in the docstring is a nice touch for the next reader.

## One non-blocking miss — the UI revert

The original design directive's step 4 was *"Revert the Overview's sequential-fetch workaround in `tokenjam/ui/index.html`."* The backend is now safe for parallel fetching, but the UI is still serializing — performance left on the table, not a correctness issue. Either:

- Send a one-line follow-up PR removing the workaround, or
- Fold it into your next PR (#139 — `buildCostSeries`, which is also in `index.html`)

Whichever you prefer. Both work; the second is slightly more economical since you'll already be in that file.

Good to merge.

anilmurty:
Thanks for the thorough review. On the UI revert (the #114 sequential-fetch workaround): I'll **fold it into #139**, per your economical option — #139 is already in `index.html`, and the workaround was the mitigation for exactly this bug, so removing it alongside the next `index.html` change keeps this PR focused on the backend fix.

Concretely, in #139 I'll swap the six sequential `await api(...)` calls in `OverviewView.load` for a single `Promise.all` (keeping each call's `.catch` fallback so one failing panel doesn't blank the whole Overview, and letting `/cost` reject into the outer error state as it does today), and add a regression guard asserting the parallel pattern (no test currently pins the sequential one). It's naturally gated on this PR merging first, since both touch `index.html`.

PR #163, opened 2026-06-19
fix(#139): coarsen long cost-chart windows instead of silently emptying (+ Overview parallel fetch)

### LINKED ISSUE #139: [follow-up] buildCostSeries 5000-bucket guard silently empties very long hourly windows

anilmurty:
Closes #139 (follow-up from PR #137 / #133 review). Also bundles the #124 review follow-up (Overview parallel-fetch revert) — same file, same theme.

## #139 — graceful bucket coarsening
`buildCostSeries` guarded pathological grids with `if (xs.length > 5000) return null`, which renders the empty state with **no explanation** if a long window (or future finer/hourly bucketing for multi-day ranges) ever crosses 5000 buckets. That's silent truncation.

**Fix (option a — graceful degradation):** coarsen the bucket up a ladder (`hour → day → week`) until the grid fits `MAX_BUCKETS`, computed closed-form (no oversized array is ever built). When coarsening happens, the chart surfaces a note — *"Showing &lt;bucket&gt; buckets — this range is too long for &lt;requested&gt;ly detail"* — so it's never silent. Genuinely empty windows (no data) still return `null` → the existing "No spend in this window" state.

Verified in node:
| input | result |
|---|---|
| 7d, daily | no coarsening, 8 points |
| 300d, hourly-requested (>5000 hourly) | coarsens to **day**, 301 points, value preserved |
| extreme (decades), hourly | falls to **week**, note fires — never empty |
| no bounds + no rows | `null` (empty state) |

## #124 follow-up — Overview parallel fetch (per the #161 review)
The #114 serial-fetch workaround is reverted now that the daemon DB layer is concurrency-safe (#124). `OverviewView.load` fans out via `Promise.all`. The error handling is **deliberately asymmetric and must not be unified** (flagged in the #161 review):
- `/cost` is **load-bearing** — no `.catch`, so its failure rejects the `Promise.all` and surfaces the error state (the page can't render without spend data).
- the other five panels each keep a `.catch` fallback so one failing panel renders empty rather than blanking the whole Overview.

## Docs + tests
- `CLAUDE.md`: REST API concurrency note and Web UI polling note now describe the parallel fetch + the asymmetric error contract (was: "fetches sequentially").
- `tests/unit/test_lens_ui_regression.py`: guards for the coarsening behavior (silent guard gone, ladder + note present) and the parallel/asymmetric fetch pattern (asserts `/cost` has **no** `.catch` while the other five do).

Full CI set green locally (766 passed), UI passes `node --check`, offline-UI test green.

> Built in an isolated `git worktree` off `main`.


anilmurty (commented):
## Review: ✅ Approve

Both items in one focused PR — exactly the bundling the #161 review asked for. Tight diff, strong tests, accurate doc updates.

## #139 — graceful bucket coarsening

Three smart implementation choices:

1. **`_BUCKET_LADDER = [hour, day, week]`** — picks the finest bucket starting from `reqBucket` whose grid fits `MAX_BUCKETS`. Genuinely empty windows (`!rows.length` and no bounds) still return `null` so the existing "No spend in this window" empty state is preserved — only the silent-on-pathological-range case is fixed. Good preservation of the existing empty-data path.

2. **Closed-form count: `(Math.floor(end / s) - Math.floor(start / s)) + 1`** — never allocates an oversized array. A naïve `for` loop would have built a 100K-entry array just to discard it. The closed-form math sidesteps that — important when someone runs a multi-year hourly query.

3. **User-facing note** when coarsening fires: *"Showing day buckets — this range is too long for hourly detail."* The series result stashes `requestedBucket` so both bucket names are available to the renderer. Honest about what changed; no silent truncation.

## #124 follow-up — Overview parallel fetch

The asymmetric error contract from the #161 review is preserved exactly:

```js
const [cost, compare, opt, al, drift, traces] = await Promise.all([
  api('/cost', { since, group_by: 'day' }),                              // ← no .catch, load-bearing
  api('/cost/compare', { since, compare: 'previous' }).catch(() => null),
  api('/optimize', { since, fast: 'true' }).catch(() => null),
  api('/alerts', { since, unread: 'true' }).catch(() => ({ alerts: [] })),
  api('/drift').catch(() => ({ agents: [] })),
  api('/traces', { since, limit: 6 }).catch(() => ({ traces: [] })),
]);
```

`/cost` is bare → its failure rejects the `Promise.all` and surfaces the outer error state. The other five panels each carry their own `.catch` fallback. The inline comment explicitly documents this is intentional and "must NOT be unified (#124 review)" — exactly the framing the review asked for.

## Tests — well-built regression guards

Three new tests, each balancing positive presence and negative absence:

- **`test_cost_series_coarsens_not_silently_empty`** — asserts `"xs.length > 5000) return null"` is gone, the `_BUCKET_LADDER` + `MAX_BUCKETS` constants are present, the user-facing "Showing X buckets" copy is present. The literal-substring assertion on the old guard is robust.
- **`test_overview_fetches_in_parallel`** — old serial-fetch comment is gone, `await Promise.all([` is present.
- **`test_overview_error_handling_is_asymmetric`** — asserts `/cost` has **no** `.catch` (positive guard for the asymmetry) **and** each of the 5 fallback endpoints has its expected `.catch(...)` clause. Locked in both contracts in one test.

## CLAUDE.md updates

Both touched paragraphs accurately reflect the new state with the explicit guidance — *"do not collapse `conn` back to one shared connection object"* in the API routes section, *"Don't unify them"* on the polling section. Future agents reading the file know not to undo either contract.

## Tiny non-blocking observation

The `${series.requestedBucket}ly` interpolation reads naturally for the current ladder (`"hourly"`, `"daily"`, `"weekly"`). If anyone ever extends the ladder with a non-clean-suffix name (e.g. `"5min"` → `"5minly"`), the copy would need adjustment. Pure forward-looking nit; nothing to do here.

Ship it.

Issue #141, opened 2026-06-18
v0.3.5 post-release smoke test — 17 issues (2 High, 3 Medium, 1 Low–Med, 11 Low) + 1 UX suggestion


ashwmu:
## Summary

Ran the post-release smoke test (`tests/manual-new-release-tests.md`) against the published **0.3.5**
release (PyPI + tag `v0.3.5`) on macOS (Homebrew Python, **pipx** install). **All 8 core sections
completed and functionally passed** — the published artifact installs and works end-to-end. The
findings below are a mix of real bugs (2 High, 1 Medium) and doc/UX issues surfaced while following
the doc.

**Environment:** macOS 23.4 · Python 3.14.5 · `pipx install tokenjam` (CLI at `~/.local/bin/tj`) ·
fresh clone of `Metabuilder-Labs/tokenjam` @ `v0.3.5`.

**Section pass status:** 1 ✅ · 2 ✅ · 3 ✅ · 4 ✅ · 5 ✅ · 6 ✅ · 7 ✅ · 8 ✅ · (Claude Code / Codex
integration sections not run).

### Severity counts
- 🔴 **High (1):** #6
- 🟡 **Medium (3):** #1, #5, #17
- 🟠 **Low–Med (2):** #4, #7
- ⚪ **Low (11):** #2, #3, #8, #9, #10, #11, #12, #13, #14, #15, #16
- 💡 **UX suggestion (1, not a bug):** #18

---

### #1 (Medium) — `pip3 install --upgrade tokenjam` hard-fails (PEP 668); pipx is the working path
The doc's Step 1 install command aborts with `error: externally-managed-environment` on macOS/Homebrew
Python — nothing installs, and `tj --version` stays stale. Reproduced from a fully clean slate.
CLAUDE.md already recommends pipx, so the smoke-test doc contradicts project guidance.
**Fix:** change Step 1 to `pipx install/upgrade tokenjam`; update the cleanup block to cover pipx.

### #2 (Low) — `tj uninstall` hint hardcodes `pip uninstall` and doesn't remove the package
`tj uninstall --yes` ends with `To remove the package itself, run: pip uninstall tokenjam` — wrong for
a pipx install. It also only removes data/config/MCP, **not** the package, which misled testing
(`pipx upgrade` later reported "already at latest").
**Fix:** detect install method; print `pipx uninstall tokenjam` / `pip uninstall tokenjam` and clarify
that `tj uninstall` does not remove the package.

---

[...]

PR #164, opened 2026-06-19
Show active (compute) time alongside wall-clock in Status (#147)

### LINKED ISSUE #147: [enhancement] Status 'Duration' is wall-clock; show active time too
Split out from the v0.3.5 smoke test (#141, finding #18 — UX suggestion, not a bug).

The Status tile shows a Claude Code session `Duration: 3087m` (~51h). This is
correct by definition (`duration_seconds = ended_at − started_at`,
`models.py`) — Claude Code sessions resume across days — but a 51h "Duration"
reads as a runaway session.

Active time (Σ span `duration_ms`) already exists in the data.

### Suggestion
Show active time alongside the wall-clock value (or relabel the wall-clock one,
e.g. "Span" / "Elapsed"), so a resumed multi-day session doesn't look like a
51-hour run.

Scope: Status tile rendering (`tokenjam/ui/index.html`) + possibly the
`/api/v1/status` payload if active time isn't already exposed.

anilmurty:
Closes #147.

## Problem
The Status tile showed only wall-clock `Duration` (`ended_at − started_at`). Claude Code sessions resume across days, so a tile reading `Duration: 3087m` (~51h) looks like a runaway session when the agent was actually active for minutes. Active time (Σ span `duration_ms`) already exists in the data — it just wasn't surfaced.

## Fix
Surface **active (compute) time** alongside the wall-clock value across every Status surface:

- **`core/db.py`** — `session_active_seconds(conn, session_id)`: sum of span durations in seconds; returns `None` when the session has no timed spans (so callers omit rather than show a misleading `0`).
- **`/api/v1/status`** — payload now carries `active_seconds` next to `duration_seconds`.
- **UI status tile** (`ui/index.html`) — new **Active** row (compute time) + the wall-clock row relabeled **Elapsed**, each with an explanatory tooltip. Added `fmtDurLong` so multi-day spans render `2d 3h` instead of `3087m`.
- **CLI `tj status`** — replaced the misleading inline `(3087m 0s)` with a `Duration: active 12m 3s · elapsed 2d 3h` row; both fields included in `--json`.

Example CLI output for a resumed session (12m active, 51h elapsed):
```
● claude-code   active
  ...
  Duration:       active 12m 3s · elapsed 2d 3h
```

## Scope note
The issue scoped UI + API; I also applied it to `tj status` (CLI) since it had the identical defect. MCP `get_status` proxies the API and maps explicit fields, so it's unaffected (the new field is simply not mapped) — left out of scope.

## Verification
- `ruff check tokenjam/` ✅ · `mypy` ✅ · full suite **768 passed** ✅
- Tests added: core helper unit tests (sum / no-spans / unknown-session), `/api/v1/status` `active_seconds`, CLI `--json` fields, and a Lens UI regression pinning the Active/Elapsed labels + `fmtDurLong` (and asserting the bare "Duration" label is gone).
- UI offline regression (`test_ui_offline.py`) still green — no external loads added.

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -540,6 +540,26 @@ async def test_status_and_traces_agree_on_agent_ids(client):
     assert trace_agents == status_agents
 
 
+async def test_status_exposes_active_seconds(client, db):
+    """Status payload carries active (compute) time alongside wall-clock (#147)."""
+    from tests.factories import make_llm_span, make_session
+
+    db.upsert_session(make_session(agent_id="a1", session_id="s1", status="completed"))
+    for ms in (1000.0, 2000.0, 3000.0):
+        sp = make_llm_span(agent_id="a1", duration_ms=ms)
+        sp.session_id = "s1"
+        db.insert_span(sp)
+
+    resp = await client.get("/api/v1/status")
+    assert resp.status_code == 200
+    agents = {a["agent_id"]: a for a in resp.json()["agents"]}
+    assert "a1" in agents
+    a = agents["a1"]
+    assert "active_seconds" in a and "duration_seconds" in a
+    # 6000 ms of spans → 6.0 s active; distinct from wall-clock duration_seconds.
+    assert a["active_seconds"] == pytest.approx(6.0)
+
+
 # ── Budget ─────────────────────────────────────────────────────────────────
 
 async def test_post_budget_zero_clears_limit(db):
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -97,6 +97,21 @@ def test_status_exits_1_when_active_alerts(runner, db, config):
     assert data["has_active_alerts"] is True
 
 
+def test_status_json_includes_active_and_elapsed(runner, db, config):
+    """Status JSON carries active (compute) time + wall-clock elapsed (#147)."""
+    session = _seed_agent_and_session(db)
+    for ms in (1500.0, 2500.0):
+        sp = make_llm_span(agent_id="test-agent", duration_ms=ms)
+        sp.session_id = session.session_id
+        db.insert_span(sp)
+
+    result = _invoke(runner, db, config, ["status", "--json"])
+    assert result.exit_code == 0
+    a = json.loads(result.output)["agents"][0]
+    assert "active_seconds" in a and "duration_seconds" in a
+    assert a["active_seconds"] == 4.0   # 4000 ms of spans
+
+
 # -- traces tests --
 
 def test_traces_json_output_is_valid_json(runner, db, config):
--- a/tests/unit/test_db_helpers.py
+++ b/tests/unit/test_db_helpers.py
@@ -1,0 +1,35 @@
+"""Unit tests for small db.py query helpers."""
+from __future__ import annotations
+
+import pytest
+
+from tokenjam.core.db import InMemoryBackend, session_active_seconds
+from tests.factories import make_llm_span, make_session
+
+
+@pytest.fixture
+def db():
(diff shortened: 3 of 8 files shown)


anilmurty (commented):
## Review: ✅ Approve pending rebase + CI

Implementation is sound. **Please rebase first** — PR #163 merged a few minutes ago and touched both `ui/index.html` and `tests/unit/test_lens_ui_regression.py`, the same files this PR edits. CI hasn't run yet because of the conflict state; once rebased and re-pushed, CI should go green and this is ready to merge.

## Strengths

**Honest semantic split.** `Active` = compute work (Σ span durations); `Elapsed` = wall-clock. The UI tooltip copy explicitly explains *"Resumed Claude Code sessions span days, so this can be far larger than Active time."* That's exactly the framing #147 (and #141's finding #18) asked for — no more "3087m" reading as a runaway session.

**`None` not `0` when there are no spans.** `session_active_seconds()` returns `None` for the missing-data case so callers can omit the field rather than display a misleading zero. The docstring explicitly justifies this — good honesty discipline.

**Layered test coverage** — every layer doing a different job, no redundancy:
- `tests/unit/test_db_helpers.py` (NEW) — the SQL helper itself, including "no spans" and "unknown session_id" branches
- `tests/integration/test_api.py` — API contract (`active_seconds` in payload, distinct from `duration_seconds`)
- `tests/integration/test_cli.py` — CLI JSON shape carries both fields
- `tests/unit/test_lens_ui_regression.py` — locks `Active`/`Elapsed` labels in, asserts the misleading bare `Duration` label is gone

**`fmtDurLong` coarse formatter does its job.** `2d 3h` instead of `3087m` for resumed sessions. The threshold (≥1h shows hours-up; otherwise minutes-and-seconds) means a real 12-minute Active span still renders fine-grained while a 51-hour Elapsed span reads naturally.

**MCP was correctly left out of scope.** `get_status` proxies the API and reads explicit fields — the new `active_seconds` field passes through transparently. Good situational awareness.

**Built in an isolated worktree** (`/tmp/tj-147-wt`). CLAUDE.md Rule 20 / Working-with-Concurrent-Agents guidance followed.

## Two non-blocking nits

1. **CLI label could read `Time:` for parity.** The CLI now renders `Duration:    active 12m 3s · elapsed 2d 3h`. Both are durations of different kinds, so the label is technically fine — but `Time:` would be slightly more honest given the line shows two semantically-distinct measures. Pure copy nit.

2. **JS `fmtDurLong` vs Python `_fmt_dur(coarse=True)`.** Two helpers doing the same conceptual job. JS could mirror the Python pattern (`fmtDur(ms, { coarse: true })` instead of a separate function). Minor symmetry thing, not blocking.

## To merge

- Rebase / merge `main` into the branch (#163's `ui/index.html` and `test_lens_ui_regression.py` changes are now in main and will overlap)
- Re-push, wait for CI
- The conflict resolution should be mechanical — both PRs only add new functions/tests to those files, not modify existing lines

Ready when CI's green.

anilmurty:
Rebased onto latest `main` (now atop #163) and force-pushed.

The only conflict was in `tests/unit/test_lens_ui_regression.py` — both PRs appended new test functions at the same spot. Resolved by keeping both sets (#139/#124 tests + the #147 Active/Elapsed test). `ui/index.html` auto-merged cleanly (both only add new functions: my `fmtDurLong` + status-tile rows coexist with #163's `_BUCKET_LADDER`/coarsening).

ruff clean; full suite **772 passed** (includes #163's tests). Branch now sits directly on top of `origin/main` — should be a clean merge.

(Left the non-blocking JS/Python helper-symmetry nit as-is per your "not blocking".)

Issue #138, opened 2026-06-18
[follow-up] Run-rate cycle should honor [budget.<provider>] cycle_start_day, not the calendar month


anilmurty:
Follow-up from PR #137 review (#134).

The run-rate projection caption ("~\$X by end of {month}") uses
`cycleRemaining()` in `tokenjam/ui/index.html`, which computes days remaining in
the **calendar month**. But a user can configure a non-calendar billing cycle
via `[budget.<provider>] cycle_start_day` (e.g. `15`), and the existing
`tj optimize budget-projection` analyzer already honors it.

With `cycle_start_day = 15`, the "by end of June" framing is off by ~half a
month, undercutting the honesty parity the rest of the optimize surface holds.

### Acceptance
- The Overview/Cost run-rate projection projects to the end of the user's
  configured cycle (per `cycle_start_day`), falling back to the calendar month
  when no provider cycle is configured.
- Caption reflects the actual cycle end ("by end of cycle" / a dated form),
  consistent with how `budget-projection` frames cycle boundaries.

Likely needs the cycle bounds surfaced to the UI (e.g. on the `/api/v1/cost`
response or a small dedicated field) since the cycle config lives server-side.

PR #165, opened 2026-06-19
fix(#154): HTTP fallback for `tj report --reuse` when daemon holds the DB lock

### LINKED ISSUE #154: [follow-up] tj report --reuse needs an HTTP fallback when daemon holds the DB lock
## Symptom

\`tj report --reuse\` errors out when \`tj serve\` is running:

\`\`\`
Error: tj report --reuse needs direct database access. Stop the daemon
with \`tj stop\` and re-run, or query via the API.
\`\`\`

Surfaced during the v0.4.0 pre-release test pass.

## Root cause

\`cmd_report\` for \`--reuse\` queries the spans table directly to fetch each cluster's planner text (for skeleton rendering) and the per-cluster example trace IDs. When the daemon holds the DuckDB write lock, this path can't open the database.

\`tj optimize\` already has the analogous fallback pattern: when a daemon is detected, it uses \`ApiBackend\` (\`tokenjam/core/api_backend.py\`) to fetch the report via \`/api/v1/optimize\` instead of querying DuckDB directly. \`tj report --reuse\` doesn't have this — and the \`/api/v1/optimize\` payload doesn't carry enough information (it returns the analyzer finding but not the planning-call text or the example session details the renderer needs).

## What this issue is

Two options to consider:

**Option A — extend \`/api/v1/optimize\` to carry skeleton-ready data.**
Add the planner text + example IDs to each Reuse cluster in the API response. \`tj report --reuse\` then uses \`ApiBackend\` to fetch and render without direct DB access.
- ✅ Single endpoint to maintain
- ❌ Payload size grows for every \`/optimize\` call, even when no report is being generated

**Option B — add a dedicated \`/api/v1/reuse/clusters\` endpoint.**
Returns the same data the analyzer produces plus the skeleton-rendering extras (planner text, example IDs). Only called when the report is generated.
- ✅ Pay-for-what-you-use
- ❌ One more endpoint to maintain

My recommendation is **B** — the cluster-rendering payload is large enough (per-cluster planner text can be many KB) that we shouldn't pay for it on every \`/optimize\` poll from the Lens Overview.

## Workaround until this lands

\`tj stop && tj report --reuse\` works fine. The error message already documents this.

## Severity

Medium UX gap, not a correctness issue. \`tj report --trim\` has the same limitation (also DB-direct only), so this isn't a Reuse regression — it's a shared limitation of the \`tj report\` family that we should address consistently when we fix it.

anilmurty:
Closes #154 (follow-up surfaced during the v0.4.0 pre-release pass).

## Problem
`tj report --reuse` queries the `spans` table directly for each cluster's planning text + example sessions, so it errored — *"needs direct database access. Stop the daemon with `tj stop`…"* — whenever `tj serve` held the DuckDB write lock. `tj optimize` already has the analogous `ApiBackend` fallback; `--reuse` didn't, and `/api/v1/optimize` doesn't carry the planner text / example details the renderer needs.

## Fix — Option B (dedicated endpoint, pay-for-what-you-use)
Per the issue's recommendation:

- **`GET /api/v1/reuse/clusters`** (new `tokenjam/api/routes/reuse.py`) returns `report_to_dict(report)` — so the CLI reconstructs the finding through the **existing** `report_from_dict` — plus the two report-only extras the renderer needs: `planning_texts` (`{session_id: completion text | null}`) and `pricing_mode`. Dedicated rather than bolted onto `/optimize` because per-cluster planner text can be many KB and the Overview polls `/optimize` every 30s; this endpoint is hit only when a report is generated.
- **`ApiBackend.fetch_reuse_clusters`** mirrors `fetch_optimize_report`.
- **`core/export/reuse_report.py`**: `prepare_renders` / `write_reuse_report` now accept injected `planning_texts` (HTTP path) as an alternative to `conn` (local path); new `gather_planning_texts()` batches the per-cluster text fetch server-side. Local rendering is unchanged.
- **`cmd_report._render_reuse_report`** dispatches exactly like `cmd_optimize`: direct `conn` when available, else `ApiBackend` + `/api/v1/reuse/clusters`. `--trim` stays DB-direct (now documented as the remaining shared limitation).

## Tests
- `test_api.py` — `/api/v1/reuse/clusters` returns the finding (`findings.reuse.clusters`) + non-empty `planning_texts` + `pricing_mode` for a seeded cluster (exercises the route handler end-to-end).
- `test_cli.py` — `tj report --reuse` over an `ApiBackend` (no `.conn`) writes the HTML + Markdown sidecar and never emits the "needs direct database access" error. The payload is captured from the real route handler, then the sync network hop is stubbed (sync httpx can't drive ASGI in-process).

`CLAUDE.md` updated (routes list + `tj report` fallback note). Full CI set green locally (768 passed), ruff + mypy clean on changed files.

> Built in an isolated `git worktree` off `main`. No file overlap with the other follow-up PRs.


anilmurty (commented):
## Review: ✅ Approve

The cleanest of the four v0.4.1 batch PRs. CI green on 3.10/3.11/3.12 + TS. Faithful to the design directive, well-tested, honestly scoped.

## Architecture — Option B implemented exactly as directed

**Dedicated `/api/v1/reuse/clusters` endpoint** (not bolted onto `/optimize`) — the route's own docstring defends this: *"per-cluster planner text can be many KB, and the Overview polls `/optimize` every 30s — we don't make every poll pay for report-only data."* That's the verbatim economic argument from #154's design comment, internalized in the code.

**Payload shape is right.** `report_to_dict(report)` (so the CLI reconstructs via existing `report_from_dict`) + two report-only extras (`planning_texts` map + `pricing_mode` string). One HTTP request gets everything the renderer needs.

**`cmd_report._render_reuse_report` mirrors `cmd_optimize`'s direct-vs-API-shim dispatch.** No novel architecture; reused the existing pattern. Exactly per the directive.

**`gather_planning_texts(conn, finding)` is batched** across all clusters in one DB query, not a naive per-cluster fetch.

## `reuse_report.py` refactor is graceful

`prepare_renders` and `write_reuse_report` now accept **either** `conn` (local path) **or** `planning_texts` (HTTP path). When `planning_texts` is supplied, no client-side capture-completions gating — the daemon already applied that filter server-side. When `conn` is supplied, gating happens locally as before. Clean separation. Existing local callers (`tj optimize reuse --export-templates`) are unaffected.

## The CLI test pattern deserves a callout

`test_report_reuse_api_mode_writes_artifacts` solves a real testing challenge: `runner.invoke()` is sync, so it can't await an `httpx.AsyncClient` against ASGI. The agent worked around it cleverly:

1. **Capture the real endpoint payload** via `asyncio.run(_fetch())` against the ASGI app
2. **Stub `fetch_reuse_clusters`** on an `ApiBackend` instance with `monkeypatch.setattr(backend, "fetch_reuse_clusters", lambda **kw: payload)`
3. **Invoke the CLI** with that stubbed backend — runs the real `cmd_report` code path, including `report_from_dict`, against a real route-handler-produced payload

Route handler is exercised genuinely; the network hop is faked. The asserts then verify the CLI took the HTTP path (`"Reuse report written"`), the old error is gone (`"needs direct database access" not in output`), and both HTML + MD files were written.

## Honest scoping

`tj report --trim` stays on the direct-DB path, **explicitly documented** in the new CLAUDE.md text as *"the remaining shared limitation."* Could have scope-creeped into fixing both; instead the PR stayed laser-focused on #154's specific symptom and named the equivalent `--trim` gap as a separate future item. Good restraint — that's the right call.

## Tiny non-blocking observations

1. **`prepare_renders` accepts both `conn=None` and `planning_texts=None`** — callers must provide exactly one. A runtime `if conn is None and planning_texts is None: raise ValueError` would be defensive. Existing call sites always provide one, so this is hypothetical.
2. **`Query(..., alias="agent_id")` on the endpoint** — the alias matches the param name, so it's functionally redundant. Harmless.

Neither warrants a follow-up.

## With this merged, the v0.4.1 batch core is complete

| Issue | PR | Status |
|---|---|---|
| #138 | #158 | ✅ merged |
| #124 | #161 | ✅ merged |
| #139 | #163 | ✅ merged |
| #154 | **#165** | ✅ this PR |

Ship it.

anilmurty:
Thanks — "ship it" noted. Folded in the first nit since it's cheap and matches the no-silent-failure ethos:

- **#1 (source guard):** `prepare_renders` now raises `ValueError` if neither `conn` nor `planning_texts` is supplied. Without it, "neither" silently rendered every cluster skeleton-less (capture off) or crashed in `_fetch_planning_texts(None, …)` (capture on). Added a focused unit test (`test_prepare_renders_requires_conn_or_planning_texts`). All existing call sites already pass exactly one source, so behavior is unchanged for them.
- **#2 (redundant `alias="agent_id"`):** left as-is on purpose — `/api/v1/optimize` uses the identical `Query(None, alias="agent_id")`, so keeping it makes the mirrored endpoint consistent with the one I patterned it on. Happy to drop it from both in a separate sweep if you'd rather.

Pushed as one commit (`29ae212`). CI re-running; 89 reuse/CLI/API tests green locally, ruff + mypy clean.

PR #166, opened 2026-06-20
Fix Recoverable Waste tile title casing + not-ready prefix (#162)

### LINKED ISSUE #162: [bug] Lens Overview — Recoverable Waste tiles have inconsistent title casing, weight, and content-line prefixes
## Summary

The Lens Overview's *Recoverable Waste* tiles render inconsistently across the 5 analyzers, in three ways. All cosmetic, none affecting numbers — but they read as bugs because the tiles sit side-by-side.

## Observed (v0.4.0)

| Tile | Title text | Title weight | Content line |
|---|---|---|---|
| Downsize | "Downsize" | normal | "No candidates" |
| Cache | "Cache" | **bold** ❌ | "✓ 100% cache efficacy" (bold green) |
| Script | "Script" | normal | "No candidates" |
| reuse | "reuse" ❌ (lowercase) | normal | "No candidates" |
| Trim | "Trim" | normal | "— not ready" ❌ (em-dash prefix) |

### 1. Title weight should be uniform across all five tiles

Cache is the only tile whose **title** renders bold. That's almost certainly because the `at_ceiling` state CSS class applies bold to the whole tile rather than just the content row. The positive-state emphasis belongs on the **content line** (the "✓ 100% cache efficacy" line, which is already styled bold green) — not on the heading. The heading should be the same weight in all five tiles.

### 2. Titles should all be title-case

The analyzer registry uses lowercase keys (`downsize`, `cache`, `script`, `trim`, `reuse`). Four of them get capitalized in the UI (presumably via a `text-transform: capitalize` rule or a per-tile string), but `reuse` slips through and renders lowercase. Either:
- Apply the same string transformation to every tile (centralize the capitalization in one place), or
- Fix the `reuse` tile's specific title string to be `"Reuse"`.

The first option is more durable — when the 6th analyzer ships, it'll capitalize automatically.

### 3. Content-line prefixes should follow a consistent scheme

Three distinct states are rendered with three different prefix patterns today:

| State | Current prefix | Example |
|---|---|---|
| `no_findings` | no prefix | "No candidates" |
| `at_ceiling` | `✓` | "✓ 100% cache efficacy" |
| `not_ready` | `—` (em-dash) | "— not ready" |

Either drop the em-dash so Trim reads simply "Not ready", **or** decide on a consistent prefix scheme for all three states (e.g., status icon per state: `✓` for at_ceiling, no icon for no_findings, `○` or `⌛` for not_ready). My recommendation: just drop the em-dash for now — the other tiles don't have a state-icon prefix, so adding one is more scope than needed for consistency.

## Files likely touched

- `tokenjam/ui/index.html` — the Recoverable Waste band rendering logic (`classifyFinding()` + the tile template).
- Possibly the inline CSS for the `.rec-tile.ok` (at_ceiling) class — the title-weight rule belongs on the `.rec-tile-content` (or equivalent) sub-element, not the whole tile.

## Tests

Add static-grep guards to `tests/unit/test_lens_ui_regression.py` mirroring the existing pattern from the #127 fix:

- Title weight: assert that the at_ceiling tile's bold rule doesn't apply to the title element (could grep for the CSS selector specificity, or render the served HTML and check the heading is in a `<div class="rec-tile-title">` without a bold class).
- Reuse capitalization: assert that the served HTML contains a "Reuse" tile (capital R) — covers both fix paths (centralized transform OR per-tile string).
- Em-dash removal: assert `"— not ready"` is NOT in the served HTML; `"Not ready"` IS.

## Acceptance criteria

1. All 5 tile titles render in the same font weight (not bold).
2. All 5 tile titles render in title-case (`Reuse`, not `reuse`).
3. Trim's "not ready" content line drops the em-dash prefix (or — if a state-icon scheme is added — uses the same prefix scheme as the other tiles consistently).
4. The Cache `at_ceiling` positive framing is preserved (✓ + bold-green content line is the intended design from #127 — only the title-weight inconsistency should change).
5. Static-grep regression guards added per the Tests section above.

## Out of scope

- Adding per-analyzer emoji icons (the marketing site uses 🪶 / 💾 / 📜 / ✂️ / 🔁 — that's a bigger design question, file separately if wanted).
- Changing tile dimensions, ordering, or grid layout.

anilmurty:
Closes #162.

Three cosmetic inconsistencies in the Lens Overview *Recoverable Waste* tiles. All static-template/CSS, no number changes.

## Fixes

**2. Title-case (`reuse` rendered lowercase).** `reuse` was missing from `ANALYZER_META`, so the tile fell back to its raw lowercase registry key. Added `reuse: { title: 'Reuse', hint: … }` **and** centralized capitalization in a `capitalize()` helper used by both fallback sites (Overview band + Optimize-detail title). Per the issue's preferred option, this is durable: a future 6th analyzer auto-title-cases instead of leaking its lowercase key.

**3. Em-dash prefix.** Trim's `not_ready` line `"— not ready"` → `"Not ready"`, matching the prefix-free scheme of the other states (recommendation from the issue: drop the em-dash rather than add a state-icon scheme).

**1. Title weight.** In current `main` this is **already uniform** — every tile state renders the title in a bare `<div class="rec-name">` (font-weight 600), and the only `.rec-tile.ok` rules are the hover border + the green content line `.rec-amount.ok`. There is no rule that bolds the title for the at_ceiling (Cache) tile, so no CSS change was needed; I added a guard test to lock it. The intended #127 positive framing (✓ + bold-green **content** line) is preserved per AC #4.

## Tests
Added to `tests/unit/test_lens_ui_regression.py` (mirroring the #127 pattern):
- `test_reuse_tile_title_is_title_cased` — `reuse` in `ANALYZER_META`, `capitalize()` helper present, raw-lowercase fallback gone.
- `test_not_ready_tile_drops_em_dash` — `"— not ready"` absent, `"Not ready"` present.
- `test_recoverable_tile_titles_share_one_weight` — all states use the identical `.rec-name` title element; `.rec-amount.ok` preserved; no `.rec-tile.ok .rec-name` override.

## Verification
- UI-offline regression green; full suite **775 passed**.

## Note on AC #1
If you still see Cache's title bold in a running build, it may be browser cache — current `main` source has uniform title weight. Happy to dig further if it reproduces. Out-of-scope items (per-analyzer emoji, layout/ordering) untouched.


anilmurty (commented):
## Review: ✅ Approve

Genuinely thoughtful work. The agent's restraint on item #1 — *"couldn't reproduce, so I'm not making speculative CSS changes, but I'm locking the current behavior with a regression test"* — is exactly the right engineering call. Spec'd three fixes; two were real code changes, one was an honest "can't reproduce" with a guard test instead of cargo-culted CSS.

## The three fixes

**#1 — Cache title bold: investigated, couldn't reproduce, locked with a guard test.**

The agent walked current `main`: all 5 tiles already use the same `<div class="rec-name">` element, and no `.rec-tile.ok .rec-name` rule exists in source. Rather than add speculative CSS to "fix" a defect they couldn't reproduce, they locked the current behavior:

```python
assert html.count('<div class="rec-name">${meta.title}</div>') >= 3
assert ".rec-tile.ok .rec-name" not in html  # no state-specific title bolding
assert ".rec-amount.ok" in html               # #127 positive emphasis preserved
```

The screenshot that triggered the issue most likely came from a stale browser cache — the daemon had been running across multiple rebuilds. If a hard-refresh post-merge confirms the title is now uniform, the agent's investigation was right and the guard test prevents regression. If the bold reappears post-refresh, there's a more interesting bug worth chasing — but the test will catch it.

**#2 — Centralized capitalization for the `reuse` lowercase bug:**

```js
function capitalize(name) {
  return name ? name.charAt(0).toUpperCase() + name.slice(1) : name;
}
```

Plus `reuse: { title: 'Reuse', hint: '...' }` added to `ANALYZER_META`. Both fallback call sites (`tiles.map` + `OptimizeFinding`) use `capitalize(t.name)`. So:
- Reuse renders correctly today via its explicit `ANALYZER_META` entry
- A future 6th analyzer auto-capitalizes via the fallback path instead of rendering raw lowercase

Immediate fix + forward-compat in one change. Smart consolidation.

**#3 — `— not ready` → `Not ready`:** one-line template change. Now prefix-free, consistent with the other tile states.

## Tests — strong-form guards

Each test asserts both presence (new code/copy is there) AND absence (old buggy substring is gone):

- **`test_reuse_tile_title_is_title_cased`** — asserts the new `ANALYZER_META` entry + `function capitalize` exist, AND the old raw-lowercase fallback (`{ title: t.name, hint: '' }`) is gone
- **`test_not_ready_tile_drops_em_dash`** — `— not ready` gone, `>Not ready<` present
- **`test_recoverable_tile_titles_share_one_weight`** — counts uniform title elements (≥3), AND asserts no state-specific title-bold rule, AND preserves `.rec-amount.ok` (the intended #127 positive emphasis)

That's the strong-form guard pattern: a future regression has to either add the bad pattern back OR remove the good pattern; both are caught.

## Quality

- Full suite 775 passed
- UI-offline regression green
- Built in an isolated worktree (CLAUDE.md guidance followed)
- 41 net additions across just 2 files — surgical scope

## One verification step post-merge

Do a hard refresh at the Lens UI after this lands. If the Cache title is now uniform with the others, the agent's investigation stands and we're done. If the bold reappears, that's a runtime CSS specificity or browser-rendering quirk worth chasing separately — but the guard test will catch any source-level reintroduction either way.

Ship it.

PR #184, opened 2026-06-21
Apply declared plan tiers across onboard, serve, and UI

HoomanDgtl:
#183

Issue #179, opened 2026-06-20
[bug] Lens Overview 24h chart doesn't reflect today's activity until Claude Code is restarted (possible)


anilmurty:
## Symptom

After running \`tj onboard --claude-code --plan max_5x\` (which writes new OTEL endpoint settings to \`~/.claude/settings.json\`), the Lens Overview's 24h cost chart shows mostly-flat activity even though the user has been actively using Claude Code that day.

The \`tj onboard\` output prints \"Restart Claude Code for settings to take effect.\" — but if the user doesn't see this or doesn't restart, today's work silently doesn't flow into TokenJam.

## What's actually happening (hypothesis)

Claude Code reads the OTLP exporter env vars at startup, not on each request. When TokenJam changes the endpoint (e.g., new ingest secret after onboard), running Claude Code instances continue using the old endpoint, which 401s silently or hits a different daemon entirely.

So:
- New Claude Code sessions started AFTER restart → flow correctly
- Existing Claude Code session running across the reconfigure → spans go to old endpoint, are lost
- Backfill catches sessions that were written to disk, but they're \"stale\" — at most current to when Claude Code last flushed

## Acceptance criteria — options to pick from

Several possible fixes, in order of effort:

1. **Make Lens surface the staleness.** Add a small status indicator on the Overview saying \"Last live span received: 3h ago\" or similar. If the gap is large, surface a hint: \"Looks like Claude Code may not be connected — restart Claude Code (or check `tj doctor`).\"
2. **Improve the onboard message.** Instead of a single \"Restart Claude Code for settings to take effect\" line at the end of \`tj onboard\` output, make it a prominent banner. Optionally pause for user confirmation before completing.
3. **Auto-restart Claude Code if possible.** Probably too invasive; flagging as an option.
4. **`tj doctor` check.** Add a check that compares the latest live-span timestamp to wall-clock; if the gap is large, prompt the user to restart Claude Code.

## Severity

Medium. Users who never restart will think TokenJam isn't working. \"I onboarded but nothing is being tracked\" is the most common silent-failure mode.

Discovered during personal-usage validation. The reconfigure path is the most affected.

Issue #175, opened 2026-06-20
[bug] tj cost shows raw dollars regardless of plan tier — bypasses core/framing


anilmurty:
## Symptom

A Max-plan user running \`tj cost --since 30d\` sees raw dollar figures (\"\$2599.13\") with no plan-tier framing, no qualifier, no suppression — implying they spent \$2,599 when they actually paid the Max plan flat fee (\$100/mo).

Other surfaces apply plan-tier framing correctly:
- \`tj tokenmaxx\` → renders \"26.0× your Max 5x plan cost (\$100/mo flat)\" — honest
- \`tj optimize\` → renders \"All sessions have unknown plan tier; dollar figures suppressed\" (when sessions have unknown tier; correct behavior)
- \`/api/v1/cost\` JSON response → includes the \`framing\` block, but the CLI rendering ignores it

\`tj cost\` is the lone holdout.

## Root cause

PR #110 (framing module) covered \`cmd_optimize\` and \`cmd_tokenmaxx\` plus the API responses. PR #120 (later follow-up) covered \`tj cost --compare\` (\`_render_diff\`). The bare \`tj cost\` rendering — i.e. the table at \`cmd_cost.py\` between the \`compute_cost\` call and \`console.print(table)\` — was left out of scope at the time. It still calls \`format_cost(...)\` for the COST column with no plan-tier check.

## Fix

Apply the same framing pattern \`cmd_optimize\` uses:

1. After computing the cost rows, call \`compute_framing(config, window_summary, ...)\` to get a \`Framing\` instance (the same way \`_compare_framing\` does in #120).
2. Wrap each COST cell render in \`render_dollar(value, framing)\` from \`core/framing.py\`.
3. Surface the \`qualifier_text\` (e.g. \"Plan tier unknown — figures may overstate actual cost\") above the table when present.
4. For subscription users, the COST column should either suppress dollar figures or show token-share framing — match the existing pattern in \`cmd_optimize._render_downgrade\`.

## Acceptance criteria

1. \`tj cost --since 30d\` on a Max-plan install shows either:
   (a) suppressed dollar figures with a clear note about subscription billing, OR
   (b) the qualifier \"X% of cycle tokens / X× implied API value\" framing
   — same shape as \`tj tokenmaxx\`.
2. \`tj cost --since 30d\` on an API-plan install shows raw dollar figures (no behavior change for that path).
3. Same test added to \`tests/unit/test_cmd_cost.py\` covering both paths.

[...]

PR #186, opened 2026-06-21
Fix #177, #178: consistent Lens chart framing + local-timezone axis labels

### LINKED ISSUE #177: [bug] Lens Overview chart framing is window-dependent (24h shows tokens, 7d/30d show dollars)
## Symptom

On the Lens Overview screen, the cost-hero chart changes its display unit based on the window selected:

- **Last 24h** → \"Spend (tokens) · last 24h: 1.7M tokens\" — token framing, no plan-tier banner
- **Last 7d**  → \"Spend · last 7d: \$1596.5273\" + \"Plan tier unknown — figures may overstate\" qualifier banner
- **Last 30d** → \"Spend · last 30d: \$2599.1322\" + same qualifier banner

The same underlying data, viewed at a different window, presents in entirely different units AND with different qualifier banners. The plan-tier-aware framing logic appears window-dependent when it should be a property of the user's plan, not the time range.

## Root cause hypothesis

The framing computation likely lives in the API response for \`/api/v1/cost?since=X\` and varies based on something like \"if total cost very low, fall back to tokens\" or similar heuristic. The 24h window has lower absolute cost than 7d/30d, possibly triggering a different code path.

Or: the framing logic correctly responds to the plan-tier and pricing-mode from the session records, but only the 24h window's data falls into a state that triggers token framing while the longer windows trigger dollar framing.

## Acceptance criteria

1. Open Lens Overview, switch between 24h / 7d / 30d / 90d — the chart's unit (tokens vs dollars), the plan-tier qualifier banner, and the framing semantics should be **consistent across all windows for the same user**.
2. If the user is on a subscription plan: all windows show tokens / token-share framing.
3. If the user is on api: all windows show dollars.
4. If plan tier is unknown: all windows show dollars with the \"may overstate\" qualifier (or tokens if that's the chosen behavior — pick one and apply consistently).
5. The Cost detail screen should match the Overview's framing for the same window.

## Severity

Medium. The data is correct under each rendering; the user is just confused why the same data presents differently. But \"the same screen shows two different mental models depending on time window\" is the kind of bug that erodes trust in the dashboard.

Discovered during personal-usage validation. Sibling to #175 (CLI \`tj cost\` framing).

### LINKED ISSUE #178: [bug] Lens charts display times in UTC instead of the user's local timezone
## Symptom

The Lens Overview / Cost spend chart's x-axis displays times in **UTC**, not the user's local timezone. A US-Pacific user looking at a 24h chart sees \"7pm\" on the axis at what was actually 12 noon their local time (7 hours offset, PDT).

## Acceptance criteria

1. Chart x-axis labels for hourly resolutions (24h window) display in the **browser's local timezone**, not UTC.
2. Date labels (7d/30d/90d windows, where labels are dates like \"Jun 13\") match the local calendar day, not UTC's.
3. The trace-detail timestamps continue to display sensibly (mid-debate whether they should also localize or remain UTC for traceability — leaving outside this issue's scope).

## Implementation hint

The chart most likely receives UTC timestamps from the API (which is correct — server-side data should be UTC). The UI layer should localize when rendering tick labels. JavaScript's \`Date.toLocaleTimeString()\` / \`toLocaleDateString()\` handle this automatically when given a UTC timestamp.

## Severity

Low-medium. Cosmetic and confusing, but not a data-correctness issue. Users in UTC won't notice. Pacific / Eastern / European users will.

Discovered during personal-usage validation.

anilmurty:
The Lens Overview / Cost spend chart presented the same user's data in different units depending on the selected time window, and rendered its x-axis ticks in UTC rather than the viewer's local timezone. Both are confusing-but-not-data-wrong bugs that erode trust in the dashboard.

## Summary
- **#177**: the chart's tokens-vs-dollars unit and the plan-tier qualifier banner are now a property of the user's **plan**, not the time window — consistent across 24h / 7d / 30d / 90d, and the Cost detail screen matches the Overview.
- **#178**: the chart x-axis tick labels render in the **browser's local timezone**; server data stays UTC.
- Backend-only fix for #177 (the UI already consumes the server `framing` block — no plan-tier framing is re-derived in JS). Honesty discipline preserved: subscription / local / unknown dollar-suppression is untouched.

## #177 — chart framing was window-dependent
**Symptom.** For one user, 24h showed token framing (no banner) while 7d/30d showed dollar framing with the "Plan tier unknown" qualifier — same data, different mental model per window.

**Re-verified against current `main`** (post-#184). It still reproduces for users **without** a declared plan, or whose session data diverges from the window. Seeding 40 long-running sessions (started 3–14 days ago, but with spans in the last hour) and hitting `/api/v1/cost`:

| window | before (main) | after |
|---|---|---|
| 24h | `api` / `show_dollars` (no banner) | `unknown` / `suppress_dollars_unknown` |
| 7d  | `unknown` / `suppress_dollars_unknown` | `unknown` / `suppress_dollars_unknown` |
| 30d | `unknown` / `suppress_dollars_unknown` | `unknown` / `suppress_dollars_unknown` |

(Undeclared, subscription-stamped sessions flipped 24h=`api`/dollars vs 30d=`subscription`/tokens — the exact reported class of bug.)

**Root cause.** `/api/v1/cost` derived its framing from a **window-scoped** session mix — `plan_tier_mix(conn, since, until, agent_id)`, keyed on `session.started_at`. That window diverges from the span-based cost totals: a 24h window with recent spans from sessions that *started* >24h ago has an **empty** session mix, so `dominant_plan({})` collapses to the `"api"` default (dollars), while a 7d/30d window sees the real `unknown` / subscription sessions and frames accordingly.

**Fix.** The pricing mode + qualifier banner are a property of the plan, so the cost route now determines them from a new **window-independent** `plan_determination_mix(conn, agent_id)` (the user's full session history). Only the window totals and the chart `series` stay window-scoped. The shared subscription qualifier is reworded scope-neutrally ("of sessions", not "this window") because the same `compute_framing` still serves the window-scoped CLI / `--compare` callers.

Acceptance criteria: subscription → tokens on every window; api → dollars; unknown → one consistent choice across all windows; Cost detail matches Overview. ✔

## #178 — x-axis showed UTC instead of local time
**Symptom.** A US-Pacific user on a 24h chart saw "7 PM" where it was actually noon local (PDT, −7h).

**Root cause.** `SpendChart`'s `fmtAxisTime` forced `{ timeZone: 'UTC' }` on both the hourly and date tick formatters.

**Fix.** Drop the explicit `timeZone` from the two tick formatters only. `new Date(epoch*1000)` already holds the correct instant; `toLocaleTimeString()` / `toLocaleDateString()` then format it in the viewer's local zone. Server buckets stay UTC. Trace-detail timestamps are untouched (out of scope per the issue).

## Tests / Verification
- `tests/integration/test_cost_framing_window.py` (new) — seeds the long-running-session divergence against a real DuckDB backend and asserts `/api/v1/cost` framing (mode + display_rule + qualifier) is identical across 24h/7d/30d for subscription (declared & data-driven) and unknown users, while the dollar/token **totals** still grow with the window.
- `tests/unit/test_framing.py` — `plan_determination_mix` passes `since=until=None` (no `started_at` clause), so the framing mix is window-independent.
- `tests/unit/test_lens_ui_regression.py` — static-grep guard asserting `fmtAxisTime` uses `toLocale*String` and no longer forces `timeZone: 'UTC'`.
- Full suite green: `pytest tests/unit/ tests/integration/ tests/synthetic/ tests/agents/` → **802 passed**. `ruff` + `mypy` clean on touched files. `node --check` clean on the extracted module.
- **Visual** (headless Chrome, machine TZ = PDT −0700, seeded subscription user):
  - Overview 24h and 30d both render "Spend (tokens)" + "% of cycle tokens" + "Max 5x plan" — consistent unit across windows.
  - Cost detail 24h chart matches the Overview (tokens), satisfying AC #5.
  - 24h axis reads local hours (12 PM → 12 PM) with the recent-activity bumps at ~6–10 AM local (their UTC afternoon equivalents would have read "1 PM"/"6 PM" pre-fix).

## What's NOT in this PR
- **Web Cost-table COST column framing.** The detail table's COST cells still render raw `$` via `fmtCost` (line 1537) for subscription users — pre-existing and independent of the chart (the chart-vs-table mismatch already existed at 7d/30d before this fix). It's the web analog of the CLI `tj cost` table fix (#175), and belongs in its own change rather than expanding this chart-framing PR.
- **Trace-detail timestamp localization** — explicitly left UTC per #178's scope note.
- **#179 staleness indicator** — a separate agent's concern.
- I did not extend the window-independent mix to `/optimize` or `/cost/compare`; #177 is scoped to the cost chart + Cost detail (both `/api/v1/cost`). Those routes remain window-scoped and can be aligned separately if desired.

Closes #177
Closes #178

### DIFF
--- a/tests/integration/test_cost_framing_window.py
+++ b/tests/integration/test_cost_framing_window.py
@@ -1,0 +1,130 @@
+"""#177 — /api/v1/cost plan-tier framing must be window-INDEPENDENT.
+
+The cost chart's tokens-vs-dollars unit and the qualifier banner are a property
+of the user's plan, not the selected time window. Before the fix, the framing
+mix was scoped to the window via ``session.started_at``; a 24h window with no
+in-window session *starts* (but recent spans from long-running sessions)
+collapsed to the empty-mix "api" default and rendered dollars, while a 7d/30d
+window rendered the subscription / unknown framing for the very same user.
+
+These tests seed long-running sessions whose ``started_at`` is outside the 24h
+window while their spans are recent, then assert the framing block is identical
+across 24h / 7d / 30d. Uses a real DuckDBBackend because the route's mix query
+runs against ``db.conn`` (InMemoryBackend has none and short-circuits to {}).
+"""
+from __future__ import annotations
+
+import dataclasses
+from datetime import timedelta
+
+import httpx
+import pytest
+
+from tokenjam.api.app import create_app
+from tokenjam.core.config import ProviderBudget, StorageConfig, TjConfig
+from tokenjam.core.db import DuckDBBackend
+from tokenjam.core.ingest import IngestPipeline
+from tokenjam.utils.time_parse import utcnow
+from tests.factories import make_llm_span, make_session
+
+
+@pytest.fixture(autouse=True)
+def _isolate_home(monkeypatch, tmp_path):
+    """Point Path.home() at an empty dir so config_declared_plan's global
+    fallback never reads the dev machine's ~/.config/tj/config.toml."""
+    monkeypatch.setattr("pathlib.Path.home", lambda: tmp_path)
+
+
+def _seed_long_running(db, plan_tier: str) -> None:
+    """40 sessions that STARTED 3–14 days ago but each have a span in the last
+    hour — the divergence that made 24h framing differ from 30d (#177)."""
+    now = utcnow()
+    for i in range(40):
+        days = 3 + (i % 12)
+        s = make_session(session_id=f"s-{i}", plan_tier=plan_tier)
+        s = dataclasses.replace(s, started_at=now - timedelta(days=days))
+        db.upsert_session(s)
+        db.insert_span(make_llm_span(
+            session_id=f"s-{i}", billing_account="anthropic",
+            start_time=now - timedelta(days=days),
+            input_tokens=5000, output_tokens=800, cost_usd=40.0,
+        ))
+        db.insert_span(make_llm_span(
+            session_id=f"s-{i}", billing_account="anthropic",
+            start_time=now - timedelta(hours=1, minutes=i % 50),
+            input_tokens=1000, output_tokens=200, cost_usd=5.0,
+        ))
+
+
+async def _framings_across_windows(db, config):
+    pipeline = IngestPipeline(db=db, config=config)
+    app = create_app(config=config, db=db, ingest_pipeline=pipeline)
+    transport = httpx.ASGITransport(app=app)
+    out = {}
+    async with httpx.AsyncClient(transport=transport, base_url="http://t") as c:
+        for win in ("24h", "7d", "30d"):
+            out[win] = (await c.get(f"/api/v1/cost?since={win}")).json()["framing"]
+    return out
+
+
+@pytest.mark.asyncio
+async def test_framing_consistent_across_windows_subscription(tmp_path):
+    """Declared subscription user → tokens framing on every window (AC #2)."""
+    db = DuckDBBackend(StorageConfig(path=str(tmp_path / "t.duckdb")))
+    config = TjConfig(version="1", budgets={"anthropic": ProviderBudget(plan="max_5x")})
+    _seed_long_running(db, plan_tier="unknown")  # data unstamped; plan declared
+    framings = await _framings_across_windows(db, config)
+    modes = {w: f["pricing_mode"] for w, f in framings.items()}
+    rules = {w: f["display_rule"] for w, f in framings.items()}
+    assert set(modes.values()) == {"subscription"}, modes
+    assert len(set(rules.values())) == 1, rules
+    db.close()
+
+
+@pytest.mark.asyncio
+async def test_framing_consistent_across_windows_subscription_from_data(tmp_path):
+    """Undeclared user whose sessions are subscription-stamped → tokens on
+    every window (no 24h dollars / 30d tokens split)."""
+    db = DuckDBBackend(StorageConfig(path=str(tmp_path / "t.duckdb")))
+    config = TjConfig(version="1")  # no declared plan
+    _seed_long_running(db, plan_tier="max_5x")
+    framings = await _framings_across_windows(db, config)
+    modes = {w: f["pricing_mode"] for w, f in framings.items()}
+    assert set(modes.values()) == {"subscription"}, modes
+    db.close()
+
+
+@pytest.mark.asyncio
+async def test_framing_consistent_across_windows_unknown(tmp_path):
+    """Undeclared user with all-unknown sessions → the SAME unknown framing on
+    every window (AC #4: one consistent choice, not 24h api / 30d unknown)."""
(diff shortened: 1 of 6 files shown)


anilmurty (commented):
## Verdict: ✅ APPROVE (pending Python CI green) — ready to merge once checks finish

Read the full diff; verified the framing change independently (honesty-sensitive). Scope is exactly the 6 expected files — `framing.py`, `routes/cost.py`, `ui/index.html` + 3 test files. test-ts green; Python jobs still running at review time.

### #177 — window-dependent framing → fixed, and honesty *improved*
- New `plan_determination_mix(conn, agent_id)` = `plan_tier_mix(conn, None, None, agent_id)` — window-INDEPENDENT. `/cost` (`cost.py:124`) now frames off it; window totals + `series` stay window-scoped. Correct separation: pricing mode is a property of the plan, not the time range.
- **Honesty audit — PASS, and strengthened.** The old 24h path (empty in-window mix → "api" default → *dollars shown to subscription users*) is exactly eliminated. Full-history mix means subscription users now get `SUPPRESS_SUBSCRIPTION` across all windows. No suppression weakened.
- Nice: qualifier reworded "of this window" → "of sessions" so it reads truthfully for *both* the history-wide `/cost` caller and the still-window-scoped CLI/compare callers. That's the right level of care.
- `compute_framing` config-fallback (`if not mix and declared`) still intact.

### #178 — UTC axis → local, hourly correct
- `fmtAxisTime` (`index.html:922`) drops `{ timeZone: 'UTC' }` from both formatters; server data stays UTC. Hourly 24h axis (the reported symptom) is now correct. Regression test asserts the buggy pattern is gone + helpers present.
- **Caveat (non-blocking, tracked):** daily buckets are UTC-day-aligned, so localizing their labels can shift a daily date by one day for non-UTC users — criterion 2 ("match local calendar day") is only partially met for multi-day windows. The primary symptom is fixed; the daily tail needs a design call (local-tz bucketing vs UTC date labels). Filed as #188 so it isn't lost when this closes #178. I do **not** think it should block — but flagging for your call.

### Tests
- `test_cost_framing_window.py` (real DuckDB, asserts framing identical across windows while totals still grow) + `test_framing.py` (mix window-independence) + `test_lens_ui_regression.py` (no `timeZone: 'UTC'`). Genuinely verify the fixes. Agent reported 802 local pass, ruff+mypy clean; CI test-ts green, Python pending.

### Follow-up filed from the agent's deferred item
- The PR's "What's NOT in this PR" correctly flagged that the web **Cost table** COST column still shows raw `$` for subscription users. I verified it's broader — the table footer, Traces list, and trace-detail also leak raw dollars (the surfaces that don't consume a `framing` block). Filed as **#187** (medium, honesty-discipline). Correct call to keep it out of this PR — it's the web analog of CLI #175, separate scope.

**Ready to merge once the three Python checks go green.** Two follow-ups (#187 medium, #188 low) capture what's deliberately left for later.

Issue #187, opened 2026-06-21
[bug] Lens web UI shows raw dollar figures to subscription users on table & trace surfaces (bypasses framing)


anilmurty:
## Symptom

Subscription/local-plan users (who should never see raw dollar spend — TokenJam's honesty discipline) still see raw `$` figures on several Lens web surfaces, even though the Overview and Cost-screen *hero* numbers correctly suppress to token framing:

- **Cost detail table** — every row's COST column (`tokenjam/ui/index.html:1534`) and the footer total row (`:1542`) render raw `fmtCost(r.cost_usd)` / `fmtCost(total)`. On the *same screen*, the hero total (`:1462`) and projection (`:1458`) correctly use `useTokens` / `fmtFramedDollar`. So a Claude Max user sees "12.3M tokens" as the headline but `$1.2345` in every table row — two different mental models on one screen.
- **Traces list** — COST column (`:1158`) renders raw `fmtCost(t.cost_usd)`.
- **Trace detail** — span cost strings at `:1243`, `:1256`, `:1286` render raw `fmtCost`.

The column-header tooltip hedges ("If you're on a subscription plan… your actual spend may differ"), but a hedge tooltip is not the suppression the rest of the app applies — Critical Rule 14 / `core/framing.py` is about *not showing the dollar figure*, not annotating it.

## Root cause

The Cost component fetches and threads the server `framing` block (`useTokens = pricing_mode in {subscription, local}`) into the hero/projection numbers, but the **table cells** were never converted — they still call `fmtCost` directly. The **Traces / trace-detail** screens fetch no `framing` block at all, so they have nothing to suppress with.

This is the web analog of the CLI `tj cost` framing bug (#175), which was fixed for the CLI in #181. The web side was partially addressed (hero numbers, #177/#186) but the tables/trace views were left raw.

## Acceptance criteria

1. For subscription/local users, the Cost detail table COST column + footer total suppress/reframe consistently with the screen's hero total (tokens or token-share — match the existing `useTokens` treatment), not raw dollars.
2. Traces list + trace-detail cost figures either consume a `framing` block and suppress for subscription/local users, or are reframed to tokens. (Decide: thread framing into those endpoints/screens vs. show tokens.)
3. API users and unknown-plan behavior are unchanged.
4. No honesty-discipline string is strengthened; the existing caveat language is preserved.

## Scope / out of scope

[...]

PR #189, opened 2026-06-21
Fix #188: keep daily chart date labels UTC-aligned

### LINKED ISSUE #188: [bug] Lens daily chart buckets are UTC-aligned but labeled in local timezone — date labels can be off by one day for non-UTC users
## Symptom

After #186 (#178 fix) localized the spend-chart x-axis labels, the **hourly** axis (24h window) is now correct — a US-Pacific user sees their local hour. But for **daily** buckets (7d/30d/90d windows), the bucket keys are still UTC-day-aligned epoch seconds while the labels now render in the browser's local zone. Result: a daily bucket holding UTC-day *Jun 15* data is labeled "Jun 14" for a Pacific (UTC-7/8) user — the label no longer matches the bucket's UTC span, and doesn't cleanly match a local calendar day either (the bucket straddles two local days).

This means #178 acceptance criterion 2 ("date labels match the local calendar day, not UTC's") is only partially met by #186: the labels shifted, but UTC-aligned buckets can't map 1:1 to local calendar days.

## Root cause

`/api/v1/cost` buckets `series` by **UTC** day (`series_bucket`, epoch-second keys at UTC midnight). `fmtAxisTime` now formats those instants with `toLocaleDateString()` (no `timeZone: 'UTC'`), so the label for a UTC-midnight key renders as the *previous* local day for west-of-UTC users. Truly matching local calendar days requires bucketing by local day, which needs the client's tz offset server-side (or client-side re-bucketing) — a bigger change than #186's client-only label fix (which followed #178's implementation hint).

## Acceptance criteria — pick one direction

1. **Local-tz bucketing:** the chart aggregates daily buckets by the viewer's local calendar day (pass tz offset to `/cost`, or re-bucket client-side), so each "Jun 15" label covers local Jun 15 00:00–24:00. (Most correct, most work.)
2. **Or revert daily labels to UTC** (keep `timeZone: 'UTC'` for the date formatter only) so the label is internally coherent with the UTC bucket span, and document that daily buckets are UTC-aligned. (Cheap, keeps hourly localized.)

Hourly localization (the primary #178 symptom) is correct and should stay.

## Scope / out of scope

- **In scope:** daily-bucket date labels on the Overview + Cost spend chart.
- **Out of scope:** hourly axis (fixed in #186); trace-detail timestamps (#178 left these UTC by design).

## Severity

Low. Cosmetic, affects only non-UTC users on multi-day windows, and the data/totals are correct. Filed so the daily-bucket tail isn't lost when #178 closes via #186.

Discovered during review of PR #186.

anilmurty:
The #178 fix (#186) localized all spend-chart x-axis labels. That's right for hourly ticks but wrong for daily buckets, which are UTC-day-aligned server-side — so a local-timezone label on a UTC-midnight key shows the previous local day for west-of-UTC users. This splits the formatter so hourly localizes and daily stays UTC.

## Summary
- `fmtAxisTime` in `ui/index.html`: hourly ticks keep localizing (#178); daily date labels regain `timeZone: 'UTC'` to stay coherent with the UTC-aligned buckets.
- Updated the UI regression test to assert the split (hourly localizes, daily forces UTC) rather than the previous "no UTC anywhere" check.

## #188 — daily labels off by a day for non-UTC users
- **Symptom:** a US-Pacific user on a 7d/30d/90d window saw daily buckets labeled one calendar day earlier than the bucket's UTC span.
- **Root cause:** `/api/v1/cost` buckets `series` by UTC day; #186 formatted those UTC-midnight keys with `toLocaleDateString()` (no `timeZone`), shifting the date for west-of-UTC viewers.
- **Fix:** daily formatter forces `timeZone: 'UTC'` again; hourly is untouched (stays local). Chose this over local-day server-side bucketing (option 1 in the issue) — that's a larger change not worth it for a low-severity cosmetic label on coarse daily granularity.

## Tests / Verification
- `tests/unit/test_lens_ui_regression.py::test_axis_time_ticks_timezone_split` — asserts the hourly line localizes and the daily line forces UTC.
- `tests/unit/test_ui_offline.py` still green; `node --check` clean on the extracted app module.

## What's NOT in this PR
- Local-day server-side bucketing (the "fully correct" option) — deferred; daily granularity in UTC is coherent and far cheaper.
- The honesty-discipline raw-dollar leak on the Cost table / Traces (#187) — separate concern, separate PR.

Closes #188

### DIFF
--- a/tests/unit/test_lens_ui_regression.py
+++ b/tests/unit/test_lens_ui_regression.py
@@ -60,20 +60,24 @@ def test_overview_chart_is_not_a_click_target(html):
     assert "View Cost details" in html
 
 
-# --- #178: chart x-axis ticks render in the browser's local timezone ------- #
-def test_axis_time_ticks_render_in_local_timezone(html):
-    # The hourly + date tick formatter must NOT pin to UTC — it formats the
-    # UTC epoch-second buckets in the viewer's local zone (a US-Pacific user
-    # sees their noon, not UTC's 7pm). Server data stays UTC; only labels shift.
+# --- #178 / #188: chart x-axis tick timezone handling --------------------- #
+def test_axis_time_ticks_timezone_split(html):
+    # #178: HOURLY ticks localize — they format the UTC epoch-second buckets in
+    # the viewer's local zone (a US-Pacific user sees their noon, not UTC's 7pm).
+    # #188: DAILY date labels stay UTC, because the buckets are UTC-day-aligned;
+    # localizing a UTC-midnight key would print the previous local day for
+    # west-of-UTC users and no longer match the bucket span.
     import re
 
     m = re.search(r"function fmtAxisTime\(epoch, bucket\) \{.*?\n\}", html, re.DOTALL)
     assert m, "fmtAxisTime helper not found"
     body = m.group(0)
-    assert "toLocaleTimeString" in body
-    assert "toLocaleDateString" in body
-    # The bug: the formatters previously forced { timeZone: 'UTC' }.
-    assert "timeZone: 'UTC'" not in body, "axis ticks must localize, not force UTC"
+    hour_line = next(line for line in body.splitlines() if "toLocaleTimeString" in line)
+    date_line = next(line for line in body.splitlines() if "toLocaleDateString" in line)
+    # Hourly localizes (must NOT force UTC).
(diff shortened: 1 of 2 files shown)


anilmurty (commented):
Approved

Issue #195, opened 2026-06-22
[bug] LiteLLM integration captures no prompt/completion content or cache tokens


anilmurty:
## Symptom

The LiteLLM integration records only `INPUT_TOKENS` / `OUTPUT_TOKENS`. It captures:
- **No prompt/completion content**, even when the `[capture]` content toggles are on → starves the analyzers that need text: `cache-recommend`, Reuse's per-cluster text rendering, and `trim`.
- **No cache-read / cache-write token counts** → the `cache` analyzer always reports 0% efficacy for LiteLLM-sourced spans.

## Root cause (confirmed in code)

`_record_usage()` in `tokenjam/sdk/integrations/litellm.py` (and the sync/async stream wrappers) extract only `usage.prompt_tokens` and `usage.completion_tokens`. There is no extraction of message/content payloads and no extraction of the cache token fields LiteLLM exposes on the usage object.

## Acceptance criteria

1. With content capture enabled, LiteLLM spans carry prompt + completion content, routed through the existing `strip_captured_content()` gate (so the four `[capture]` toggles are honored — no behavior change when capture is off).
2. Cache-read and cache-write token counts are extracted from the LiteLLM `ModelResponse` usage object when present, so `cache` efficacy reflects reality (the `NormalizedSpan` already has `cache_tokens` / `cache_write_tokens` fields).
3. Tests (via `tests/factories.py`) covering both content capture (on/off) and cache-token capture on the LiteLLM path.

## Scope / out of scope

- **In scope:** span enrichment in the LiteLLM integration (content + cache tokens), honoring existing capture toggles and content stripping.
- **Out of scope:** provider attribution (separate issue).

## Severity

Medium-High. Capability gap that silently disables three analyzers and makes the `cache` analyzer always read 0% for LiteLLM users.

Discovered during a real-agent instrumentation pilot (Aider via the LiteLLM integration).

PR #198, opened 2026-06-22
Fix #194: resolve LiteLLM provider from model name (no more "litellm" provider)

### LINKED ISSUE #194: [bug] LiteLLM integration records provider as "litellm" → wrong pricing + NULL billing_account + suppressed dollars
## Symptom

Spans created by the LiteLLM integration can be recorded with `provider = "litellm"`, which is not a real provider. Consequences cascade:
- No pricing entry for "litellm" → cost falls back to default rates. Measured **~42% cost undercount** on a Claude Haiku 4.5 workload (token counts were correct; only the per-token rate was wrong).
- `billing_account` is provider-derived, so it becomes NULL → `SessionRecord.plan_tier` stays `unknown` even when the user declared `plan = "api"` → the CLI suppresses all dollar figures.

## Root cause (confirmed in code)

`_parse_provider()` in `tokenjam/sdk/integrations/litellm.py` (around L33–44) returns the literal `"litellm"` when both fallbacks miss:

```python
def _parse_provider(model: str, response: object) -> str:
    hidden = getattr(response, "_hidden_params", None)
    if isinstance(hidden, dict):
        provider = hidden.get("custom_llm_provider")
        if provider:
            return str(provider)
    if "/" in model:
        return model.split("/", 1)[0]
    return "litellm"          # <-- bogus provider
```

LiteLLM ≥ 1.75.0 returns `custom_llm_provider = None` in `_hidden_params`, and many callers (e.g. Aider) pass a **bare** model name (`claude-haiku-4-5`, no `anthropic/` prefix). Both fallbacks miss → `"litellm"` is recorded.

## Acceptance criteria

1. When `custom_llm_provider` is absent **and** the model is bare, resolve the provider from the model name (e.g. `claude-*` → `anthropic`, `gpt-*` → `openai`, `gemini-*` → `google`) so pricing and `billing_account` resolve correctly.
2. `"litellm"` is never written as a final provider value. If truly unresolvable, fall back to `"unknown"` (which must not be priced as if it were a real provider).
3. `billing_account` derives correctly for resolved providers; `plan_tier` propagates from declared config.
4. Tests (via `tests/factories.py`): a LiteLLM span with `custom_llm_provider=None` + bare `claude-haiku-4-5` resolves to provider `anthropic`, correct Haiku pricing, and `billing_account = anthropic`.

## Scope / out of scope

- **In scope:** provider resolution + downstream pricing/`billing_account`/`plan_tier`.
- **Out of scope:** the separate content/cache-token capture gap on the same integration (filed separately).

## Severity

High. Cost-accuracy correctness bug affecting every LiteLLM-integration user whose provider can't be resolved from hidden params or a prefixed model name, and it knocks out the plan-tier framing the honesty discipline depends on.

Discovered during a real-agent instrumentation pilot (Aider via the LiteLLM integration).

anilmurty:
The LiteLLM integration could record `provider = "litellm"` — not a real provider — which missed pricing (a measured **~42% cost undercount** on a Claude Haiku 4.5 workload), left `billing_account` NULL, and so suppressed all plan-tier dollar framing. Discovered during the Aider instrumentation pilot.

## Summary
- Add a canonical `provider_for_model()` resolver to `core/pricing.py` and use it in `_parse_provider()` so a **bare** model name (no `provider/` prefix) resolves to the real provider when LiteLLM doesn't tell us (`custom_llm_provider = None`).
- **`"litellm"` is never written as a final provider value again** — an unattributable model yields `"unknown"`, which has no pricing table (not billed as a real provider) and maps to a NULL `billing_account`.
- Fix the identical second copy of the bug in `sdk/client.py`.

## Symptom → root cause → fix
- **Symptom:** cost recorded ~42% low; `billing_account` NULL → `plan_tier` stayed `unknown` even with `plan = "api"` declared → CLI suppressed dollars.
- **Root cause:** `_parse_provider()` in `tokenjam/sdk/integrations/litellm.py` returned the literal `"litellm"` when `custom_llm_provider` was absent **and** the model was bare. LiteLLM ≥ 1.75 returns `custom_llm_provider = None`, and callers like Aider pass `claude-haiku-4-5` (no `anthropic/` prefix), so both fallbacks missed.
- **Fix:** new `provider_for_model()` (claude-* → anthropic, gpt-*/o3/o4/chatgpt → openai, gemini-* → google, open-weight families → local.ollama; `None` when unattributable). `_parse_provider()` now does: hidden-param provider → `provider/` prefix → `provider_for_model(model)` → `"unknown"`. Downstream `_provider_to_billing_account("anthropic")` → `anthropic` → `plan_tier` resolves from declared `[budget.anthropic]`. `"unknown"`/`"litellm"` both map to NULL billing_account, and neither has a pricing table.

The issue suggested checking `core/pricing.py` for existing model→provider knowledge — there was none there (only adapter-private copies in `core/ingest_adapters/langfuse.py` and `core/backfill.py`, which carry adapter-specific defaults). I added the canonical helper to `pricing.py` and left those adapter copies in place (noted in the helper docstring as a future consolidation point).

## What's also in this PR (adjacent, same defect)
`tokenjam/sdk/client.py` had a **second** `_parse_provider` with the identical `return "litellm"` bug (the `TokenJamClient.emit_litellm_span` path used by the upstream LiteLLM named callback). Fixed identically so acceptance criterion 2 ("`litellm` is never written as a final provider value") holds repo-wide.

## Tests / Verification
New `tests/unit/test_provider_attribution_194.py` (factory-based):
- `provider_for_model` resolution table + unresolvable → `None`.
- `_parse_provider` with `custom_llm_provider=None` + bare `claude-haiku-4-5` → `anthropic`; unresolvable → `unknown`, never `litellm`; hidden-param and prefix paths still win.
- Correct **Haiku pricing** ($0.80/MTok) vs the `litellm`/default $0.50 undercount.
- `billing_account` derivation; `"litellm"`/`"unknown"` → NULL.
- **plan_tier propagation** through `IngestPipeline` from declared config (resolved → `api`; unresolved → `unknown`).
- Updated the two tests that codified the old `"litellm"` behavior (`test_litellm_integration.py`, `test_litellm_client.py`).

`pytest tests/unit tests/synthetic tests/agents tests/integration` → **845 passed**. `ruff` + `mypy` clean on all touched files.

## What's NOT in this PR
- The content/cache-token capture gap on the same integration (#195) — deliberately untouched (same file; sequenced as a separate PR after this merges to avoid conflicts).
- Consolidating the Langfuse/backfill model→provider helpers into `provider_for_model` — left as a noted future cleanup to keep this scoped.

Closes #194

### DIFF
--- a/tests/unit/test_litellm_client.py
+++ b/tests/unit/test_litellm_client.py
@@ -119,7 +119,8 @@ def test_build_span_missing_model_defaults_to_unknown():
     )
     attrs = _attrs_to_dict(span["attributes"])
     assert attrs[GenAIAttributes.REQUEST_MODEL] == "unknown"
-    assert attrs[GenAIAttributes.PROVIDER_NAME] == "litellm"
+    # Unresolvable model -> 'unknown' provider, never the bogus 'litellm' (#194).
+    assert attrs[GenAIAttributes.PROVIDER_NAME] == "unknown"
 
 
 # --- TokenJamClient HTTP behavior ------------------------------------------
--- a/tests/unit/test_litellm_integration.py
+++ b/tests/unit/test_litellm_integration.py
@@ -151,8 +151,37 @@ class TestLiteLLMIntegration:
         span = spans[0]
         assert span.attributes[GenAIAttributes.PROVIDER_NAME] == "together"
 
-    def test_provider_fallback_no_slash(self, tracer_and_spans):
-        """Model with no slash and no hidden params falls back to 'litellm'."""
+    def test_bare_model_no_hidden_resolves_provider(self, tracer_and_spans):
+        """Bare model + no custom_llm_provider is resolved from the model name (#194).
+
+        Aider hits exactly this: LiteLLM >= 1.75 returns custom_llm_provider=None
+        and the model is bare (no ``anthropic/`` prefix). The provider must
+        resolve to ``anthropic`` (never the bogus ``"litellm"``) so pricing and
+        billing_account are correct.
+        """
+        tracer, spans = tracer_and_spans
+        import litellm
+
+        def no_provider(*a, **kw):
+            resp = MagicMock()
+            resp.usage.prompt_tokens = 10
+            resp.usage.completion_tokens = 5
+            resp._hidden_params = {"custom_llm_provider": None}
+            resp.model = "claude-haiku-4-5-20251001"
+            return resp
+        litellm.completion = no_provider
+
+        from tokenjam.sdk.integrations.litellm import LiteLLMIntegration
+        integration = LiteLLMIntegration()
+        integration.install(tracer)
+
+        litellm.completion(model="claude-haiku-4-5", messages=[])
+
+        span = spans[0]
+        assert span.attributes[GenAIAttributes.PROVIDER_NAME] == "anthropic"
+
+    def test_unresolvable_bare_model_is_unknown_not_litellm(self, tracer_and_spans):
+        """An unattributable bare model yields 'unknown', never 'litellm' (#194)."""
         tracer, spans = tracer_and_spans
         import litellm
 
@@ -161,7 +190,7 @@ class TestLiteLLMIntegration:
             resp.usage.prompt_tokens = 10
             resp.usage.completion_tokens = 5
             resp._hidden_params = {}
-            resp.model = "gpt-4o-mini"
+            resp.model = "some-internal-model-x"
             return resp
         litellm.completion = no_provider
 
@@ -169,10 +198,11 @@ class TestLiteLLMIntegration:
         integration = LiteLLMIntegration()
         integration.install(tracer)
 
-        litellm.completion(model="gpt-4o-mini", messages=[])
+        litellm.completion(model="some-internal-model-x", messages=[])
 
         span = spans[0]
-        assert span.attributes[GenAIAttributes.PROVIDER_NAME] == "litellm"
+        assert span.attributes[GenAIAttributes.PROVIDER_NAME] == "unknown"
+        assert span.attributes[GenAIAttributes.PROVIDER_NAME] != "litellm"
 
     def test_sync_streaming(self, tracer_and_spans):
         tracer, spans = tracer_and_spans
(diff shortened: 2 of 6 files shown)


anilmurty (commented):
## Verdict: ✅ APPROVE on the #194 fix — one judgment call to decide before merge

Reviewed the diff independently; verified the root in code prior. CI all 4 green (845 tests), MERGEABLE.

### Per-criterion (#194)
1. **Provider resolved from bare model name** — new `provider_for_model()` in `core/pricing.py` (claude→anthropic, gpt/o1/o3/o4/chatgpt→openai, gemini→google, open-weight→local.ollama). ✅
2. **"litellm" never written as final value** — both call sites now `provider_for_model(model) or "unknown"`. Caught and fixed the **second copy** in `sdk/client.py` (`emit_litellm_span` named-callback path) too — necessary for the criterion to hold repo-wide, correctly flagged as "also in this PR." ✅
3. **billing_account + plan_tier propagate** — covered by the new test through IngestPipeline. ✅
4. **Tests** — `test_provider_attribution_194.py` (factory-based: resolution, Haiku-vs-default pricing, billing_account, plan propagation); the two tests that codified the old "litellm" behavior updated. ✅

Good scope discipline: left #195 (same file) untouched, no parallel branch, and **resisted consolidating** the Langfuse/backfill private model→provider maps (flagged for future, correct call).

### One decision before merge — the open-weight default (honesty discipline)
`provider_for_model` maps bare open-weight names (`llama`/`qwen`/`mistral`/`gemma`/`deepseek`/…) → `local.ollama`, which sets `pricing_mode = local` → the UI/CLI assert **"Local inference — no marginal cost."** That's correct for a real Ollama user, but **wrong (and an over-claim) for someone running those models on a *paid* host** (Groq / Together / Bedrock / Fireworks) — it tells a paying user they have zero cost.

- It's an **edge case** (only reached when `custom_llm_provider` is None *and* the model is bare; most hosted open-weight calls set `custom_llm_provider` and never reach this branch).
- But asserting "free" for possibly-paid inference is exactly the over-claim the honesty discipline forbids. The conservative default is **`None` → "unknown"** (dollars shown with a "may overstate" qualifier) rather than a confident "free."
- Counter-view: bare-open-weight-with-no-provider most often *is* local Ollama, where "unknown→dollars" would wrongly show cost for free inference.

My recommendation: **change the open-weight branch to return `None` (→ "unknown")** — when unsure, don't assert free. It's a one-line change in the still-open PR, cleaner than a fast-follow. But it's genuinely debatable and edge-case, so it's your call.

**Ready to merge as-is on #194's merits; flagging this so you decide whether to amend the open-weight line first.**

PR #202, opened 2026-06-22
Fix #197: CLI and Lens render identical plan-tier framing (window-independent mix)

### LINKED ISSUE #197: [bug] CLI and Lens disagree on plan-tier framing for the same DB
## Symptom

For the **same** local DuckDB, the CLI and the web UI render contradictory plan-tier framing:
- `tj cost` / `tj optimize` (CLI) **suppress** dollar figures (treating sessions as unknown/subscription).
- The Lens web UI shows **"API billing" + dollar amounts** for the same data.

Two surfaces, one dataset, opposite framing — the user gets told two different things.

## Root cause (hypothesis — investigate)

Surfaced during the Aider pilot, where `billing_account` was NULL / `plan_tier` unknown (a state created by #194). Two possibilities:
1. **Downstream of #194:** CLI and API simply handle the unknown/NULL plan-tier case *differently*, and fixing #194 (so plan/billing resolve) makes them agree.
2. **Genuine divergence:** the CLI and the API/Lens consume `core/framing.py` inconsistently (or one re-derives). Per the #110 design, `core/framing.py` is the single source of truth and both surfaces must consume it without re-deriving — if they diverge on identical data, that's a real bug.

**Sequencing:** re-check after #194 lands. This may resolve on its own; if it doesn't, it reveals an independent framing-consumption divergence worth fixing. Filed now to capture it — close if #194 resolves it.

## Acceptance criteria

1. CLI and Lens render **identical** framing (units + qualifier) for the same DB / window / plan-tier state — including the unknown/NULL case.
2. Both consume `core/framing.py`; neither re-derives the suppression rules.
3. A test that pins CLI and API framing to the same result for a shared fixture across the api / subscription / unknown matrix.

## Severity

Medium. Honesty/consistency — two surfaces giving the user contradictory dollar framing undercuts the same discipline the recent framing sweep (#175/#177/#187/#191) enforced.

Discovered during a real-agent instrumentation pilot. Likely entangled with #194.

anilmurty:
For the same local DuckDB, `tj cost` (CLI) could suppress dollar figures while the Lens web UI showed "API billing" + dollars — two surfaces, one dataset, opposite framing. Confirmed (issue comment, 2026-06-22) as an independent framing-consumption divergence, not gated on #194.

## Summary
- `_cost_framing` and `_compare_framing` (`tokenjam/cli/cmd_cost.py`) now decide framing from the **window-independent** `plan_determination_mix(conn, agent)` — the same basis as `api/routes/cost.py` — instead of the window-scoped `plan_tier_mix(conn, since_dt, until_dt, agent)`.
- Window totals (cost / tokens / `--compare` deltas) stay window-scoped; only the tokens-vs-dollars unit + qualifier are plan-wide.
- New integration test pins CLI framing to the live `/api/v1/cost` framing across api / subscription / unknown.

## Symptom → root cause → fix
- **Symptom:** CLI suppressed dollars (unknown/subscription framing) while Lens showed dollars, for the same DB.
- **Root cause:** the pricing mode is a property of the user's **plan**, not the selected window. `/api/v1/cost` moved to a window-independent mix in #177/#186/#187; the CLI's direct-conn path was left on the old **window-scoped** mix. When the in-window session mix differed from full history (e.g. a 24h window whose sessions all *started* earlier but have recent spans), the two computed a different `pricing_mode`/suppression. Daemon-off only — with `tj serve` up, the CLI shims to the API's `framing` block and they already agreed.
- **Fix:** use `plan_determination_mix` for the framing **decision** in both CLI functions, mirroring the API construction exactly (`sessions = sum(mix.values())`, totals window-scoped). Both surfaces consume `core/framing`; neither re-derives the suppression rules.

## Scope note (left as-is, flagged for awareness)
`tj optimize` / `tj report` also use the window-scoped `plan_tier_mix`, **but so does `/api/v1/optimize`** — so the CLI and Lens *agree* on the optimize surface; it is not a CLI-vs-Lens divergence and is out of scope for #197. Whether the optimize surface should also adopt window-independent framing (for cost/optimize internal consistency) is a separate design question for triage, not a bug fix here.

## Tests / Verification
New `tests/integration/test_cli_api_framing_parity.py`: seeds 20 sessions started 5 days ago (outside a 24h window) with spans in the last hour, so the window-scoped mix is empty while full history carries the real plan — the exact divergence. It asserts the CLI `_cost_framing` decision fields (`pricing_mode`, `plan_tier`, `plan_label(s)`, `display_rule`, `qualifier_text`, share pcts) equal the live `/api/v1/cost` `framing` block across **api / subscription / unknown**.

- Verified the test **fails on pre-fix code** (CLI collapses to the empty-window "api" default — wrong mode/labels/share-pcts) and passes after.
- `pytest tests/unit tests/synthetic tests/agents tests/integration` → **854 passed**. `ruff` + `mypy` clean on touched files.

## What's NOT in this PR
- `tj optimize` / `tj report` framing basis (see scope note — not a CLI-vs-Lens divergence; deferred to triage).
- `litellm.py` (#195, owned by a parallel agent).

Closes #197

### DIFF
--- a/tests/integration/test_cli_api_framing_parity.py
+++ b/tests/integration/test_cli_api_framing_parity.py
@@ -1,0 +1,118 @@
+"""#197 — `tj cost` (CLI) and `/api/v1/cost` (Lens) must render IDENTICAL
+plan-tier framing for the same DB / window / plan state.
+
+Both surfaces consume ``core/framing``, but the CLI direct-conn path framed off
+a window-SCOPED session mix while the API moved to a window-INDEPENDENT mix
+(#177). When the in-window mix differed from full history (daemon-off), they
+disagreed — the CLI suppressed dollars while Lens showed them, or vice-versa.
+
+These tests seed sessions that STARTED outside a 24h window but have recent
+spans, so the window-scoped mix is empty while full history carries the real
+plan — the exact divergence. They then assert the CLI's ``_cost_framing`` and
+the live ``/api/v1/cost`` framing block agree across api / subscription /
+unknown. (Pre-fix, the subscription and unknown cases fail.)
+"""
+from __future__ import annotations
+
+import dataclasses
+from datetime import timedelta
+from types import SimpleNamespace
+
+import httpx
+import pytest
+
+from tokenjam.api.app import create_app
+from tokenjam.cli.cmd_cost import _cost_framing
+from tokenjam.core.config import ProviderBudget, StorageConfig, TjConfig
+from tokenjam.core.db import DuckDBBackend
+from tokenjam.core.ingest import IngestPipeline
+from tokenjam.utils.time_parse import utcnow
+from tests.factories import make_llm_span, make_session
+
+# The framing fields that express "units + qualifier" (AC #1) — the decision,
+# not the window-scoped dollar totals carried alongside it.
+_DECISION_FIELDS = (
+    "pricing_mode",
+    "plan_tier",
+    "plan_label",
+    "plan_labels",
+    "display_rule",
+    "qualifier_text",
+    "subscription_share_pct",
+    "api_share_pct",
+)
+
+
+@pytest.fixture(autouse=True)
+def _isolate_home(monkeypatch, tmp_path):
+    """Point Path.home() at an empty dir so config_declared_plan's global
+    fallback never reads the dev machine's ~/.config/tj/config.toml."""
+    monkeypatch.setattr("pathlib.Path.home", lambda: tmp_path)
+
+
+def _seed_out_of_window(db, plan_tier: str) -> None:
+    """20 sessions that STARTED 5 days ago (outside a 24h window) but each have
+    a span in the last hour. Window-scoped session mix → empty; full-history
+    mix → {plan_tier: 20}. That gap is exactly what made the CLI and API
+    disagree daemon-off (#197)."""
+    now = utcnow()
+    for i in range(20):
+        s = make_session(session_id=f"s-{i}", plan_tier=plan_tier)
+        s = dataclasses.replace(s, started_at=now - timedelta(days=5))
+        db.upsert_session(s)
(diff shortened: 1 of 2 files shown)


anilmurty (commented):
## Verdict: ✅ APPROVE — ready to merge

Reviewed independently against the confirmed root cause. CI all 4 green (854 tests), MERGEABLE, tight scope (`cmd_cost.py` + one parity test, 138/4).

### Per-criterion (#197)
- `_cost_framing` (and `_compare_framing`) now use `plan_determination_mix(conn, agent)` — **window-independent** — for the framing decision, mirroring `api/routes/cost.py`. Window totals/deltas stay window-scoped. Both consume `core/framing`; neither re-derives. CLI and Lens now agree on the same DB, daemon-on and daemon-off. ✅
- New `test_cli_api_framing_parity` pins CLI `_cost_framing` to the live `/api/v1/cost` framing across api/subscription/unknown; verified fails-before / passes-after. ✅

### Verified the scope decision independently
Confirmed `/api/v1/optimize` (`optimize.py:108`) uses **window-scoped** `plan_tier_mix` too — so leaving `tj optimize` window-scoped keeps CLI-optimize == Lens-optimize (no divergence there). Correctly not touched, and the "should optimize also be window-independent?" question correctly flagged rather than silently expanded. Good discipline.

### One follow-up the PR surfaced (tracked for master triage, not blocking)
After this, the **cost** surfaces frame off the window-independent mix while the **optimize** surfaces frame off the window-scoped mix — so within the product, cost-view and optimize-view can show different framing for the same user/window. It's *not* a CLI-vs-UI contradiction (each surface agrees with its counterpart), so it doesn't block this. But it's the same #177 principle (framing as a plan property), so whether optimize should also go window-independent is a real design question for separate triage.

**Ready to merge.** Stayed out of `litellm.py` (no #195 collision).

PR #274, opened 2026-06-24
Fix #242: raise default traces limit from 50 to 200

### LINKED ISSUE #242: [Lens] Traces endpoint 50-row silent cap makes 24h and 7d look identical
On the Traces screen, switching the window from **24h to 7d shows the same number of rows** (1h shows fewer). Cause: `list_traces` defaults to **`limit=50`** (`tokenjam/api/routes/traces.py:45`). Any window with >50 traces returns exactly 50, so 24h and 7d are indistinguishable.

This violates the "no silent caps" principle — the user can't tell the window filter is working, and assumes 7d has no more data than 24h.

## Fix options (pick)
- Surface the cap in the UI: "showing latest 50 of N" + a **Load more** / pagination control, or
- Raise/parameterize the limit with an explicit "showing N" count.

Either way the row count should reflect the selected window (or clearly say it's truncated). UI + endpoint; respect plan-tier framing on the cost column as today.

sanmaxdev:
The default `limit=50` silently capped trace results across all time windows. When the user switched from 24h to 7d on the Traces screen, both returned exactly 50 rows when >50 traces existed in each window — making them indistinguishable.

## Changes
- Raise the default limit from 50 to 200 so callers relying on the default (and direct REST API users) get meaningful visibility across typical debugging windows.

## Verification
- `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` — 1005 passed
- `ruff check tokenjam/api/routes/traces.py` — clean
- `mypy tokenjam/api/routes/traces.py` — clean

Closes #242


anilmurty:
Hey @sanmaxdev - Thanks so much for jumping on this!

Raising the default to 200 definitely helps in the common case. The heart of #242, though, is that the cap is silent rather than that it's small: a user can't tell they're looking at a truncated slice, and even at 200 a busy 7-day window still looks identical to 24h (both just hit the ceiling). So we'd love to turn this into the complete fix: surface the truncation rather than just raise the limit.

Good news: the endpoint already takes an offset, so most of the pieces are there. The full fix is roughly:

1. Endpoint (tokenjam/api/routes/traces.py) should also return a total count for the window, so the UI knows how many traces exist (not just the page it got).
2. UI (tokenjam/ui/index.html, the Traces list) should show "showing N of M" and a "Load more" button that pages via the existing offset.
3. A small static-grep regression in tests/unit/test_lens_ui_regression.py asserting the count + Load-more render — that's how the offline single-file UI is tested (no JS runner; the existing tests there show the pattern).

Two repo conventions to keep in mind: the UI is a single-file offline SPA (no new dependencies / no external fetches at render time), and any cost figures must go through the existing plan-tier framing helper. CONTRIBUTING.md and AGENTS.md have the details.

anilmurty (changes requested):
noted in the comment above

sanmaxdev:
Thanks for the clear direction. I pushed a follow-up that turns this into the full pagination fix:

- `/api/v1/traces` now returns `total_count` for the filtered window
- Lens shows `Showing N of M traces`
- Lens uses the existing `offset` parameter behind a `Load more` button
- added API and static UI regression coverage

Verification:
- `python3 -m pytest tests/integration/test_api.py::test_get_traces_returns_total_count_for_pagination tests/unit/test_lens_ui_regression.py::test_traces_list_surfaces_pagination -q` — 2 passed
- `ruff check tokenjam/api/routes/traces.py tokenjam/core/db.py tests/integration/test_api.py tests/unit/test_lens_ui_regression.py` — passed
- `mypy tokenjam/api/routes/traces.py tokenjam/core/db.py` — passed

anilmurty (commented):
Thanks @sanmaxdev — this is the complete fix for #242. Verified against the issue's fix options:

- The endpoint now returns `total_count` for the selected window (`count_traces(filters)` on the storage backend), and the UI surfaces **"Showing N of M traces"** + a **Load more** button with offset pagination — option (a), the fuller path.
- Default limit raised 50 → 200.
- Cost-column plan-tier framing left untouched.
- Tests cover both the endpoint (`total_count`) and the UI strings; all four CI jobs are green.

✅ Ready to merge.

PR #304, opened 2026-06-25
Fix #303: add long-window options to Traces

sanmaxdev:
## Summary

The Traces view already accepts `since=30d` and `since=90d` from the URL/API, but its window dropdown only exposed 1h, 24h, and 7d. This adds the missing options so deep links like `#/traces?since=30d` show a selected value instead of a blank control.

## Tests / Verification
- Added a Lens UI regression assertion for the Traces window options.
- `python3 -m pytest tests/unit/test_lens_ui_regression.py -q`
- `ruff check tokenjam/ tests/unit/test_lens_ui_regression.py`
- `mypy tokenjam/`
- `python3 -m pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/ -q`

Closes #303

PR #305, opened 2026-06-25
Fix #299: align recent activity trace window

### LINKED ISSUE #299: [Lens] Dashboard 'Recent activity: N' doesn't match the traces view it links to
The Dashboard "Recent activity / last sessions" tile shows **6**, but clicking through to Traces (default **Last 24h** filter) shows "Showing 4 of 4 traces". The tile counts recent sessions over one window; the traces view applies a different (24h) window, so the numbers don't reconcile and the drill-through is confusing.

## Fix
- Make the tile's click land on a traces view filtered to **match the tile's window** (so the count you clicked equals what you land on), and/or
- Align the counting basis (sessions vs traces, same window) between the tile and the traces list.

UI + possibly the window the tile's count is computed over.

sanmaxdev:
The Dashboard Recent activity tile counted traces with the selected Dashboard window, but its drill-through opened the Traces default window. This makes the drill-through carry the same window so the destination list matches the tile basis.

## Summary
- Add a small Traces href helper that preserves the Dashboard `since` window.
- Point the Recent activity tile at that window-aware href.
- Add a static Lens regression test for the drill-through.

## Fix #299
The Dashboard defaults to `30d`, while Traces defaults to `24h`. The tile now links to `#/traces?since=30d` for the Dashboard default, and keeps `#/traces` clean when the selected window is already the Traces default (`24h`).

## Tests / Verification
- `ruff check tokenjam/ tests/unit/test_lens_ui_regression.py`
- `mypy tokenjam/`
- `pytest tests/unit/test_lens_ui_regression.py -q`
- `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/ -q`
- `node --check /tmp/tokenjam-module.js`
- `git diff --check`

Closes #299

### DIFF
--- a/tests/unit/test_lens_ui_regression.py
+++ b/tests/unit/test_lens_ui_regression.py
@@ -29,6 +29,15 @@ def test_traces_window_select_exposes_longer_supported_windows(html):
     assert '<option value="90d">Last 90d</option>' in traces_view
 
 
+def test_dashboard_recent_activity_drills_into_matching_traces_window(html):
+    # The Dashboard defaults to 30d while Traces defaults to 24h. Keep the Recent
+    # activity drill-through tied to the Dashboard window so the tile count and
+    # destination list use the same basis (#299).
+    assert "function tracesHrefForWindow" in html
+    assert "tracesHrefForWindow(since)" in html
+    assert 'label="Recent activity" value=${(d.traces || []).length} attention=${errTraces > 0} href="#/traces"' not in html
+
+
 # --- #126: Downsize typed slot always rendered ----------------------------- #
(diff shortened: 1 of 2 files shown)


anilmurty (commented):
Thanks @sanmaxdev, fourth one! Exactly the fix for #299. The Recent activity tile now drills into the same window it counts over (Dashboard's 30d), so the count and the destination list reconcile. Nice touches: the helper omits the param when it's already the Traces default (matching the omit-defaults convention), and it composes cleanly with the 30d/90d options from #304. Verified 'since' is correctly in scope in DashboardView, regression test guards it, all four CI jobs green.

Ready to merge.

Issue #308, opened 2026-06-25
[enhancement] Surface sample size + confidence interval on analyzer savings estimates


anilmurty:
Analyzers that carry a savings estimate (e.g. `downsize` monthly projection, the recoverable-savings figures) surface a single number with **no indication of how much data it's based on**. A projection computed from 5 sessions and one computed from 500 sessions render identically, even though the 5-session estimate is far less reliable.

This is a sampling-confidence gap (distinct from quality/validation confidence), and surfacing it directly reinforces the honesty discipline: an "estimated recoverable" figure should carry its own reliability signal.

## Proposal
- Add `n_sessions` (the sample the estimate is drawn from) and a confidence interval (`ci_low` / `ci_high`) to each savings-bearing result dataclass.
- Compute a simple interval over the window sessions — a pure-Python bootstrap or Wilson-style interval is sufficient; no new heavy dependency required.
- Surface `n` and the interval in the CLI savings line and the dashboard tiles (e.g. "~$X/mo recoverable (n=42, 95% CI $Y–$Z)").
- Round-trip the new fields through `report_to_dict` / `report_from_dict`.

## Honesty framing
This is **sampling** confidence on the projection, not a claim that the optimization preserves quality. Keep the existing "estimated recoverable / review before switching" caveats; this only quantifies how much usage the estimate rests on.

PR #314, opened 2026-06-25
[chore] Complete StorageBackend protocol, remove direct db.conn access (#309)

### LINKED ISSUE #309: [chore] Complete the StorageBackend protocol — remove direct db.conn access in CostEngine / cmd_status
The `StorageBackend` protocol (`tokenjam/core/db.py`) doesn't cover every query, so `CostEngine`, `cmd_status`, and a few other callers reach into `db.conn` directly. This breaks the abstraction boundary and means those paths can't be exercised against `InMemoryBackend` in unit tests (they're implicitly coupled to DuckDB).

## Proposal
- Add the missing methods to the `StorageBackend` protocol, e.g. `update_span_cost()`, `get_daily_cost_by_provider()`, `get_active_sessions()` (audit the actual direct-`db.conn` call sites for the full set).
- Move all direct `db.conn` access in `core/cost.py`, `cmd_status`, etc. behind those methods.
- Implement them on both `DuckDBBackend` and `InMemoryBackend` so the affected paths gain in-memory test coverage.

Engineering hygiene; no behavior change. Improves testability of the cost path specifically.

anilmurty:
`CostEngine`, `cmd_status`, and the cost-compare path satisfied a handful of queries by reaching into `db.conn` directly, bypassing the `StorageBackend` abstraction. That coupled those paths to DuckDB and meant they could only be tested against a hand-rolled stub exposing a raw connection — the cost path in particular had no real in-memory backend coverage. This completes the protocol and routes every call site through it, with no behavior change.

## Summary
- Added 8 methods to the `StorageBackend` protocol and implemented them on `DuckDBBackend` (so `InMemoryBackend`, which subclasses it, inherits all of them):
  - `update_span_cost`, `increment_session_cost` — `CostEngine` writes
  - `get_distinct_agent_ids`, `get_active_session`, `get_session_active_seconds`, `count_unknown_plan_tier_sessions` — `cmd_status` reads
  - `get_window_cost_totals`, `get_cost_delta_by_group` — back `compute_window_totals` / `compute_cost_diff`
- Moved every direct `db.conn` query in `core/cost.py` and `cmd_status` behind those methods.

## Audit of the call sites that were converted
| Site | Was | Now |
|------|-----|-----|
| `CostEngine.process_span` | `self.db.conn.execute(UPDATE spans/sessions …)` | `update_span_cost` / `increment_session_cost` |
| `cmd_status` agent list | `db.conn.execute(SELECT DISTINCT agent_id …)` | `get_distinct_agent_ids` |
| `cmd_status` active session | `db.conn.execute(SELECT * … status='active')` + `_row_to_session` | `get_active_session` |
| `cmd_status` active seconds | `session_active_seconds(db.conn, …)` | `get_session_active_seconds` |
| `cmd_status` unknown-plan count | `db.conn.execute(SELECT COUNT(*) …)` | `count_unknown_plan_tier_sessions` |
| `compute_window_totals` | took a bare `conn` | takes `db`, calls `get_window_cost_totals` |
| `compute_cost_diff` | `getattr(db, "conn")` + inline SQL | `get_cost_delta_by_group` |

## The win — real in-memory coverage of the cost path
- `tests/synthetic/test_cost_tracking.py` drops its `FakeDB`-with-`.conn` stub and runs `CostEngine` against `InMemoryBackend`, reading span cost back via `get_trace_spans` and the session total via `get_session().total_cost_usd`. All original scenarios preserved (span/session cost, accumulation, cache-only / cache-write pricing, pre-priced no-double-count, no-op guards).
- `tests/unit/test_db_helpers.py` gains direct coverage of each new method (incl. the `get_cost_delta_by_group` allow-list rejecting an unknown column).

## Guardrails honored
- **Per-thread cursor (#124) untouched** — `conn` stays a `threading.local` per-thread cursor; the new methods call `self.conn` from inside the backend, they don't change its shape.
- **No collision with #308** — `core/optimize/` and `cmd_optimize.py` are not modified. `compute_cost_diff` / `compute_window_totals` keep stable public signatures, so `cmd_optimize`, `cmd_cost`, and the `cost_compare` route are unaffected (their `compute_cost_diff` calls are already guarded to the local-DB backend).
- `cmd_status` keeps its `hasattr(db, "conn")` mode discriminator (local DB vs API backend) — an attribute-existence check, not a query.

## Tests / Verification
- `pytest tests/unit tests/synthetic tests/agents tests/integration` → **1137 passed**.
- `ruff check` and `mypy` clean on `core/db.py`, `core/cost.py`, `cmd_status.py`.

## What's NOT in this PR
- `session_active_seconds` and `_row_to_session` remain module-level functions in `db.py` — still used by `api/routes/status.py` (outside this issue's ownership) and `tests/unit/test_db_helpers.py`. The new `get_session_active_seconds` wraps the former.
- `ApiBackend` was not given the new methods — the converted call sites only invoke them in local-DB mode, which is the issue's scope (DuckDB + InMemory).

Closes #309

### DIFF
--- a/tests/synthetic/test_cost_tracking.py
+++ b/tests/synthetic/test_cost_tracking.py
@@ -1,104 +1,85 @@
-"""Synthetic tests for CostEngine with an in-memory DuckDB backend."""
+"""Synthetic tests for CostEngine against the real InMemoryBackend.
+
+Previously these ran against a hand-rolled `FakeDB` stub that exposed a raw
+`.conn` — which only worked because CostEngine reached into `db.conn` directly.
+Issue #309 moved those writes behind the StorageBackend protocol
+(`update_span_cost` / `increment_session_cost`), so the cost path now exercises
+the same backend production uses. Span cost is read back through
+`get_trace_spans`; the session total through `get_session().total_cost_usd`
+(the stored column, distinct from `get_session_cost`, which sums span costs).
+"""
 from __future__ import annotations
 
-import duckdb
 import pytest
 
 from tokenjam.core.cost import CostEngine
+from tokenjam.core.db import InMemoryBackend
 from tests.factories import make_llm_span, make_session
 
 
-class FakeDB:
-    """Minimal DB stub with just enough schema for CostEngine tests."""
-
-    def __init__(self) -> None:
-        self.conn = duckdb.connect(":memory:")
-        self.conn.execute("""
-            CREATE TABLE spans (
-                span_id TEXT PRIMARY KEY,
-                cost_usd DOUBLE
-            )
-        """)
-        self.conn.execute("""
-            CREATE TABLE sessions (
-                session_id TEXT PRIMARY KEY,
-                total_cost_usd DOUBLE
-            )
-        """)
-
-    def insert_span_stub(self, span_id: str) -> None:
-        self.conn.execute(
-            "INSERT INTO spans (span_id, cost_usd) VALUES (?, NULL)",
-            [span_id],
-        )
-
-    def insert_session_stub(self, session_id: str) -> None:
-        self.conn.execute(
-            "INSERT INTO sessions (session_id, total_cost_usd) VALUES (?, NULL)",
-            [session_id],
-        )
-
-    def get_span_cost(self, span_id: str) -> float | None:
-        row = self.conn.execute(
-            "SELECT cost_usd FROM spans WHERE span_id = ?", [span_id]
-        ).fetchone()
-        return row[0] if row else None
-
-    def get_session_cost(self, session_id: str) -> float | None:
-        row = self.conn.execute(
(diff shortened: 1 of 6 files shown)


anilmurty (commented):
Verified against #309. The 8 methods land on the StorageBackend protocol and are implemented once on DuckDBBackend, which InMemoryBackend inherits (and runs, since it's :memory:-backed) — that's the design that lets test_cost_tracking exercise the real CostEngine path in-memory instead of a FakeDB stub. Both guardrails honored: the threading.local per-thread cursor (#124) is untouched, and core/optimize is clean (no #308 collision). No f-string SQL, parameterized throughout. Good scope discipline on _row_to_session and ApiBackend. All four CI jobs green.

Ready to merge.

PR #315, opened 2026-06-25
fix(lens): gate Analytics Stack control to stacking charts (#295)

### LINKED ISSUE #295: [Lens] Analytics 'Stack' dropdown returns 'No data' for most combinations
In the Analytics / Explore pivot (Metric × Dimension × **Stack** × Chart), choosing a Stack dimension yields "No data in this window" for most combinations — e.g. Metric=Events, By=Model, Stack=Tool category, Chart=Leaderboard renders empty. The Stack control is confusing: it produces empty cross-tabs for many metric×dimension pairs, and stacking isn't meaningful for some chart types (e.g. Leaderboard).

## Fix options (pick)
- Gate the Stack control to chart types where stacking applies (stacked bar/area), hiding it otherwise; and/or
- Only offer Stack dimensions that actually produce data for the current Metric×By; and/or
- Render a clear "no rows for this breakdown" state instead of a generic empty chart.

UI-only (single-file SPA). Whichever path, the Stack control should never sit there returning nothing for the default selections.

ashwmu:
Fixes #295.

## Problem
In the Analytics / Explore pivot, choosing a **Stack** dimension returned "No data in this window" for many combinations (e.g. Metric=Events, By=Model, Stack=Tool category, Chart=Leaderboard). Two root causes:

1. **Stacking is meaningless on the Leaderboard (`hbar`).** It collapses the time dimension into a single per-group total and ignores `stack` entirely — yet the query still sent `stack_by`, producing an empty cross-tab and stranding the view on "No data".
2. **Many cross-tabs are structurally empty** because a span carries only one dimension — e.g. an LLM call has a `model` but no `tool`; a tool span has a `tool` but no `model`. So `Model × Tool category` has zero rows.

## Fix (UI-only, `tokenjam/ui/index.html`)
- **Gate the Stack control to stacking charts (bar/line).** It's hidden on the Leaderboard.
- **Drop `stack_by` from the `/analytics` query for non-stacking charts**, so a stale stack in the URL can't strand the leaderboard on "No data". The chosen stack is preserved in the URL and returns when switching back to bar/line.
- **Clear empty state for empty cross-tabs.** On bar/line, a structurally-empty breakdown now shows *"No rows for X × Y … "* with a **Clear stack** button instead of a bare "No data in this window".

## Verified
- Leaderboard + stale `stack=tool_category` → Stack control hidden, model leaderboard renders (was "No data").
- Bars + `Model × Tool category` → "Clear stack" empty state.
- Bars + `Tool × Tool category` (valid cross-tab) → real stacked bar chart.

## Tests
- Static regression guards in `tests/unit/test_lens_ui_regression.py` (proven to fail on the buggy UI, pass on the fix).
- Full UI guard suite green: `test_lens_ui_regression.py` + `test_ui_offline.py` (115 passed).

### DIFF
--- a/tests/unit/test_lens_ui_regression.py
+++ b/tests/unit/test_lens_ui_regression.py
@@ -468,6 +468,29 @@ def test_analytics_leaderboard_has_inline_bars(html):
     assert ".lb-bar" in html
 
 
+# --- #295: Stack gated to stacking charts; empty cross-tab gets a clear state - #
+def test_analytics_stack_gated_to_stacking_charts(html):
+    # Stack only applies to the multi-series charts (bar/line). The Leaderboard
+    # (hbar) ignores stack, so the control is hidden AND stack_by is dropped from
+    # the query for non-stacking charts — otherwise a stale stack strands the
+    # leaderboard on an empty cross-tab ("No data", #295).
+    assert "const stackApplies = chart === 'bar' || chart === 'line'" in html
+    assert "const effStack = stackApplies ? stack : ''" in html
+    # query drops stack_by when the chart doesn't stack
+    assert "stack_by: effStack || undefined" in html
+    assert "stack_by: stack || undefined" not in html  # the buggy unconditional form is gone
+    # the Stack control is conditionally rendered (hidden on the leaderboard)
+    assert "${stackApplies ? html`<label class=\"ctl\">Stack" in html
+
+
+def test_analytics_empty_cross_tab_offers_clear_stack(html):
+    # A structurally-empty stacked breakdown (e.g. Model x Tool category, since a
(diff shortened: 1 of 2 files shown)


anilmurty (commented):
Thanks @ashwmu — clean fix for #295, and the diagnosis is spot on: the leaderboard ignores stack entirely, and Model × Tool category is structurally empty since a span carries one or the other. Gating the Stack control to bar/line and dropping stack_by from the query for non-stacking charts is the right call, and preserving the stack in the URL so it returns on switch-back is a nice touch. The regression tests are thorough (assert the new gating is present AND the buggy unconditional stack_by form is gone). Verified the module JS is syntactically clean (node --check on the real 147KB block), and all four CI jobs pass.

Ready to merge.

PR #316, opened 2026-06-25
[enhancement] Surface sample size + confidence interval on savings estimates (#308)

### LINKED ISSUE #308: [enhancement] Surface sample size + confidence interval on analyzer savings estimates

anilmurty:
A savings projection drawn from 5 sessions rendered identically to one drawn from 500, even though the small-n estimate is far less reliable. This attaches a **sampling** confidence signal to the `downsize` projection so the figure carries its own reliability indicator — directly reinforcing the honesty discipline.

## Summary
- New pure-Python `tokenjam/core/optimize/stats.py` — a seeded bootstrap CI (no scipy/numpy).
- `DowngradeFinding` gains `n_sessions`, `ci_low`, `ci_high`.
- The analyzer computes the interval over per-candidate-session window savings.
- Round-trips through `report_to_dict` / `report_from_dict`.
- The CLI savings line surfaces `(n=…, 95% CI $Y–$Z/mo)`.

## How it works
`bootstrap_ci(per_session, scale=…)` resamples the candidate sessions with replacement, recomputes the scaled total each time, and takes the central 95% quantiles — capturing how much the projection would swing if the same number of sessions had landed differently. A **fixed seed** keeps a report from jittering between runs; it returns `None` below 2 samples (a single point has no spread). `model_downgrade.py` collects each candidate session's `(actual − cheaper-model)` saving and computes the interval on the **monthly** projection using the same `30 / window_days` scale as `monthly_savings_usd`, so the CI brackets that exact figure.

### CLI render (api mode)
```
     • Projected savings if pattern holds: $0.4000/mo  (n=8 sessions, 95% CI $0.3200–$0.4700/mo — sampling confidence, not a safety claim)
     ! Candidate-flagging heuristic, not a quality judgment. Review the example sessions before changing models.
```

## Honesty framing (Critical Rule 14 — load-bearing)
This is **sampling** confidence on the projection (how much usage it rests on), **not** a claim the model swap preserves quality. The existing `MODEL_DOWNGRADE_CAVEAT` and "estimated recoverable / review before switching" framing are **untouched**, and the CI label is worded so it can't read as a safety/validation claim — a test asserts the suffix says "sampling confidence … not a safety claim" and never says "safe to" / "validated" / "guarantee" / "preserves quality".

## Tests (`tests/unit/test_optimize_savings_ci.py`, all via factories)
- bootstrap: None below 2 samples, deterministic, brackets the point estimate, scale projects correctly;
- **the load-bearing property**: a 5-session run yields a visibly WIDER interval than a 500-session run drawn from the same per-session distribution (same point estimate);
- the analyzer populates `n_sessions` + CI, and a single candidate session surfaces `n` with CI `None`;
- the fields round-trip through the report dict, and older payloads without the keys still parse (defaults);
- CLI suffix: shows n + interval, reads as sampling-not-safety, surfaces `n` alone when too thin, empty when n=0.

`pytest tests/unit tests/synthetic tests/agents tests/integration` = **1142 passed**; `ruff` + `mypy` clean.

## Ownership / collision (per the brief)
Only `core/optimize/` (new `stats.py`, `types.py`, `model_downgrade.py`, `runner.py`) + `cmd_optimize.py` are touched. **`core/db.py` / `core/cost.py` / `cmd_status` (#309) are NOT modified** — confirmed clean in the diff.

## What's NOT in this PR
- **Dashboard tile (index.html)** — explicitly optional in the brief. The dashboard's recoverable tiles aggregate across analyzers via a *different* path (`/cost/components` recoverable), not the `DowngradeFinding` object, so surfacing per-finding sampling CI there would need separate server plumbing — disproportionate to "keep it to one tile, minimize merge surface." Left for a focused follow-up. The required CLI + data layer is complete and the fields are already on the serialised report for any UI consumer.
- **CI on the other savings-bearing findings** (`reuse`, etc.) — the issue's example and the brief center on the `downsize` monthly projection; `reuse`'s per-cluster recoverable is a different shape (cache-reuse vs script-replacement bounds) and is best handled separately to avoid conflating two interval semantics.

Closes #308

### DIFF
--- a/tests/unit/test_optimize_savings_ci.py
+++ b/tests/unit/test_optimize_savings_ci.py
@@ -1,0 +1,218 @@
+"""Sampling confidence interval on savings estimates (#308).
+
+Covers the bootstrap helper, that the downgrade analyzer populates
+n_sessions + ci_low/ci_high, that the fields round-trip through the report dict,
+and the load-bearing property: a small-n run produces a visibly WIDER interval
+than a large-n run drawn from the same per-session distribution.
+
+All synthetic data is built via tests/factories (Critical Rule 8).
+"""
+from __future__ import annotations
+
+from datetime import timedelta
+
+import pytest
+
+from tokenjam.core.db import InMemoryBackend
+from tokenjam.core.optimize import analyze_model_downgrade
+from tokenjam.core.optimize.runner import report_from_dict, report_to_dict
+from tokenjam.core.optimize.stats import bootstrap_ci
+from tokenjam.core.optimize.types import OptimizeReport, WindowSummary
+from tokenjam.utils.time_parse import utcnow
+from tests.factories import make_llm_span
+
+
+@pytest.fixture
+def db():
+    backend = InMemoryBackend()
+    yield backend
+    backend.close()
+
+
+def _small_opus(db, session_id, *, cost=0.030):
+    """One Opus session matching the downgrade heuristic (small in/out, no tools)."""
+    db.insert_span(make_llm_span(
+        agent_id="claude-code-x", model="claude-opus-4-7", provider="anthropic",
+        input_tokens=1000, output_tokens=200, cost_usd=cost,
+        session_id=session_id, start_time=utcnow() - timedelta(days=2),
+    ))
+
+
+# --- bootstrap_ci (the pure helper) --------------------------------------- #
+
+def test_bootstrap_ci_none_below_two_samples():
+    assert bootstrap_ci([]) is None
+    assert bootstrap_ci([5.0]) is None
+
+
+def test_bootstrap_ci_is_deterministic():
+    # A report must not jitter between runs — same input, same interval.
+    vals = [0.5, 1.0, 1.5, 2.0, 0.8]
+    assert bootstrap_ci(vals) == bootstrap_ci(vals)
+
+
+def test_bootstrap_ci_brackets_the_point_estimate():
+    vals = [1.0, 1.2, 0.9, 1.1, 1.0, 1.3]
+    point = sum(vals)  # scale=1.0
+    lo, hi = bootstrap_ci(vals)  # type: ignore[misc]
+    assert lo <= point <= hi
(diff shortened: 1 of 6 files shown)


anilmurty (commented):
Verified against #308, with extra attention on the honesty framing since that's the whole point. The CLI label reads "sampling confidence, not a safety claim," and the guard test asserts both phrases are present and bans "safe to" / "validated" / "guarantee" / "preserves quality." That's the Rule 14 line held exactly: it quantifies how much usage the estimate rests on, never implies the swap is safe. Pure-Python seeded bootstrap (no scipy/numpy), round-trips through report_to_dict/from_dict with older payloads still parsing, and the small-n-yields-wider-interval property is tested. No #309 collision (core/db.py, cost.py, cmd_status untouched). All four CI jobs green.

Ready to merge.

PR #317, opened 2026-06-26
Fix #310: use utcnow instead of datetime.now in ApiBackend

### LINKED ISSUE #310: [bug] api_backend.py uses datetime.now() instead of utcnow() (Critical Rule 9)
`tokenjam/core/api_backend.py:235` uses a bare `datetime.now()`:

```python
if b.get("computed_at") else datetime.now()
```

Per Critical Rule 9, all timestamps must use `tokenjam.utils.time_parse.utcnow()` (timezone-aware UTC), never `datetime.now()` / `datetime.utcnow()`. A naive local-time `datetime.now()` here can produce an incorrect / tz-naive fallback timestamp.

## Fix
Import and use `utcnow()` from `tokenjam.utils.time_parse` for the fallback. Trivial one-line change; add/adjust a test if one covers this path.

tarun73:
Fix #310: use utcnow instead of datetime.now in ApiBackend

## Summary
- Replaced a timezone-naive `datetime.now()` fallback with `utcnow()` in `ApiBackend.get_baseline` to satisfy Critical Rule 9 (requiring timezone-aware UTC timestamps).
- Added unit tests covering both present and absent `computed_at` fields in the mock response.

## Tests / Verification
- Added unit test cases under `TestApiBackendGetBaseline` in `tests/unit/test_drift.py`.
- All 1,153 tests passed locally.

Closes #310

@anilmurty , please review

### DIFF
--- a/tests/unit/test_drift.py
+++ b/tests/unit/test_drift.py
@@ -76,3 +76,45 @@ class TestDriftDetectorAgentFallback:
 
         # Baseline should have been built despite no agent config
         assert db.upsert_baseline.called
+
+
+class TestApiBackendGetBaseline:
+    def test_get_baseline_computed_at_fallback(self, monkeypatch):
+        from tokenjam.core.api_backend import ApiBackend
+        from datetime import datetime
+        from tokenjam.utils.time_parse import utcnow
+
+        backend = ApiBackend("http://localhost:8000")
+
+        # Test case 1: computed_at is present in the API response
+        mock_data = {
+            "baseline": {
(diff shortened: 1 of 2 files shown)


anilmurty (commented):
Thanks @tarun73, and welcome — clean first PR. Correct fix for #310: utcnow() in place of the naive datetime.now() (utcnow is already imported, so no NameError), and the test genuinely guards it by asserting the fallback is timezone-aware and within a few seconds of now. All four CI jobs green.

Ready to merge.

Issue #209, opened 2026-06-22
[enhancement] Capture full request (sampling params + tools), not just message content, on LLM spans


anilmurty:
## Context

TokenJam captures LLM message **content** (#195) on spans, but not the rest of the request: **sampling parameters** (temperature, top_p, max_tokens, stop, etc.) and the **tools / tool_choice** payload. So a span doesn't carry enough to reconstruct the exact request that produced it.

This surfaced while building a replay-based experiment: faithful re-execution of a captured call required side-logging the full request separately, because the span alone (messages only) under-specifies it.

## Why it matters

A span that stores the full request is **self-contained and replayable** — you can re-run it against a different model/config from the captured telemetry alone. That's valuable for any replay/what-if analysis (and is a prerequisite for trustworthy offline replay generally). It also makes captured traces more complete for debugging and reproduction.

## Acceptance criteria

1. LLM spans capture sampling params (temperature, top_p, max_tokens, stop, etc.) and the tools / tool_choice payload, under the existing `[capture]` gating + `strip_captured_content()` discipline (so it's honored on/off like message content; no behavior change when capture is off).
2. Covers the in-process provider patches and the LiteLLM path (where this came up).
3. The captured fields round-trip through `NormalizedSpan` → DB and back.
4. Tests via `tests/factories.py`.

## Scope / out of scope

- In scope: extending span capture to the full request payload, gated by `[capture]`.
- Out of scope: any replay/validation engine itself (that's separate, non-public work) — this issue is purely the OSS capture-completeness improvement.

## Severity

Enhancement. Capture-completeness; unblocks replay/what-if use cases and improves trace fidelity.

PR #321, opened 2026-06-26
Fix #320: capture prompt + completion content on provider patches

### LINKED ISSUE #320: [capture] Provider patches capture #209 request_params but not the #195 prompt content (spans not replayable)
The full-request capture (#209) is incomplete for the **provider monkey-patches**. With `capture.prompts = true`, a `patch_anthropic()` (and by inspection `patch_openai`/`gemini`/`bedrock`) LLM span gets `request_params` populated, but **not** `gen_ai.prompt.content` (the messages). Only `sdk/agent.py` (the `@watch`/`record_llm_call` path) and `litellm.py` call `_record_prompt_content` / set `PROMPT_CONTENT`.

## Why it matters
A span without the prompt content isn't self-contained: you can't reconstruct or replay the request from it, which is the whole point of #209 ("makes a span self-contained enough to replay"). So `patch_anthropic` users get params-without-messages, a half-captured request.

## Repro
Run an Anthropic call through `patch_anthropic()` with `capture.prompts = true`, then query:
```sql
SELECT COUNT(*) FROM spans WHERE name='gen_ai.llm.call' AND request_params IS NOT NULL;        -- > 0
SELECT COUNT(*) FROM spans WHERE name='gen_ai.llm.call'
  AND json_extract_string(attributes,'$."gen_ai.prompt.content"') IS NOT NULL;                 -- 0
```
(Found while building a replay harness that reads captured spans; `patch_litellm` captures the prompt correctly, the provider patches don't.)

## Fix
Wire prompt-content capture into the shared provider-patch path (the same `_request_capture` helper that records params, or each provider integration), gated under `capture.prompts` exactly like the litellm path. Same for completion content (`capture.completions`).

anilmurty:
The full-request capture (#209) was incomplete for the provider monkey-patches: with `capture.prompts = true`, a `patch_anthropic` / `patch_openai` / `patch_gemini` / `patch_bedrock` LLM span got `request_params` populated but **not** `gen_ai.prompt.content` (the messages) or `gen_ai.completion.content`. Only `litellm.py` and `sdk/agent.py` set them. A span without the prompt isn't self-contained enough to replay — the whole point of #209.

## Summary
- Added shared `record_prompt_content(span, messages)` / `record_completion_content(span, text)` helpers to `_request_capture.py`, alongside `record_request_params`/`record_request_tools`, using the `GenAIAttributes.PROMPT_CONTENT` / `COMPLETION_CONTENT` semconv constants (Rule 10).
- **`record_prompt_content` serializes as `json.dumps(messages)`** — the exact shape litellm/`sdk/agent.py` already use, so a replay harness / backfill reading `gen_ai.prompt.content` gets **one** consistent serialization regardless of which patch produced the span.
- Wired both into all four provider patches, set **unconditionally** on the LLM-call span. `strip_captured_content()` at ingest already gates them per the `[capture] prompts` / `completions` toggles — same pattern litellm uses; no ingest-gate change, no default change (capture stays off by default).

## Per provider
- **anthropic**: `messages` kwarg + response `content` text blocks (joined; `tool_use` blocks skipped). Streaming path captures the prompt; completion there would need stream buffering (out of scope, noted below).
- **openai**: `messages` kwarg + `choices[0].message.content`.
- **gemini**: `contents` (first positional arg or kwarg) + `response.text`, on both the sync and async `generate_content` patches.
- **bedrock**: `messages`/`prompt`/`inputText` from the JSON `body`; completion from the **already-parsed** response body inside `_extract_bedrock_usage` (avoids a second read of the response stream).

## Also in this PR (the optional refactor)
Refactored `litellm.py::_record_prompt_content` + the non-streaming completion setter to call the shared helpers, so there's **one** prompt serialization across every capture path. Dropped the now-duplicate `_serialize_prompt` (its `messages`-positional / text-`prompt` fallbacks are preserved in `_record_prompt_content`). Behavior-preserving — all existing litellm tests pass.

## Tests
`tests/unit/test_provider_content_capture.py`:
- the shared helpers (incl. `json.dumps(messages)` shape, `None` no-ops, non-serialisable fallback);
- each provider's completion/contents extraction (SDK-independent);
- **anthropic + openai patches end-to-end** through a fake upstream — asserting the span carries `PROMPT_CONTENT == json.dumps(messages)` and `COMPLETION_CONTENT`;
- **gemini + bedrock** via the exact capture calls their patches make (their SDKs aren't installed in CI, so the install path early-returns);
- the **ingest gate**: both stripped when `[capture]` is off, kept when on, and gate independently (prompts on / completions off).

Full suite: **1173 passed**. `ruff` + `mypy` clean on all touched files.

## What's NOT in this PR
- **Completion content on streaming paths** (anthropic/openai stream wrappers): the wrappers don't aggregate the streamed assistant text, so capturing it would mean buffering the stream — a separate concern. The **prompt** is captured on streaming paths (it's known up front); only streaming *completion* is deferred.
- **Framework patches** (langchain / langgraph / crewai / autogen / llamaindex) — explicitly out of scope for this issue.

Closes #320

### DIFF
--- a/tests/unit/test_provider_content_capture.py
+++ b/tests/unit/test_provider_content_capture.py
@@ -1,0 +1,319 @@
+"""Prompt + completion content capture on the provider monkey-patches (#320).
+
+Before this, only litellm.py and sdk/agent.py set gen_ai.prompt.content /
+gen_ai.completion.content; the anthropic/openai/gemini/bedrock provider patches
+captured #209 request_params but NOT the messages or the completion text — so
+their spans weren't self-contained enough to replay.
+
+These tests verify, per provider:
+  - the LLM span carries PROMPT_CONTENT == json.dumps(messages) and the
+    COMPLETION_CONTENT, set UNCONDITIONALLY by the patch;
+  - with [capture] prompts/completions OFF, both are stripped at the single
+    ingest gate (strip_captured_content) — same path litellm uses.
+
+anthropic + openai are exercised through the real installed patch (with a fake
+upstream method); gemini + bedrock are exercised through the exact capture calls
+their patches make (their SDKs aren't installed in CI).
+"""
+from __future__ import annotations
+
+import json
+import types
+
+import pytest
+
+from tokenjam.core.config import CaptureConfig, TjConfig
+from tokenjam.core.db import InMemoryBackend
+from tokenjam.core.ingest import IngestPipeline
+from tokenjam.otel.semconv import GenAIAttributes
+from tokenjam.sdk.integrations._request_capture import (
+    extract_anthropic_completion,
+    extract_gemini_completion,
+    extract_openai_completion,
+    record_completion_content,
+    record_prompt_content,
+)
+from tests.factories import make_llm_span, make_session
+
+
+# ---------------------------------------------------------------------------
+# A recording fake span (mirrors the litellm test's mock tracer span).
+# ---------------------------------------------------------------------------
+
+class _FakeSpan:
+    """Minimal OTel-span stand-in recording set_attribute; the rest are no-ops
+    so a full patch closure (which also calls set_status/end) can run."""
+
+    def __init__(self) -> None:
+        self.attributes: dict = {}
+
+    def set_attribute(self, key, value) -> None:
+        self.attributes[key] = value
+
+    def set_status(self, *_a, **_k) -> None:
+        pass
+
+    def end(self, *_a, **_k) -> None:
+        pass
+
+    def is_recording(self) -> bool:
+        return True
+
+
+def _recording_tracer():
+    """A tracer whose start_span returns a _FakeSpan recording set_attribute."""
+    spans: list[_FakeSpan] = []
+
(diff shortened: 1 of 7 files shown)


anilmurty (commented):
Verified against #320, with attention on the one contract that matters most: the serialization. record_prompt_content writes exactly json.dumps(messages), the same shape litellm and sdk/agent already use, with a test asserting it — so a span from any capture path now reads back identically, which is what a replay harness / backfill depends on. Shared helpers in _request_capture.py, semconv constants (Rule 10), wired into all four provider patches, set unconditionally with strip_captured_content gating at ingest (no default change). The litellm consolidation to one serialization is a clean bonus, and the scope discipline is good — streaming completion and framework patches explicitly deferred. Comprehensive tests (helpers, per-provider extraction, anthropic/openai end-to-end, the ingest gate's independent toggles), 1173 passed, ruff + mypy clean, all four CI jobs green.

Ready to merge.

Issue #313, opened 2026-06-25
[Lens] Analytics leaderboard shows no total for the current breakdown (KPI Events ≠ chart subtotal when grouping by a partial dimension)


ashwmu:
## Summary

In the Analytics / Explore pivot, the **Leaderboard** chart shows a ranked list of
per-item bars with individual counts, but **no total/subtotal** for the breakdown itself.
When you group by a dimension that **not every span carries** (e.g. `Tool`, `Tool category`),
this is confusing: the **Events KPI tile** shows the *all-events* total, which does **not**
match the sum of the bars on screen, and nothing reconciles the two.

## Steps to reproduce

1. Analytics → Metric = **Events**, Chart = **Leaderboard**.
2. Set **By = Tool** (or **Tool category**).
3. Observe:
   - The bars sum to **295** (the tool spans) — but you have to add them up yourself; no total is shown.
   - The **Events KPI tile** reads **1,804** (all spans), which does not match the 295 in the chart.

Example (By = Tool): Bash 120 + Edit 112 + Read 44 + Write 8 + WebFetch 4 + AskUserQuestion 4 + ToolSearch 3 = **295**.
Example (By = Tool category): file 164 + shell 120 + other 7 + web 4 = **295**.
KPI Events tile in both cases: **1,804**.

## Why it's confusing (not a data bug)

This is *expected* behavior — grouping by a **partial** dimension drops spans that lack
that dimension (the 295 tool spans have a tool but no model; conversely the ~1,509 LLM
spans have a model but no tool). So the breakdown subtotal is legitimately smaller than the
all-events KPI. The problem is purely **presentation**: the screen shows the all-events
number (1,804) and the per-item bars, but never the **breakdown's own subtotal**, so the
two numbers look contradictory.

## Fix options (pick)

- Show a small header/footer on the leaderboard with **item count + subtotal**, e.g.
  `7 tools · 295 events` (and for the time-series charts, the windowed total).
- And/or when the breakdown subtotal differs from the all-events KPI, add a one-line note:
  `295 of 1,804 events have a tool` — making the gap explicit instead of surprising.

## Screenshots

_By Tool — bars sum to 295, KPI tile shows 1,804, no subtotal shown:_

<img width="1717" height="965" alt="Image" src="https://github.com/user-attachments/assets/eeb1e2bb-eb3e-4c10-803d-89d415121cc7" />

[...]

PR #322, opened 2026-06-26
feat(lens): show breakdown subtotal on the active KPI tile (#318)

### LINKED ISSUE #318: [Lens] Analytics KPI tile doesn't reflect the active breakdown (shows the all-events total even when grouping by a partial dimension)
## Summary

In the Analytics / Explore pivot, the KPI tiles (Spend / Tokens / Sessions / **Events**) always
show the **window-wide total** for each metric. That's correct as a headline, but when the chart
below is grouped by a **partial dimension** (one that only some spans carry — e.g. `Tool`,
`Tool category`, or `Model` when tool spans exist), the **chart's subtotal differs from the tile**,
which can read as a contradiction.

Example: with Metric=Events, By=Tool, the **Events tile shows 1,804** while the leaderboard sums to
**295** (only the tool spans). #313 added an inline note in the chart ("Only 295 of 1,804 events have
a tool…") to explain the gap, but the **tile itself** still shows 1,804 with no indication that the
current breakdown covers a subset.

## Why it matters

The active-metric tile is visually emphasized (it's the selected metric), so a user naturally reads
it as "the number for what I'm looking at." When the breakdown only covers part of the data, the tile
and the chart subtotal disagree with nothing on the tile to signal why.

## Enhancement options (pick / discuss)

- **Reflect the breakdown on the active tile:** when grouping by a partial dimension, show the
  breakdown subtotal on the selected metric's tile (e.g. `295` with a sublabel "in this breakdown"),
  or show both (`295 / 1,804`).
- **Or a subtle indicator:** keep the all-events total but add a small hint/tooltip on the tile when
  the active breakdown subtotal is smaller (e.g. an asterisk linking to the chart's gap note).
- **Or leave the tiles as window totals** and rely solely on the chart's #313 gap note (status quo) —
  documenting that the tiles are always window-wide is also a valid resolution.

Whatever the direction, the tile and the chart should not silently disagree for the active metric.

## Scope

UI-only (single-file SPA, `tokenjam/ui/index.html`, `AnalyticsView` KPI-tiles block). Follow-up to
#313 (which added the in-chart gap note) and related to #295. Not a data bug — the numbers are
correct; this is about reconciling the tile with the active breakdown.

## Environment
- tokenjam 0.5.2 (dev), macOS, Python 3.14
- Surfaced while verifying #295 + #313

## Screenshot

_Events by Tool, Leaderboard: the **Events KPI tile shows 1,804** (window-wide) while the leaderboard breakdown sums to **295** — the tile doesn't reflect the active partial-dimension breakdown:_

Image

ashwmu:
Fixes #318.

## Problem
In the Analytics pivot, when grouping by a **partial dimension** (one only some spans carry — e.g. `Tool`), the active count-metric KPI tile showed the **window-wide total** (e.g. Events = 1,804) while the chart below covered only a **subset** (295). Nothing on the tile reconciled the two, so the tile and the chart looked like they disagreed.

## Fix (UI-only, `tokenjam/ui/index.html`)
Split the **active** tile: the window total stays the headline, and below a divider it shows the **breakdown subtotal** with the dimension label:

```
1,804
Events
──────────
295  by tool
```

- **Count metrics only** (events/sessions — spend/tokens framing is more complex).
- **Only when there's an actual gap** (subtotal < window total) — so grouping by a universal dimension (Day/Agent) shows just the total, no sub-line.
- `KpiTile` gains an optional `sub` prop **defaulting to `null`**, so every other caller (incl. the Dashboard preview) is unchanged.

(Vertical/stacked layout chosen over a horizontal side-by-side after an A/B — the vertical keeps the total-vs-subset hierarchy clear.)

## Verified
- Events by Tool → tile shows `1,804 / Events / 295 by tool`; reconciles with the leaderboard's "Total: 295 · 7 tools" (#313).
- Events by Model → `1,509 by model`. Events by Day → no sub-line (nothing missing). Spend/Tokens → no sub-line.
- Other screens (Dashboard tiles) unchanged.

## Tests
- Static regression guard in `tests/unit/test_lens_ui_regression.py`.
- Full UI guard suite green (117 passed). Rebased on top of the merged #313.

### DIFF
--- a/tests/unit/test_lens_ui_regression.py
+++ b/tests/unit/test_lens_ui_regression.py
@@ -468,6 +468,21 @@ def test_analytics_leaderboard_has_inline_bars(html):
     assert ".lb-bar" in html
 
 
+# --- #318: active tile shows the breakdown subtotal for a partial dimension --- #
+def test_analytics_active_tile_shows_breakdown_subtotal(html):
+    # When grouping by a PARTIAL dimension (only some spans carry it, e.g. tool),
+    # the active count-metric tile shows the breakdown subtotal beneath the window
+    # total so the tile reconciles with the smaller chart subtotal (#318). Count
+    # metrics only; only when there's an actual gap.
+    assert "const breakdownTotal = (resp.rows || []).reduce" in html
+    assert "const activeSub = ((metric === 'events' || metric === 'sessions')" in html
+    assert "by ${breakdownDim}" in html
+    # KpiTile renders the optional sub-line; defaults null so other callers (e.g.
+    # the Dashboard preview) are unaffected.
+    assert "onSelect, sub = null }" in html
+    assert "kpi-sub-val" in html
+
+
 # --- #295: Stack gated to stacking charts; empty cross-tab gets a clear state - #
 def test_analytics_stack_gated_to_stacking_charts(html):
     # Stack only applies to the multi-series charts (bar/line). The Leaderboard
--- a/tokenjam/ui/index.html
+++ b/tokenjam/ui/index.html
@@ -381,6 +381,9 @@ tr.clickable:hover td.chevron { color: var(--accent); }
 .kpi-tile .delta-chip { font-family: 'Geist Mono', monospace; font-size: 11px; margin-top: 6px; display: inline-block; }
 .kpi-val { font-family: 'Geist Mono', monospace; font-size: 20px; font-weight: 600; }
 .kpi-lab { font-size: 11px; color: var(--text-dim); margin-top: 2px; }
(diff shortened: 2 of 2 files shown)


anilmurty (commented):
Thanks @ashwmu — clean close to the Analytics-clarity arc (#295 → #313 → #318). The active count-metric tile now shows the breakdown subtotal beneath the window total when there's a real gap, reconciling with the leaderboard's "Total: N · X tools" from #313. Nicely scoped: count metrics only, only when a gap exists, and KpiTile's new sub prop defaults null so every other caller (Dashboard preview included) is untouched. Regression test guards the derived values + the optional sub-line, module JS clean, all four CI green.

Ready to merge.

PR #323, opened 2026-06-26
Fix #298: Clarify drift baseline agent scope and live session rules

### LINKED ISSUE #298: [Lens/drift] 'Run at least 10 sessions' shown despite >10 sessions; backfill builds no baselines
The Drift screen shows "No baselines built yet. Run at least 10 sessions to establish a behavioral baseline" even when there are 13+ sessions.

Two problems:
1. **The threshold is per-AGENT** (10 sessions for a given agent), not total. With several agents, none may have hit 10 — but the message reads as a global count, so it's misleading.
2. **Backfilled sessions don't build baselines at all** — the drift baseline is built at *live* session end (DriftDetector), whereas backfill writes SessionRecords directly. So drift never works on backfilled data, which is most users' first-run experience.

## Fix
- Clarify the message: per-agent threshold, and that baselines build from live sessions (so backfill alone won't populate drift).
- Decide whether backfill should also build baselines (bigger change — can be a follow-up; the message clarification is the quick win).

tarun73:
Explain that behavioral drift baseline thresholds are computed per-agent (rather than globally) and build only from live sessions. Update the empty page warning on the Web UI and the CLI fallback message to explicitly state this.

## Summary
- **Web UI**: Updated the Drift View empty-state warning message in `tokenjam/ui/index.html` to clarify the per-agent live threshold.
- **CLI**: Updated the fallback warning printed by `tj drift` when no baselines exist in `tokenjam/cli/cmd_drift.py` to match the Web UI message.

## Tests / Verification
- Added `test_drift_no_baselines_warning_message` in `tests/integration/test_cli.py` to verify that `tj drift` prints the warning header when no baselines exist.
- Executed the full test suite successfully (**1,154 passed**).

## What's NOT in this PR (if scope was deliberately limited)
- **Phase 2 (Backfill baseline compilation)**: Auto-building baselines during `tj backfill` is deferred to a future follow-up issue/PR.

## Related issue
Closes #298

@anilmurty please review

## Checklist

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -314,6 +314,12 @@ def test_drift_no_baselines_exits_0(runner, db, config):
     assert data["agents"] == []
 
 
+def test_drift_no_baselines_warning_message(runner, db, config):
+    result = _invoke(runner, db, config, ["drift"])
+    assert result.exit_code == 0
+    assert "No drift baselines found." in result.output
+
+
 def test_drift_with_baseline_no_violations(runner, db, config):
     """Normal session (tokens within threshold) -> exit 0."""
     # Use a baseline without tool sequences so Jaccard doesn't fire
--- a/tokenjam/cli/cmd_drift.py
+++ b/tokenjam/cli/cmd_drift.py
@@ -45,7 +45,8 @@ def cmd_drift(ctx: click.Context, agent: str | None, output_json: bool) -> None:
         else:
             console.print(
                 "[dim]No drift baselines found. "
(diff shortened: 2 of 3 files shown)


anilmurty (commented):
Thanks @tarun73 — exactly the clarity fix for #298, and on both surfaces. The new message states the two things that confused the original (per-agent threshold, and live-only so backfilled sessions won't build a baseline), and CLI + Web UI now match. Good scope call deferring the backfill-baseline compilation to a follow-up — that was the bigger half of #298 and is rightly separate. Test asserts the warning, module JS clean, all four CI green.

Ready to merge.

PR #325, opened 2026-06-27
feat(ui): dynamic analytics csv export with loading state

### LINKED ISSUE #250: [Lens] Analytics CSV export: richer columns + ensure it matches the on-screen pivot
The Analytics "Export CSV" (`analyticsCsv(resp)`) is built client-side from the loaded response. Two issues plus an enhancement.

## 1. Export should match the on-screen pivot (possible bug)
Observed: the view was set to **Tokens × Model**, but the exported CSV header was **bucket_epoch, bucket_iso, provider, kind, sessions, tokens** — a different group_by/stack/metric than what was on screen. Either the export ran against a **stale `resp`** (dropdown changed without a re-fetch) or there's a mismatch. The export must always reflect the **current** metric / group_by / stack / window. Confirm the response is refreshed on control change so Export can't diverge from what the user sees.

## 2. Honor + document the window (works, but verify)
The CSV honors `since` (it's built from the windowed response — 90d returns the ~46 days that exist). Add a quick regression so this can't regress, and make the filename or a header row note the window (e.g. `tokenjam-analytics_2026-05-08_2026-06-23.csv`) so an exported file is self-describing.

## 3. Richer columns (the main ask)
Today's CSV emits only the **selected metric + tokens** (2 value columns). For real spreadsheet analysis, export the full per-bucket picture regardless of the on-screen chart metric:
- **cost** (API-plan) / implied-value + **% of cycle** (subscription) — respect plan-tier framing
- **input / output / cache-read / cache-write tokens** (separate columns)
- **sessions**, **events**
- the dimension keys (model / agent / provider / tool, per the active group_by + stack)

Consider a "full export" that dumps all metrics for the window (not just the charted one), so the CSV is a complete analysis artifact, not a 2-column slice of the current view.

## Constraints
Client-side, offline (Blob download, no server round-trip) as today. Respect plan-tier framing in any cost column. Static-grep regression for the new columns + on-screen-pivot fidelity; keep `test_ui_offline.py` green.

tarun73:
## Summary

Resolves #250
- Dynamically format CSV filenames with selected dates
- Export full 12 column metrics in CSV regardless of current view
- Disable export button + charts instead of blank are not at 50% opacity and set loading state while fetching 

## Related issue

Resolves #250: Analytics CSV Export Improvements
This PR fully resolves the three analytics export issues outlined in #250 by eliminating stale data downloads, fixing filename formatting, and vastly expanding the exported column payload.

**1. Stale Data Export Fix**
Previously, due to the asynchronous nature of data loading, users could click the "Export CSV" button before a newly requested time window or metric finished loading, resulting in a CSV of the old data slice.

**Fix:** We now immediately clear the response state (resp = null) and toggle a loading state whenever a filter (date, metric, group by) is changed.
UX Change: The **Export CSV button** is now explicitly tied to the loading state (disabled=${loading}). It immediately **greys out** and becomes **completely unclickable** while the network request is in flight and the **Charts block**  go to **50% Opacity** instead of going completely **blank**. (screenshot attached)

**2. Dynamic Filename Formatting**
Exported files were previously returning with an inconsistent, hard-to-read, space-separated format.

**Fix:** We implemented dynamic, ISO-compliant filename formatting.
When exporting a specific date range, the filename now accurately formats as: tokenjam-analytics_YYYY-MM-DD_YYYY-MM-DD.csv
When exporting without a date window (e.g., "All time"), it falls back cleanly to tokenjam-analytics.csv.
Regression Prevention: Added static-grep UI assertions to test_lens_ui_regression.py to guarantee this exact string format cannot regress.

**3. Full 12-Column Data Export**
Previously, the CSV export only dumped a tiny slice of data representing exactly what the current chart was rendering (e.g., only exporting the "Cost" column).

**Fix:** We upgraded the export functionality to act as a Full Export. The backend /analytics endpoint now safely supplies all metrics on every request.
The CSV generator now outputs 12 columns simultaneously regardless of the active UI metric: bucket_epoch, bucket_iso, group, stack, cost, cycle_share_pct, input_tokens, output_tokens, cache_read_tokens, cache_write_tokens, sessions, and events.
Framing Parity: The CSV generator natively respects the application's plan-tier rules. It zeroes out cost data for local modes and accurately populates the cycle_share_pct column for users on subscription modes.

Test screenshots:
Part 1:
1. Normal conditions (export button is functional and not greyed out)
Screenshot 2026-06-27 at 8 32 13 AM

2. During asynchronous loading (export button is disabled and greyed and if you hover over it shows loading) and the charts section becomes 50% opacity
greyed_out_csv_export_chart_50%_opacity_indicating_data_reload

Part 2:
CSV file names in the exact format
Screenshot 2026-06-27 at 9 06 08 AM

Part 3:
Persisted 12 columns if stack selected or not (for consistency)
example 1: Stack filter selected
Screenshot 2026-06-27 at 9 02 20 AM

example 2: No Stack filter but subscription showing cost column also None
subscription_empty_stack_empty_cost_12_columns

**Constraints Satisfied**
1. **Client-Side / Offline Export:** Retained the existing Blob-based download architecture (analyticsCsv & downloadCsv), ensuring the export happens entirely offline in the browser with zero additional server round-trips.
2. **Framing Parity:** The new CSV exporter strictly honors plan-tier constraints, accurately suppressing dollar values for local mode and injecting cycle_share_pct for subscription users.
4. **Test Fidelity:** Expanded test_lens_ui_regression.py to statically assert the presence of all 12 columns and dynamic filename logic, maintaining on-screen pivot parity while keeping test_ui_offline.py completely green (verified against the full 1,178 test suite).

## Checklist
- [X] CLAUDE.md updated (if architecture changed)
- [X] Test spans use `tests/factories.py` (not raw `NormalizedSpan`)

@anilmurty  can you please review

### DIFF
--- a/tests/unit/test_lens_ui_regression.py
+++ b/tests/unit/test_lens_ui_regression.py
@@ -429,9 +429,24 @@ def test_analytics_presets_and_csv_export(html):
     assert "function analyticsCsv" in html
     assert "function downloadCsv" in html
     assert "Export CSV" in html
+    assert "disabled=${loading}" in html
     # the leaderboard preset closes #214; spend-by-model line closes #216
     assert "'leaderboard'" in html
     assert "'spend-by-model'" in html
+
+    # Richer CSV columns assertions
+    assert "'cycle_share_pct'" in html
+    assert "'input_tokens'" in html
+    assert "'output_tokens'" in html
+    assert "'cache_read_tokens'" in html
+    assert "'cache_write_tokens'" in html
+    assert "'sessions'" in html
+    assert "'events'" in html
+    # Filename generation check
+    assert "getFilename" in html
+    assert "tokenjam-analytics.csv" in html
+    assert "tokenjam-analytics_${startStr}_${endStr}.csv" in html
+
 
 
 def test_analytics_url_is_source_of_truth(html):
--- a/tokenjam/api/routes/analytics.py
+++ b/tokenjam/api/routes/analytics.py
@@ -200,6 +200,13 @@ async def get_analytics(
         group_cols.append("g2")
     select_cols.append(metric_expr + " AS val")
     select_cols.append(_TOKENS_EXPR + " AS toks")
+    select_cols.append("COALESCE(SUM(cost_usd), 0.0) AS cost")
+    select_cols.append("COALESCE(SUM(input_tokens), 0) AS input_toks")
+    select_cols.append("COALESCE(SUM(output_tokens), 0) AS output_toks")
+    select_cols.append("COALESCE(SUM(cache_tokens), 0) AS cache_read_toks")
+    select_cols.append("COALESCE(SUM(cache_write_tokens), 0) AS cache_write_toks")
+    select_cols.append("COUNT(*) AS events")
+    select_cols.append("COUNT(DISTINCT session_id) AS sessions")
 
     sql = (
         "SELECT " + ", ".join(select_cols) + " FROM spans WHERE " + where
@@ -217,13 +224,33 @@ async def get_analytics(
             g2v = r[2] if r[2] is not None else "(none)"
             val = float(r[3] or 0.0)
             toks = int(r[4] or 0)
+            cost = float(r[5] or 0.0)
+            input_toks = int(r[6] or 0)
+            output_toks = int(r[7] or 0)
+            cache_read = int(r[8] or 0)
+            cache_write = int(r[9] or 0)
+            events = int(r[10] or 0)
+            sessions = int(r[11] or 0)
         else:
             g2v = None
             val = float(r[2] or 0.0)
             toks = int(r[3] or 0)
-        rows.append({"bucket": b, "group": str(g1v),
-                     "stack": (str(g2v) if g2v is not None else None),
-                     "value": val, "tokens": toks})
+            cost = float(r[4] or 0.0)
+            input_toks = int(r[5] or 0)
+            output_toks = int(r[6] or 0)
+            cache_read = int(r[7] or 0)
+            cache_write = int(r[8] or 0)
+            events = int(r[9] or 0)
+            sessions = int(r[10] or 0)
+        rows.append({
+            "bucket": b, "group": str(g1v),
+            "stack": (str(g2v) if g2v is not None else None),
+            "value": val, "tokens": toks,
+            "cost": cost, "input_tokens": input_toks,
+            "output_tokens": output_toks, "cache_read_tokens": cache_read,
+            "cache_write_tokens": cache_write, "events": events,
+            "sessions": sessions,
+        })
         totals[str(g1v)] = totals.get(str(g1v), 0.0) + val
         if g2v is not None:
             stacks[str(g2v)] = stacks.get(str(g2v), 0.0) + val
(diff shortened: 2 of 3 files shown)


anilmurty (commented):
Thanks @tarun73 — this nails the structure of #250: the stale-data guard (clearing `resp` + disabling Export while loading), the self-describing windowed filename (`tokenjam-analytics_<start>_<end>.csv`), and the richer per-bucket columns wired through the route, UI, and a regression test. CI's green and the module JS checks out.

One thing to fix before merge — a plan-tier framing leak:

In `analyticsCsv`, the **subscription** branch exports raw dollars: `costVal = String(rawCost)`. That contradicts the framing contract (Critical Rule 14 / `core/framing.py`): subscription users must never see raw dollar **spend** — and every other surface in this same file already honors that (`fmtSpend` renders subscription as "Implied value · N× plan value", `useTokens` switches subscription/local to token framing, `valueIsDollars` is false for subscription). #250 also asked specifically for "implied-value + % of cycle (subscription)", not a raw cost column. The CSV is currently the one surface that leaks the dollar figure.

Fix: for subscription, don't emit raw dollars in `cost` — blank it or emit tokens / implied-value consistent with the existing `useTokens` / `fmtSpend` helpers, and let `cycle_share_pct` carry the signal. Best to route the column through the existing framing helpers rather than re-deriving the `mode` logic inline in `analyticsCsv` (that inline re-derivation is where the leak crept in — the codebase rule is "consume the framing block, don't re-implement the suppression").

Minor (non-blocking):
- The PR body has a leftover placeholder line ("Resolves #250 (or whatever the issue number is...)") — please tidy it.
- `setSt({ ...resp: null })` on every filter change now blanks the chart on each control tweak (it used to keep showing prior data during the refetch). Fine for correctness, just noting the UX change.

Once the subscription cost column is suppressed/reframed, happy to merge.

tarun73:
Thanks for the detailed feedback. will work on the corrections you suggested.

anilmurty:
Thanks @tarun73 - please let me know when ready (update branch as well) and I'll review/ merge. thanks again for all the work!

tarun73:
Sure , i will ping here after all the updates, please ignore updates until then

tarun73:
Branch updated and rebased on latest main. Changes in the fix commit:

1. Routed CSV cost column through perItemUsesTokens() to suppress raw dollars for subscription/local (fixes the framing leak)
2. Stabilized CSV to strict 12-column schema (stack column always present)
3. Restored chart UX: dims to 50% opacity during refetch instead of blanking, shimmer only on first load
All tests green. Ready for review!

@anilmurty let me know if this works. 

Happy to contribute.

anilmurty (commented):
Thanks for the quick turnaround @tarun73 — the framing leak is properly fixed. `costVal = perItemUsesTokens(framing) ? '' : String(rawCost)` blanks the cost column for subscription/local (with `cycle_share_pct` carrying the subscription signal) and emits real dollars only for api/unknown, and it now routes through the existing `perItemUsesTokens` helper instead of re-deriving the mode logic — which is the right call. Stable 12-column schema, PR body cleaned, module JS checks out, all four CI jobs green.

One optional nice-to-have (not blocking): a one-line regression assert that the CSV cost goes through `perItemUsesTokens` — so this suppression can't silently regress, the way the #187 guards lock it for the table/trace surfaces. Happy to merge with or without it.

Resolves #250 cleanly. Ready to merge.

anilmurty (approved):
Framing leak fixed via the perItemUsesTokens helper; stable schema; CI green. Resolves #250.

PR #353, opened 2026-06-28
fix: batch of 7 bug-hunt fixes (optimize, mcp, db, analytics, sdk)

anshss:
Batch fix of 7 confirmed bugs surfaced by a codebase-wide bug-hunt sweep (each adversarially verified, then fixed TDD-style with a failing test first). All commits are atomic and independently reviewable; full suite is green (`1189 passed, 1 skipped`), `ruff` + `mypy` clean.

## Fixes

| Commit | Area | Bug |
|--------|------|-----|
| `ca81860` | `cli/cmd_optimize.py` | Subscription `tj optimize --json` leaked the dollar savings projection the human renderer suppresses — `monthly_savings_usd` zeroing was nested under the `local`-only branch, so subscription payloads kept the misleading dollar figure. Now zeroed for both subscription and local. |
| `39d5077` | `core/optimize/runner.py` | `summarize_window` total excluded cache-read tokens, so the "tokens this cycle" header disagreed with the downsize token-share line and the canonical window total on cache-heavy workloads. Now includes `cache_tokens`, matching the downsize denominator. |
| `2f31b9e` | `mcp/server.py` | `acknowledge_alert` opened a second read-write DuckDB connection while the read-only fallback connection was open, raising a raw `ConnectionException` — non-functional in the entire RO fallback mode. Now returns an actionable message instead of a raw connection error. |
| `709cb26` | `mcp/server.py`, new `api/routes/sessions.py` | `list_active_sessions` proxied to `/api/v1/status` (one row per agent) in serve mode, dropping concurrent sessions for the same agent and undercounting. Added a `GET /api/v1/sessions` route (one row per session) and pointed the serve path at it, restoring parity with the direct-DB path. |
| `53ee1af` | `core/db.py` | `repair_spans_stats` rebuilt the spans table with bare `CREATE TABLE … AS SELECT`, permanently dropping the `span_id` PRIMARY KEY, NOT NULL constraints, and all `idx_spans_*` indexes. Now single-sources the canonical DDL and recreates the full schema + indexes after the copy. |
| `7696e48` | `api/routes/analytics.py` | The `sessions` metric summed per-bucket `COUNT(DISTINCT session_id)` into `totals_by_group`, double-counting sessions that span multiple time buckets (so the grouped totals couldn't reconcile with the Sessions KPI tile). Now computes per-group totals with a window-wide distinct query, mirroring the KPI. |
| `257ca1a` | `sdk/integrations/litellm.py` | A streaming LiteLLM call that raised before the stream wrapper was built leaked the `_tj_litellm_active` suppression contextvar (and the OTel span), silently dropping all subsequent Anthropic/OpenAI telemetry on that context. Added a `handed_off` flag so the `finally` cleans up on the streaming-error path (sync + async) while still deferring to the wrapper on success. |

## Test plan
- [x] New failing-first test added per fix (RED confirmed, then GREEN)
- [x] `make test` → 1189 passed, 1 skipped
- [x] `make lint` (ruff) → clean
- [x] `make typecheck` (mypy) → no issues in 133 source files

### DIFF
--- a/tests/integration/test_analytics.py
+++ b/tests/integration/test_analytics.py
@@ -189,3 +189,30 @@ async def test_kpi_delta_non_null_when_prior_window_has_data():
     assert d["kpi_prev"]["spend"] > 0, "prior window should have seeded data"
     assert d["kpi_deltas"]["spend"] is not None
     assert isinstance(d["kpi_deltas"]["spend"], float)
+
+
+@pytest.mark.asyncio
+async def test_session_spanning_buckets_counted_once_in_totals_by_group():
+    """#45: a session active across multiple time buckets must count once in
+    `totals_by_group` for the sessions metric (COUNT(DISTINCT) is non-additive),
+    so the grouped totals reconcile with the window-wide Sessions KPI rather than
+    inflating to the per-bucket sum."""
+    db = InMemoryBackend()
+    cfg = TjConfig(version="1")
+    now = utcnow()
+    sid = "spanning"
+    db.upsert_session(make_session(session_id=sid, plan_tier="api", agent_id="cc"))
+    # One session with two LLM spans on consecutive days -> two daily buckets.
+    for d in (0, 1):
+        db.insert_span(make_llm_span(
+            session_id=sid, agent_id="cc", model="claude-opus-4-7",
+            provider="anthropic", input_tokens=1000, output_tokens=100,
+            cost_usd=0.05, start_time=now - timedelta(days=d),
+        ))
+    d = await _get(db, cfg, "metric=sessions&group_by=agent&since=30d")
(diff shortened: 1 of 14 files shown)


anilmurty:
Thanks @anshss — I went through all 7 fixes. Each is a genuine bug with a failing-first test, and the optimize one actually tightens the subscription honesty framing (zeroing the dollar projection the renderer already suppresses — good catch). No correctness or convention issues, CI green.

One note for next time, not blocking this: the DB DDL rebuild (`repair_spans_stats`) and the new `/api/v1/sessions` route each carry more blast radius than the other five one-liners — those two are easier to review and safer to revert as their own PRs. The other five are textbook batch-sized.

Issue #329, opened 2026-06-28
[enhancement] Summary — initial CLI + MCP surface (scan → mechanism → apply → delivery)


dankellysh:
Part of #327 (the overall Summary epic, which stays open through UI + follow-ons).
This issue tracks the initial reviewable surface — the headless core plus the `tj summarize` CLI and MCP tools.  
It is **closed by the final PR in the delivery series**.

Quick note, I'm going to partially deliver some files in each PR, culminating in the full deployment.  
Some files will be edited multiple times over the series.
Each PR is set to be a stopping point where discussion / change of surface can happen.

## Scope (for this issue)
The structure-preserving prompt-summarization path, end to end, minus UI:

- **Discovery** — `tj summarize list`: find instruction-style prompt files worth summarizing
  (CLAUDE.md / AGENTS.md / persona / skill files; opens to all `*.md` on request) and show the
  estimated per-call token saving. Read-only, advisory.
- **Mechanism** — `prep` / `check`: detect structure (fenced/inline code, tags, templates,
  tables), wrap each block behind an id'd marker, summarize **only the prose**, restore by id.
  A check gate **refuses to stage any rewrite that dropped or altered a structure marker**
  (structure is a hard guarantee; behavior is validated by tests). No-scratch: the file on disk
  is the source of truth, re-read and hash-guarded.
- **Apply** — `apply` / `undo`: stage → review → write, **dry-run by default**, explicit `--go`,
  backup + restore, refuse-on-drift, **no `--force`**. Skips files already summarized + unchanged.
- **Delivery** — who produces the rewrite: manual (paste), `--via claude-p` (your local Claude
  Code, headless), `--via api` (Anthropic with your own `TJ_ANTHROPIC_API_KEY`; required
  `[summarize] api_model`, no default), or the MCP path (the agent rewrites in-session). API
  delivery shows a "pays for itself in ~N uses" line from the real charge ÷ the estimated saving.

## Ethos
- **Local-first** — no telemetry needed; pure stored-prompt + filesystem.
- **Estimations** — savings are **estimated** (from the stored prompt's size reduction,
  amortized over reuse); structure is a guarantee, behavior is *validated* not asserted;
  compression is calibrated (~50%+ on prose-heavy prompts, less on structure-heavy).
- **Ask-first** — default dry-run, explicit apply, no escape-hatch flag, refuse-on-drift; the
  user can reject the candidate at discovery or the output before apply.
  
[...]

PR #354, opened 2026-06-28
Add summarize scan surface: tj summarize list + MCP tool

dankellysh:
## Summary

PR1 of the stacked Summarize series — an advisory, read-only scan. `tj summarize list`
and the `list_summarize_candidates` MCP tool find instruction-style prompt files worth
summarizing and report an estimated per-call token saving per file. Nothing is written;
user selection and the prep/check/apply flow come in later PRs.

- **Catalog-default, opt-in reach:** bare `list` = known prompt locations + globals; any
  scope-widening input (PATH / `--repo` / `--recursive` / `--ext`) opens the net to `*.md`
  (+ any `--ext` extensions).
- **Pure-filesystem repo detection** (nearest ancestor `.git`, no git subprocess) with
  boundary safety (never above home / fs-root / a bare top-level). `--recursive` walks only a
  validated root or an explicit PATH, with skip-dirs + a stat pre-filter + a hard file cap.
- **Sectioned, honest output:** scanned location first, always-on catalog globals after a
  divider; prompts rank before other files within each section; reports root + depth +
  globals checked. Framing stays "candidate — review" / "estimated per-call reduction,"
  never "saves you."
- **Config-only:** summarize never opens the DB.

Includes a fix found in review: a missing explicit PATH was silent under `--recursive` (the
recursive branch ran before the not-found check, set the walk root to the non-existent target,
walked it to an empty result, emitted no note). The existence check is now hoisted above the
scope dispatch, so a typo'd PATH reports "PATH not found:" in every mode (plain / `--repo` /
`--recursive`); covered by a regression test.

## Related issue

Part of #329 (Summarize feature series) — first of the stack, so no `Closes`.

## What's NOT in this PR (stacked series)

- `prep`/`check` (wrap → summarize → restore → verify) — next (mechanism) PR
- `apply`/`undo`/backup + skip-already-done — apply PR
- delivery (`--via claude-p` / `--via api`) — delivery PRs
- pre-existing `test_mcp_server.py` import debt (E402/F811) — separate lint chore

## Checklist

- [x] Tests pass (`pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/`) — 1208 passed, 2 skipped
- [x] CLAUDE.md updated (if architecture changed) — N/A, additive feature, no arch change
- [x] Test spans use tests/factories.py (not raw NormalizedSpan) — N/A, scan tests are filesystem/catalog, no spans
- [ ] Requested @anilmurty as reviewer — set via `--reviewer anilmurty` on PR creation

### DIFF
--- a/tests/integration/test_summarize_cli.py
+++ b/tests/integration/test_summarize_cli.py
@@ -1,0 +1,99 @@
+"""Integration tests for `tj summarize list` via CliRunner."""
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+from click.testing import CliRunner
+
+from tokenjam.cli.main import cli
+from tokenjam.core.config import TjConfig
+from tokenjam.core.summarize import candidates
+from tokenjam.core.summarize.catalog import Catalog
+
+
+@pytest.fixture
+def runner():
+    return CliRunner()
+
+
+@pytest.fixture(autouse=True)
+def iso_catalog(monkeypatch):
+    """No real globals — controlled catalog so output is deterministic."""
+    fake = Catalog(project_files=frozenset({"CLAUDE.md", "AGENTS.md"}),
+                   project_globs=(), global_paths=(), forbidden_roots=())
+    monkeypatch.setattr(candidates, "load_catalog", lambda: fake)
+
+
+def _invoke(runner, args, open_db_side_effect=None):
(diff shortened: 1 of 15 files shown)


anilmurty (approved):
Thanks @auspexlabs — merging. Cleanly scoped as the read-only scan surface, kept in its own `core/summarize/` package rather than folded into the analyzer registry. The honesty framing is exactly right: standing only behind the token estimate and refusing to fabricate a dollar figure is the call I'd want here. Good base for the mechanism in #355.

PR #355, opened 2026-06-28
Add summarize mechanism: prep/check — CLI (manual) + MCP (in-session)

dankellysh:
## Summary

Second of the stacked Summarize series (#329), built on the scan surface. The
structure-aware prep/check mechanism, reachable two ways — both rewrite only the
PROSE and must prove the structure survived before anything is staged:

- CLI (manual / copy) — the universal compatibility path: works with ANY model +
  workflow, no integration required. `tj summarize prep <file>` wraps the prompt's
  structure behind id'd <tj-keep> markers and emits the wrapped prompt + rewrite
  rules + a source hash. You summarize the prose in whatever model you like, then
  `tj summarize check <file> --summary <out> --prepped-hash <hash>` restores,
  verifies, and stages.
- MCP (in-session): the `summarize_prep` / `summarize_check` tools run the same path
  with the agent rewriting in-conversation; the staged result records
  `produced_by = in-session`.

Both paths (-> all, once a later PR lands) converge on `check`. The system re-reads
the file (refusing if it changed since prep), re-derives the wrap, restores every
block by id, and stages the result ONLY if structure survives. This is a hard gate
to prevent dropped or altered markers from persisting into the final result. If this
happens, the operation is refused and nothing is staged.

Additional: must-keep word movement is recorded on the staged result for later
metrics only.

There is no scratch state or database. The file on disk is the source of truth.
Prep persists nothing, and check re-derives from the live file + the hash. The manual
path is the always-works baseline; TJ-drives-the-model delivery (--via claude-p /
--via api) is the convenience layer on top of it, landing in a later PR along with
apply/undo.

## Related issue

Part of #329 (Summarize feature series) — second of the stack, so no `Closes`.

## What's NOT in this PR (stacked series)

- `apply`/`undo`/backup (commit staged results to disk) — next (apply) PR
- TJ-drives-the-model delivery (`--via claude-p` / `--via api`) — delivery PRs
- pre-existing `test_mcp_server.py` import debt (E402/F811) — separate lint chore

## Checklist

- [x] Tests pass (`pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/`) — 1254 passed, 2 skipped
- [x] CLAUDE.md updated — `summarize` added to `no_db_commands`; `tj summarize` (CLI Commands), `core/summarize/` (Key Modules), and the MCP tools documented for the current surface
- [ ] Requested @anilmurty as reviewer — set via `--reviewer anilmurty` on PR creation


anilmurty:
Thanks @auspexlabs — the mechanism itself reads great: non-destructive (staging only, never touches the source file), hash-guarded, symlinks refused, and well-tested across the failure modes. The honesty framing holds up too — "structure preserved" is mechanically true via verbatim restore-by-id, and it never claims meaning or quality is preserved.

Logistics first: this PR was based on #354's branch, so when I merged #354 (and deleted that branch) GitHub auto-closed this one. I've reopened it and retargeted the base to `main`.

Two things before it can land:

1. **Rebase onto `main`.** #354 was squash-merged, so its scan files now live in `main` as one commit — which conflicts with the copies still in this branch (the PR currently shows as conflicting, ~2400 lines). Rebasing `feat/summarize-mechanism` onto `main` drops the duplicate scan commits, resolves the conflict, leaves just the mechanism diff, and triggers CI (which hasn't run on this branch yet).

2. **Token label.** The CLI verdict labels savings `~X tok_out/call`, but summarizing a reused prompt saves *input* tokens, not output — `prompt tok/call` would be accurate. (The MCP docstring already gets this right.)

Once it's rebased and green I'll take another pass.

dankellysh:
Sounds good let me fix the re-base issue here and change the way the next ones were going to release so this isn't an issue.

anilmurty (approved):
Thanks @auspexlabs — merging. I rebased this onto main (dropped the scan commits now in via #354) and relabeled the per-call figure as `prompt tok/call`. The prep/check mechanism is non-destructive (staging only), hash-guarded, and well-tested — clean base for the apply layer in #356.

PR #356, opened 2026-06-28
Add summarize apply/undo + backup (#329)

dankellysh:
The write half of the summarize feature: `tj summarize apply` writes a staged rewrite back to its file, and `tj summarize undo` restores it.
PR1 was read-only and PR2 was prep/check + staging. Neither wrote anything. This is the first summarize surface that touches a user's files.

## Summary
- `tj summarize apply [PATH]` — applies staged results (take-all, or one PATH). **Default dry-run** (prints the unified diff); `--go` writes. `--dry-run`/`--go` are mutually exclusive.
- On `--go`, each file must clear a chain of guards before it's touched: structure check passed, file exists, **not a symlink**, **owned by the current user**, **content hash still matches** what `check` saw. Any failure is skipped and reported, never written. There is no `--force` where an agent could override dangerously.
- Writes are atomic (temp in the same dir → `os.replace`) and preserve the file's mode. The original is backed up first (gzip + a JSON meta sidecar under `~/.tj/summary/backups/`).
- `tj summarize undo PATH` — restores a file from its backup.
- The scan (`tj summarize list`) now skips a file we already summarized and haven't changed since, and treats a corrupt/partial backup sidecar as "not summarized" rather than crashing.
- Exposed through both front doors: the `apply`/`undo` CLI commands and the `summarize_apply`/`summarize_undo` MCP tools.

## Tests / Verification
- New `tests/unit/test_summarize_apply.py` covers the rails: dry-run-writes-nothing, go-writes-backs-up-clears, skip-drifted, skip-unowned, mode preservation, take-all-partial, undo round-trip / dry-run / refuse-on-drift / refuse-symlink / no-backup, and corrupt-meta tolerance.
- CLI + MCP round-trips and the `--dry-run`/`--go` mutual-exclusion added to the existing suites.
- Rebased onto the current `main`/mechanism; corrected the savings label to `prompt tok/call` (matches the #355 fix).
- Full CI (`unit/synthetic/agents/integration`): **1291 passed, 2 skipped**. `ruff check tokenjam/` and `mypy tokenjam/` clean.

## What's NOT in this PR
- **Delivery** — `--via claude-p` / `--via api` (have a model produce the summary) and the "pays for itself" amortization land in the next PRs of the stack.
- **A `[summarize]` config section** — apply needs no config knobs yet; the backup/staging anchor is `~/.tj/summary/`.

## Checklist
- [x] One concern (apply/undo/backup + skip-already-done)
- [x] `ruff check tokenjam/` + `mypy tokenjam/` clean
- [x] Full suite green standalone on the branch (1291 passed / 2 skipped)

Part of #329.


anilmurty (approved):
Thanks @auspexlabs — merging. Rebased onto main after #355. The write rails are exactly right: dry-run by default with an explicit `--go`, the guard chain (structure → exists → not-symlink → owned-by-user → hash-match) skips and reports rather than writing, no `--force`, atomic write + gzip backup, and `undo` refuses on drift/symlink/no-backup. The most sensitive surface in the stack, handled carefully.

PR #360, opened 2026-06-30
chore: add CI-gated GHCR Docker image publish workflow

vigneshvinfra:
## Summary

Publishes the tokenjam container image to GHCR on every successful `main` build. A new `docker-publish` workflow triggers via `workflow_run` only after CI succeeds (so a broken commit never ships an image), with build/login/tag logic factored into a reusable composite action at `.github/actions/docker-publish`.

## Related issue
N/A

## Checklist

- [x] Tests pass — N/A, no Python/test code touched (CI-config only)
- [x] Lint clean — N/A, no `tokenjam/` files changed
- [x] Type check clean — N/A, no `tokenjam/` files changed
- [N/A] CLAUDE.md updated (if architecture changed)
- [N/A ] Test spans use `tests/factories.py` (not raw `NormalizedSpan`)


anilmurty:
Thanks @vigneshvinfra, this is clean work. The `workflow_run`-after-CI pattern is the right safe approach, and I verified it wires up correctly: CI's workflow name matches, it gates on a successful conclusion, and it pins the build to the exact commit that passed.

One change before we merge. I'd like the image to publish on **releases**, not on every main commit. Our PyPI and npm packages are release-gated with semver versions, and I want the container to line up with them so `ghcr.io/metabuilder-labs/tokenjam:0.5.2` matches `pip install tokenjam==0.5.2`. Publishing `latest` on every main push lets the image run ahead of the released version, which gets confusing fast.

Concretely:
- Trigger on `release: [published]` instead of `workflow_run` (keep `workflow_dispatch` for manual runs if you like).
- Tags: add `type=semver,pattern={{version}}` and `type=semver,pattern={{major}}.{{minor}}`, and keep `latest` on release. Drop the per-commit short-SHA `latest`.

That mirrors how `publish-pypi.yml` and `publish-npm.yml` fire, so all three channels ship together on a release.

Two optional nice-to-haves, not blockers: pin the actions to commit SHAs for supply-chain hardening, and note the GHCR package is private on first publish so we'll flip it public.

Once it's release-gated, this is good to merge. Thanks again for picking this up.

anilmurty (changes requested):
noted above and make sure CI checks pass

anilmurty (approved):
Approving per review (critical rules clean, CI green). Merging for a clean pre-release board.

PR #361, opened 2026-06-30
feat(cli): add read-only tj pricing list command

### LINKED ISSUE #282: [enhancement] Add read-only tj pricing list to inspect the resolved pricing table
Users can add `[pricing]` overrides (documented in `docs/configuration.md`), but there's no way to inspect what rate actually **resolved** for a given model — an override vs the packaged table vs the `$0.50`/`$2.00` default. A read-only `tj pricing list` would surface the resolved table.

`tokenjam/core/pricing.py` already exposes `load_pricing_table()`, `get_rates()`, and the override-merge logic — this is a thin CLI over existing functions.

## Acceptance criteria
- New `tj pricing list` subcommand (new `tokenjam/cli/cmd_pricing.py`, registered in `main.py`, added to `no_db_commands` since it reads only config/pricing).
- Prints a table of `(provider, model, input, output, cache_read, cache_write, source)` derived from the loaded table + override layers.
- Supports `--json` and a `--model <name>` filter.
- A unit test covering the resolved-rate output.

Note: the `set` half of a future `tj pricing set|list` is **out of scope** here — read-only `list` only. No new domain logic; reuses existing pricing loaders.

Axiya3749:
## Summary

read-only tj pricing list, with --json and --model, surfaces the resolved pricing table. Includes a source column (override/packaged/default) derived by comparing the packaged table against the override layers, as described in the issue.

cc @anilmurty

## Related issue

Closes #282

## Checklist

- [ x ] Tests pass (`pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/`)
- [ x ] Lint clean (`ruff check tokenjam/`)
- [ x ] Type check clean (`mypy tokenjam/`)
- [ x ] Requested `@anilmurty` as reviewer (or @-mentioned him above)

### DIFF
--- a/tests/unit/test_cmd_pricing.py
+++ b/tests/unit/test_cmd_pricing.py
@@ -52,12 +52,19 @@ _FAKE_TABLE = {
     },
 }
 
+# claude-sonnet-4 comes from an override layer; gpt-4o from the packaged table.
+_FAKE_SOURCES = {
+    ("anthropic", "claude-sonnet-4"): "override",
+    ("openai", "gpt-4o"): "packaged",
+}
+
 
 def test_pricing_list_json_outputs_resolved_rates(runner):
     """--json emits one object per model with the resolved rate fields."""
     with patch("tokenjam.cli.cmd_pricing.load_pricing_table",
                return_value=_FAKE_TABLE), \
-         patch("tokenjam.cli.cmd_pricing._source_map", return_value={}):
+         patch("tokenjam.cli.cmd_pricing.load_pricing_sources",
+               return_value=_FAKE_SOURCES):
         result = _invoke(runner, ["pricing", "list", "--json"])
 
     assert result.exit_code == 0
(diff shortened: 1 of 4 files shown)


anilmurty:
Thanks @Axiya3749, this is a clean first PR — it hits all the acceptance criteria on #282 (read-only `list`, `--json`, `--model`, the source column, a unit test, and `set` correctly left out), and the TOML handling and `no_db_commands` wiring are right. CI is green across the board.

One change before merge, plus a small follow-on:

**1. Keep the source derivation in core, not the CLI.** `_source_map()` imports `_override_raw_sources` and `_split_pricing_raw` (both private) and re-implements the packaged→override precedence — your docstring notes it mirrors `pricing._build_pricing()`. That duplication will drift if the precedence changes. Could you add a public helper to `tokenjam/core/pricing.py` next to `_build_pricing()`, e.g. `load_pricing_sources() -> dict[tuple[str, str], str]` returning each `(provider, model)` → `"override"`/`"packaged"`, and have the CLI call it? Then `cmd_pricing.py` only touches public API and there's one source of truth for precedence.

**2. The "default" source can't occur for a listed row.** `load_pricing_table()` only holds packaged + override models; the `$0.50`/`$2.00` default applies in `get_rates` to models *not* in the table, which `list` never enumerates. So a listed row is always packaged or override. Once source comes from the core helper, drop the "default" branch (or keep it as a defensive fallback, not an asserted one) and adjust that test to a real packaged/override row.

Everything else looks good. Thanks for picking this one up!

anilmurty (changes requested):
noted in previous comment

Axiya3749:
Thanks for the detailed review! I've addressed both points:

Moved the source logic into a public load_pricing_sources() in core/pricing.py (next to _build_pricing()).
Dropped the "default" branch. The fallback is now "packaged" (defensive, not asserted), and I updated the test to check a real packaged/override row.

Both pushed. Let me know if you'd like any other changes!

anilmurty (approved):

PR #363, opened 2026-07-01
Feat/312 async hooks

### LINKED ISSUE #312: [enhancement] Make heavy post-ingest hooks asynchronous so they don't block span-ack
All post-ingest hooks (`CostEngine`, `AlertEngine`, `SchemaValidator`) run **synchronously** on the ingest path after each span is written. Today that's fine because the hooks are fast, but it means any heavier hook would add directly to span-acknowledgment latency, which constrains what we can run inline.

## Proposal
- Introduce an optional async path (asyncio task queue, or an APScheduler periodic drain — APScheduler is already a dependency) for hooks that can tolerate a small delay.
- Keep `CostEngine` **synchronous** (cost updates are needed immediately for budget checks).
- Allow `AlertEngine` and `SchemaValidator` to run off the critical path.
- Preserve the existing error-tolerance contract: hook failures stay logged-and-swallowed, never propagated to the ingest caller.

Enhancement / architecture; unlocks heavier analysis on ingest without slowing acknowledgment. Lower priority than the hygiene items.

tarun73:
## Summary
This PR introduces an optional asynchronous path for post-ingest hooks (`AlertEngine` and `SchemaValidator`) to run off the critical span ingestion path, avoiding span-acknowledgment latency overhead. `CostEngine` remains
  synchronous for immediate budget checks. 

NOTE: Additionally, this PR fixes a timezone/billing-cycle boundary test failure in `test_budget_cap_policy.py`. (this is not directly related but this test fails since i am running it at late 30th June and the UTC date is in the next month (July 1st). If you think this should be handled separately we can do that too)

### Changes:
- **AlertsConfig**: Added `async_hooks` boolean configuration flag (defaults to `false`).
- **IngestPipeline**: Added background thread-safe queue (`_hook_queue`) and worker thread (`TjHookWorker`) to run Alert and Schema Validator hooks asynchronously.
- **Graceful Shutdown**: Added flush and close hooks to OTel `TjSpanExporter` and FastAPI daemon serve lifespan to ensure no queued alerts are lost on exit.
- **Budget Cap Policy Tests**: Injected a deterministic clock (`now_fn`) and mid-month baseline datetime to resolve boundary failures on the 1st of the month UTC.

## Related issue

Closes #312

## Checklist

### Other Testing evidence:
1. Ran couple of more testing to check real life examples and it works as expected. 

@anilmurty please review.

Feedback appreciated


anilmurty:
Thanks @tarun73 — this is a well-scoped take on #312 (CostEngine stays sync, error-tolerance preserved, default-off), and I appreciate you flagging the timezone test fix explicitly. A few things to resolve before it goes into the ingest path, since it's threading on a hot path:

**1. Shutdown can drop queued alerts.** The worker loop exits on `_shutdown_event` at the top of each iteration, so once `close()` sets the event it stops without draining the queue — and `cmd_serve.py`'s lifespan calls `close()` without `flush()` first. Net: queued alerts are lost on shutdown, which is the opposite of the goal. Could you have the worker drain to empty on shutdown, and make every shutdown path `flush()` then `close()`? Also guard `flush()` so it can't hang if the worker already exited (`queue.join()` with no consumer blocks forever).

**2. Concurrent DB writes.** In async mode the worker thread writes alerts while the main thread writes spans/cost. Our per-thread-cursor setup (#124) is validated for concurrent *reads*; concurrent *writes* from two threads can hit DuckDB write-write conflicts under load, and since hooks swallow errors that'd silently drop alerts. Can you add a test that runs async hooks under concurrent ingest and confirm no conflicts?

**3. Bound the queue.** `queue.Queue()` has no `maxsize`; under a slow hook + high volume (the case this targets) memory grows unbounded. Please bound it with an explicit overflow policy, and log if anything is ever dropped.

**4. CLAUDE.md.** The ingest description still says hooks run synchronously — please document the async path + the `async_hooks` flag (the checklist box is ticked but the file isn't in the diff).

Minor: the `hasattr(config, "alerts")` guards are unnecessary now; and since the budget-cap fix adds a `now_fn` to production code, note it under "What's also in this PR" or split it.

Good direction — just want the shutdown-drain, write-safety, and the bound solid before this lands on the write path.

anilmurty (changes requested):
as noted above

anilmurty:
Pushed a follow-up commit (`f0cf652`) addressing the four review blockers, plus merged `main` (the branch was behind). Original authorship preserved via `Co-Authored-By`.

## Blockers addressed
1. **Shutdown dropped queued alerts.** The worker no longer bails at the top of the loop on `_shutdown_event`; on shutdown it drains the queue to empty (sentinel + only-exit-on-empty-once-shutdown). `close()` now flushes first, and the `tj serve` lifespan does `flush()` then `close()`. `flush()` is guarded so it returns immediately when no worker is alive rather than blocking forever on `queue.join()`.
2. **Concurrent DB writes.** In async mode the worker writes alerts while the ingest thread writes spans/cost; DuckDB's optimistic concurrency can abort one with a write-write conflict that the error-swallowing hook path would silently drop. `DuckDBBackend` now serializes all mutating methods through a re-entrant `write_lock`; the per-thread-cursor read path (#124) stays lock-free. Added `tests/synthetic/test_async_hooks_concurrency.py` running async hooks under concurrent (and multi-producer) ingest, asserting no lost writes / no conflict errors.
3. **Unbounded queue.** Now bounded (`HOOK_QUEUE_MAXSIZE = 10_000`) with an explicit **drop-oldest** overflow policy; drops are counted and logged at WARNING — never silent.
4. **CLAUDE.md.** Updated to document the async path accurately (was still described as synchronous): the `async_hooks` flag, cost-stays-sync, the write lock, and the lossless-shutdown contract.

Also removed the now-unnecessary `hasattr(config, "alerts")` guards. The budget-cap test's clock fix was absorbed by the `main` merge (`proxy/engine.py` already carries `now_fn` in production — no new production code for the test).

## Verification
- `ruff check tokenjam/` clean; `mypy tokenjam/core/ingest.py tokenjam/core/db.py` clean.
- `pytest tests/unit/ tests/synthetic/ tests/integration/` — **1817 passed**, incl. the new concurrency/shutdown/overflow tests.
- Default-OFF confirmed by `test_default_off_is_synchronous`: a config without `async_hooks` runs hooks inline with no worker thread.

anilmurty (dismissed):
Approving. The async-hooks implementation is correct on every dimension that matters for a core storage change: default path is a true no-op (no thread/queue created), no lock-ordering deadlock (write_lock released before the non-blocking enqueue), provably lossless shutdown, bounded drop-oldest queue with logging. The write_lock covers every DuckDBBackend writer and the new same-row-UPDATE contention test is verified load-bearing (fails with the lock swapped for a no-op). Thanks @tarun73 for the original work. Merging for a clean board.

anilmurty (approved):
Re-approving on the deflaked commit — the bounded-queue drop test is now deterministic (worker held via an 'entered' Event during the flood, drops asserted exactly). CI green, 1901 tests pass. Merging — clean board.

tarun73:
@anilmurty thanks for completing this.

PR #306, opened 2026-06-25
feat: Claude Code session-status dashboard — lifecycle fix, session detail/timeline/map, run grouping, quickstart

anshss:
Turns the Status page from a flat agent list into a real **session-status dashboard** for Claude Code — fixing the root bug where live sessions showed as "completed", then building the drill-in surfaces (detail / timeline / map), cross-session run grouping, and a zero-install first run. Rebased onto and merged with current `main` (v0.5.2).

## Summary
- **Root fix:** Claude Code/Codex export OTLP *logs* (not traces); each `user_prompt` becomes a zero-duration `invoke_agent` span that `ingest` mistook for a session *completion* — so every live session force-completed on its first prompt. Completion is now gated on real duration (`end_time > start_time`); non-end spans re-activate mistakenly-completed sessions.
- **Session lifecycle:** active / idle / stale / closed tiers (`SessionRecord.status_at`), close-signal endpoint + `tj session-end`, archive view, per-terminal tiles, display labels (`service.instance.id` + manual overrides), and a `claude` shell wrapper for per-terminal naming.
- **Project grouping:** dashboard tiles roll up by OTel `service.namespace`, with a server-side `[agents.<id>].project` fallback so already-running sessions group without a restart.
- **Session drill-in:** a per-session detail view with model-mix + context-growth, a deterministic **Timeline** (transcript play-by-play incl. recursive subagent logs, no LLM), and a graphical **Map** of what the agent did per ask.
- **Cross-session Run grouping:** a fan-out harness stamps `tokenjam.run_id` / `parent_session_id`; tj groups spawned workers into one Run (`/api/v1/runs`), plus a `setup_harness` MCP tool + integration doc.
- **Sub-agent cost attribution:** `sub_agent_id` on spans, per-subagent cost breakdown, and a subagent right-sizing analyzer.
- **Zero-install first run + Claude-Code wedge:** `tj quickstart` (`npx tj quickstart` / `uvx`), `tj context` cost diagnostic, `tj tokenmaxx` quota card, retroactive Opus quota audit, branded home screen on bare `tj`.
- **Cost/pricing correctness:** capture cache-creation tokens separately from reads, current Anthropic/Haiku-4.5 rates, framing block on `/api/v1/sessions/{id}`, bulk-insert backfill (perf).

## Merge with `main` (v0.5.2)
[...]

PR #367, opened 2026-07-02
fix: Claude Code session status + lifecycle (1/5 from #306)

anshss:
Carved from #306 as **PR 1 of 5** — a focused split of the session-status dashboard branch. This one delivers the root bug fix + the honest session lifecycle model it enables.

## Root fix
Claude Code / Codex export OTLP **logs**, not traces — so each `user_prompt` turn produces a zero-duration `invoke_agent` span (`end_time == start_time`). `ingest` mistook that for a session **completion**, so every live session force-completed on its first prompt (dashboard showed active work as "completed" with 0 duration, and it tripped drift/alert false-positives). Completion is now gated on **real duration** (`end_time > start_time`); a non-end span **re-activates** a mistakenly-completed session.

## Lifecycle model (built on the fix)
- **Tiers** active / idle / stale / closed via `SessionRecord.status_at()` (5 min stale, 4h idle default), with a transcript-mtime rescue for live-but-span-quiet sessions.
- **Explicit close** — `POST /api/v1/sessions/close` + `tj session-end` (preserves `ended_at` / "Last seen"; only stamps status).
- **Project grouping** — OTel `service.namespace` + a server-side `agent→project` fallback (groups already-running sessions without a restart).
- **Display labels** — `service.instance.id` + manual overrides (right-click rename) + `tj otel-resource-attrs`.
- **Split-zones Status** — coding sessions (cards + archived) vs SDK services (per-minute cost/err sparklines); 0-signal "zombie" sessions filtered from the archive.

## Migrations
`8` service_namespace · `9` service_instance_id · `10` ended_at repair · `11` session_labels (renumbered contiguous from main's current max of 7).

## Notes for the series
- **Decoupled from later PRs:** the `capture_session_method` snapshot on close, the `session_token_cost_rollup` (#18) used by the zombie filter, and the `core.transcript` mtime-rescue are all deferred — PR1's close endpoint just closes; the zombie filter uses stored token/cost. #18 lands in the cost PR; method-capture in the Approach/Map/Timeline PR.
- `uv.lock` (introduced on the branch, untracked on `main`) is intentionally **not** included here.

## Tests
Full suite green: **1400 passed, 1 skipped** (`pytest tests/unit tests/synthetic tests/agents tests/integration`). `ruff check` clean; `index.html` module validated with `node --check`.

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -8,8 +8,47 @@ from tokenjam.core.models import (
     NormalizedSpan, SessionRecord,
     SpanStatus, SpanKind,
 )
+from tokenjam.otel.semconv import GenAIAttributes
 from tokenjam.utils.ids import new_uuid, new_trace_id, new_span_id
 from tokenjam.utils.time_parse import utcnow
+
+
+def make_invoke_agent_span(
+    agent_id: str = "test-agent",
+    session_id: str | None = None,
+    conversation_id: str | None = None,
+    duration_ms: float = 0.0,
+    start_time=None,
+    trace_id: str | None = None,
+    service_namespace: str | None = None,
+) -> NormalizedSpan:
+    """Create an ``invoke_agent`` span.
+
+    ``duration_ms == 0`` -> a zero-duration turn-start marker, exactly what the
+    Claude Code / Codex logs path emits for every ``user_prompt`` event
+    (``end_time == start_time``). These mark the *start* of a turn and must NOT
+    complete the session.
+
+    ``duration_ms > 0`` -> a real session-wrapping span, what the SDK
+    ``@watch()`` path emits to bracket a whole agent run. This DOES complete the
+    session.
+    """
+    now = start_time or utcnow()
+    end = now + timedelta(milliseconds=duration_ms)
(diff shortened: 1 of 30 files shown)


anilmurty:
Reviewed the full lifecycle fix — no blockers from me. Root diagnosis is right: the logs path emits zero-duration `invoke_agent` markers every turn, and gating completion on `end_time > start_time` plus re-activating on later activity is the correct fix. Ran the touched suites locally, green. Rules clean — SQL is `$N`-bound, the new `/sessions/close` write endpoint is in `PROTECTED_PATHS`, migrations 8–11 idempotent. Three tiny non-blockers: `status_with_transcript_mtime` ships unused (fine as scaffolding), `_serialise` writes a `[sessions]` block on every config save, and the shell wrapper double-fires `session-end` on interrupt (idempotent, harmless). This is the base of the stack, so it merges first.

anilmurty (approved):
Approving per the review above — root lifecycle fix is sound, rules clean, suite green. Merging as the base of the stack (#371 follows).

PR #368, opened 2026-07-02
feat: tj context cost diagnostics (3/5 from #306)

anshss:
Carved from #306 as **PR 3 of 5**. The Claude-Code cost-diagnostics wedge — read-only analysis commands, no schema changes.

## What's here
- **`tj context`** — context-cost diagnostic: breaks a session's token spend into real work vs re-reads / recurring inclusions (files, searches, prompts, tool output), and names prompt-cache-miss overhead as its own source. The Claude-Code wedge card.
- **`tj quota-audit`** — retroactive Opus quota audit: finds past sessions that ran Opus where a downgrade would've been safe (`core/optimize/analyzers/model_downgrade.py`).
- **`tj tokenmaxx`** — reframed into a quota / efficiency card.

## Scope
- **No database migrations.** All commands are read-only over existing columns.
- Files: `core/context_diagnostic.py`, `cli/cmd_context.py`, `cli/cmd_quota_audit.py`, `cli/cmd_tokenmaxx.py`, `core/optimize/{types,__init__,analyzers/model_downgrade}.py`, `cli/main.py` (registration), tests. `core/framing.py` is reused from `main`.

## Carve note
`context_diagnostic._load_turns` selects `agent_id` (main's column) rather than `sub_agent_id` — that column arrives with PR 5's migration 14. The diagnostic's `sub_agent_id` JSON field is inert here; once PR 5 lands, a one-line follow-up can point the SELECT back at the real column.

## Tests
**1354 passed, 0 failed, 1 skipped**; `ruff check` clean.


anilmurty:
No blockers. The honesty framing here is exemplary — quota/token-share headlines, dollars gated to api/unknown plans and labeled "calibration only." Confirmed the `{where}` f-string in `model_downgrade.py` is the existing analyzer idiom (constants-only, values `$N`-bound), not new injection surface. 35 factory-based tests. Two cosmetic nits: `_export_snippet` uses `datetime.now()` instead of `utcnow()` (Rule 9), and `_alt_unit_cost` omits cache-write (conservative — never overstates savings). Fine to merge.

anilmurty (approved):
Approving per the earlier review. Resolved the cli/main.py command-registration conflict against #367 (union of both command sets — verified no_db_commands retains session-end/otel-resource-attrs and context/quota-audit correctly stay DB-backed). Full suite green locally (1423 passed), ruff clean.

Issue #240, opened 2026-06-23
Onboarding first-run experience: welcome banner + 'next steps' nudge


anilmurty:
The marketing promise is "install → onboard → value in <10 seconds / 3 steps." But the highest-value commands (`tj optimize`, `tj tokenmaxx`, `tj serve`) are exactly what a new user *doesn't* discover before getting distracted. Resolve the tension with **progressive disclosure**: keep the 3-step marketing headline, but have the CLI reveal the depth at the moment onboarding completes.

## 1. Welcome banner + version
`tj onboard` (and bare `tj`) currently open straight into a `Daily budget in USD` prompt — no branded moment, no version. Add a small branded banner (ASCII wordmark + `TokenJam vX.Y.Z` + one-line value prop) at the top. Build our own (don't copy any third-party banner code/licensing).

## 2. "▸ Next steps" nudge at the end of onboard
After onboarding, print a curated next-steps block. Two principles:
- **Lead with the no-restart wins.** Onboard ends on "Restart Claude Code" — the exact point we lose people. But `tj tokenmaxx`, `tj optimize`, and `tj serve` all work on the **just-backfilled data immediately**. So: "Your last 30 days are already loaded — try these now:" → `tj tokenmaxx` (shareable spend-tier), `tj optimize` (savings candidates), `tj serve` (open Lens) — *then* the restart note for live telemetry.
- **Curate, don't dump.** ~3 commands ordered by wow-factor, not a `--help` wall.

## 3. (Optional) bare `tj` "home" command
A re-orienting landing for returning users: quick status + next-best-action, so the nudge isn't a one-time thing they scroll past.

## 4. Smaller polish (same surface)
- Prompt order: onboard asks `Daily budget in USD` *before* `How do you pay for Claude?` — plan-first reads more naturally.

## Constraints
CLI-only; don't change the documented 3-step marketing flow itself. Honesty discipline intact (no over-promising in banner copy). Keep onboard idempotent/re-runnable.

PR #369, opened 2026-07-02
feat: zero-install / npm package (2/5 from #306, stacked on #368)

anshss:
Carved from #306 as **PR 2 of 5** — the zero-install / npm-package surface.

> **Stacked on #368** (`tj context` diagnostics). `tj quickstart` imports `compute_context_diagnostic`, so this branch includes #368's commits. Review this PR's **own 4 commits** (`be6c9a1..1f35730`), and merge **after** #368; once #368 lands, rebase and this diff collapses to just the quickstart/npm layer (11 files, +1042/−6).

## What's here
- **`tj quickstart`** — zero-setup first run: `npx tokenjam` / `uvx tj` reads `~/.claude` transcripts directly (no daemon, no pip env) and prints a context-cost summary. Capped on large `~/.claude` histories (#13).
- **npm wrapper** — `npm-wrapper/` publishes `tokenjam` (the bare `tj` name is squatted on npm); bare `npx tokenjam` routes to `quickstart`.
- **`core/session_timeline.py`** — pure-dataclass timeline records feeding the quickstart summary (zero tokenjam imports).
- Local bare `tj` keeps the branded **home screen** (#240); the npx surface is what routes to quickstart.

## Scope
- **No database migrations.** `session_timeline.py` avoids the `sessions.cache_write_tokens` column (added by a later PR); inert here.
- `cmd_otel` / `tj otel-resource-attrs` is **not** here — it ships with the lifecycle PR (#367).

## Tests
**1368 passed, 0 failed, 1 skipped**; `ruff` clean; `node --check npm-wrapper/bin/tj.js` OK; CLI smoke-tested (bare `tj` home screen, `tj quickstart`, `--version`).


anilmurty:
Nice — the npm wrapper is a genuinely thin, secure launcher: array-arg `spawnSync` with `shell:false` (no injection), no postinstall, downloads nothing itself, delegates to `uvx`/`pipx run` against the correctly-named `tokenjam` package. No duplicated Python logic. Two things before publish: (1) `npm-wrapper/package.json` is at `0.5.0` while everything else is `0.5.2` — align it now so the first publish isn't stale (Rule 15); (2) on Windows, `spawnSync` with `shell:false` won't resolve `uvx.cmd`, so the `has()` probe returns false even when uv/pipx are installed and the wrapper always hits the install-guidance fallback — `shell: process.platform === "win32"` (with arg quoting) fixes it. Stacked on #368, so it merges after. Neither gates merge.

anilmurty (dismissed):
Approving per the earlier review. Resolved the cli/main.py conflict as a 3-way union (main's #367+#368 command set + this PR's quickstart, with quickstart in no_db_commands). Full suite green locally (1437 passed), ruff clean.

anilmurty (approved):
Re-approving (prior review was auto-dismissed). Conflict resolution is the cli/main.py command-set union; CI green, suite passed locally.

PR #366, opened 2026-07-01
Add Bedrock Nova Micro/Premier, Anthropic-on-Bedrock, and LiteLLM provider pricing entries

Axiya3749:
…vider pricing entries

Closes #311

## Summary

Fills in previously-missing AWS Bedrock pricing entries (Nova Micro, Nova Premier,
Anthropic-on-Bedrock via the Opus 4.1 US profile and Opus 4.5 global profile) and
adds three commonly-used LiteLLM-routed providers (Mistral Large, DeepSeek
Chat/Reasoner, Cohere Command R+) to the packaged pricing table, following the
existing [provider.model] structure with verified current rates.

@anilmurty

## Related issue

Closes #311

## Checklist

- [x] Tests pass (`pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/`)
- [x] Lint clean (`ruff check tokenjam/`)
- [ ] Type check clean (`mypy tokenjam/`)
- [ ] CLAUDE.md updated (if architecture changed)
- [ ] Test spans use `tests/factories.py` (not raw `NormalizedSpan`)
- [x] Requested `@anilmurty` as reviewer (or @-mentioned him above)

Issue #94, opened 2026-06-15
schema: add cache_write_tokens column for per-token-class reporting


anilmurty:
Follow-up to #92. After that PR, `cache_write_tokens` flows through `NormalizedSpan` and `calculate_cost`, so cache-write spend is correctly **costed** — but the raw token count is never persisted. Only the resulting `cost_usd` lands in DuckDB.

## What's missing

The \`spans\` table has columns for \`input_tokens\`, \`output_tokens\`, and \`cache_tokens\` (cache-read) but **no \`cache_write_tokens\` column**. The count is used at costing time and then discarded.

## Why it matters

This blocks queries like:
- *\"How many cache-creation tokens did agent X emit last month?\"*
- *\"What percentage of cache-bound spend is creation vs read?\"*
- *\"Which sessions had unusually heavy cache-write?\"*

The \`cache\` analyzer (currently shows efficacy = cache_read / (cache_read + input)) could surface a creation-cost breakdown if it had the raw count.

## Asks

1. Add a \`cache_write_tokens INTEGER DEFAULT 0\` column to the \`spans\` table.
2. Update \`_row_to_span\` / \`_serialize_span\` (or whatever the equivalents are now) to round-trip the field.
3. Update \`tokenjam/core/ingest.py\` so the ingest pipeline writes \`span.cache_write_tokens\` into the new column.
4. Update aggregation paths (\`get_cost_summary\` etc.) to surface the new column where useful.

## No migration story needed (yet)

Pre-1.0, we're not promising backward-compat for existing DuckDB files. Implementer can just add the column directly without writing a migration step or worrying about existing-DB upgrade paths. Users with old DBs can run \`tj uninstall && tj onboard\` to start fresh; the analyzers will repopulate from \`~/.claude/projects/\` via the auto-backfill on first run.

## Out of scope

Threading \`cache_write_tokens\` through the Claude Code logs adapter + Langfuse/Helicone backfill — tracked separately in #93.

PR #370, opened 2026-07-02
fix: Claude Code cost & alert correctness (4/5 from #306)

anshss:
Carved from #306 as **PR 4 of 5** — Claude-Code cost accounting + alert-tuning correctness. Independent of the other four (off `main`).

## What's here
- **Cache-creation vs cache-read tokens** — `main` already tracks span-side `cache_write_tokens` (migration 5, #94); this adds the missing **sessions-side** rollup (`SessionRecord.cache_write_tokens`, sessions column, ingest aggregation) so per-session cache-write cost is correct. Migration **12**.
- **Current Anthropic / Haiku-4.5 rates** — corrects `claude-haiku-4-5` to 1.00/5.00/0.10/1.25 and adds `claude-fable-5`, `claude-sonnet-4-5` to `models.toml` (all of main's #366 Bedrock/Nova/LiteLLM rows preserved).
- **Alert false-positives on Claude Code** — suppresses spurious `retry_loop` / `drift` / `duration` alerts; `failure_rate` fires once per session instead of cascading.
- **Backfill correctness/perf** — reconciles backfilled session-row totals to the sum of their spans; bulk-inserts spans (chunked existence check + `executemany`) instead of per-span SELECT+INSERT (#15).
- **#18 trace-keyed cost rollup** — shipped as an **unwired `session_token_cost_rollup` db primitive** (+ tests). The status/detail wiring that consumes it lives in the UI PRs (#367 / the Approach-Map-Timeline PR), so this PR doesn't touch the status page.

## Migration
`12` — `cache_write_tokens` on `sessions` (the `spans` line is a no-op; column exists since migration 5). Numbered 12 to stay unique vs #367's 8–11 and the UI PR's 13–15.

## Reconciliation notes
- Branch's `cache_creation_tokens` → main's `cache_write_tokens` throughout (main's naming wins).
- `status.py` and `test_api.py` are byte-identical to `main`.

## Tests
**1353 passed, 0 failed, 1 skipped**; `ruff check` clean.


anilmurty:
No blockers — this is a clean correctness fix. Verified every pricing change against current Anthropic rates: the core one is `claude-haiku-4-5` moving off the retired 3.5 rates (0.80/4.00 → 1.00/5.00), plus new `fable-5`/`sonnet-4-5` rows. Cache-write aggregates separately from cache-read (dedicated test confirms, no double-count). The three alert fixes are right: retry-loop now needs identical arg-signatures + real execution spans (arg-sig hashed at ingest before content stripping, so Rule 5 holds), failure-rate fires once per session, duration/drift skip interactive coding agents. Two carry-forwards: `session_token_cost_rollup`'s trace disjunction is inert here but goes live when #371 wires it — worth a look there. And heads-up this overlaps #366 on `models.toml`, so whichever merges second will need a rebase.

anilmurty (approved):
Approving per the earlier review. Resolved the merge against #367: MIGRATIONS reordered 8→9→10→11→12 (append-only, no collision); upsert_session INSERT combined both sides' new columns (service_namespace + service_instance_id + cache_write_tokens, 16 cols/placeholders/params aligned); is_interactive_coding_agent comment+docstring merged to cover both alert-gating and /status classification; migration-count tests updated to 12. Full suite green (1458 passed), ruff clean.

PR #372, opened 2026-07-02
fix: deflake test_budget_cap_policy timezone-boundary test

anilmurty:
Extracts the standalone test fix from #363 so it can land in `main` immediately, independent of the async-hooks work (which has open review items).

## Why
`tests/unit/test_budget_cap_policy.py::test_cycle_spend_read_from_duckdb` (and `test_old_spend_outside_cycle_is_excluded`) seed spans relative to `utcnow()` and assert on cycle-scoped spend. When CI runs near a billing-cycle/month boundary in UTC, the "this cycle" spans fall into the previous cycle, the ceiling check flips `would_block` → `noop`, and the test fails. It's a real flaky test living in `main` — it just red-flagged the CI on unrelated PR #360 (a Docker-workflow-only change).

## What
- Seed a fixed mid-month datetime (`2026-06-15 12:00 UTC`) instead of `utcnow()` in the two cycle-spend tests.
- Pass a mocked `now_fn=lambda: now` into the policy engine so the evaluator's clock is pinned to the same instant. (`PolicyContext` already supports `now_fn` on `main` — `tokenjam/proxy/engine.py:111` — so this is test-only; no production change.)

## Tests / Verification
- `ruff check tests/unit/test_budget_cap_policy.py` — clean
- `pytest tests/unit/test_budget_cap_policy.py` — 12 passed
- The tests are now deterministic regardless of wall-clock time.

## Scope
Test-only, one file. Credit to @tarun73, who wrote this fix as part of #363 — this PR just cherry-picks it out so we can deflake `main` now.

Co-Authored-By: tarun73 <noreply@github.com>

### DIFF
--- a/tests/unit/test_budget_cap_policy.py
+++ b/tests/unit/test_budget_cap_policy.py
@@ -11,7 +11,7 @@ traffic; and the envelope carries the unvalidated label + a would-do action
 """
 from __future__ import annotations
 
-from datetime import timedelta
+from datetime import datetime, timedelta, timezone
 
 import pytest
 
@@ -40,7 +40,7 @@ def _cfg(*, ceilings: dict[str, float | None], policy_params: dict | None = None
     )
 
 
-def _engine(cfg: TjConfig, *, spend, db=None) -> PolicyEngine:
+def _engine(cfg: TjConfig, *, spend, db=None, now_fn=None) -> PolicyEngine:
     # `spend` is a {provider: usd} map (or a single float) injected as the
     # cycle-spend source so the evaluator is testable without a DB.
     if callable(spend):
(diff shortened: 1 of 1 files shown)


anilmurty:
Pulling this out of #363 on its own so we can deflake `main` right away. The async-hooks work in #363 still has open review items, but this test fix is independent and doesn't need to wait on them.

Context on why it's worth landing now: this exact test just failed CI on #360 — a PR that only adds a Docker workflow and touches zero Python. The flaky boundary condition in `main` is generating false negatives on unrelated PRs, so getting it in stops the noise.

Full credit to @tarun73 — this is their fix, cherry-picked verbatim (authorship preserved). Verified locally: ruff clean, 12/12 tests pass, and it's now deterministic instead of clock-dependent.

PR #371, opened 2026-07-02
feat: session Approach / Map / Timeline (5/5 from #306, stacked on #367)

anshss:
Carved from #306 as **PR 5 of 5** — the marquee feature: drill into *how* an agent worked.

> **Stacked on #367** (session lifecycle). This branch is built on PR#1, so its diff-vs-`main` includes #367's commits. Review the **own delta** (50 files, +13.3k, from `pr1/session-lifecycle..HEAD`) and merge **after** #367.

## What's here
- **Session detail view** — drill into one session from the Status page: model mix, context-growth, per-subagent cost breakdown.
- **Timeline** — deterministic transcript play-by-play (recursive subagent logs, no LLM), from `core/transcript.py`.
- **Map** — graphical board of what the agent did per ask: phase / tools / sub-agents / context / cost swimlanes + a codebase-territory treemap (`core/workmap.py`, `core/sessionmap.py`, `core/phases.py`).
- **Approach** — deterministic method spine (`act`/`delegate`/`verify`/`dead_end`), recursive, with cross-terminal child splice (`core/method_spine.py`).
- **Method persistence** — snapshots the reconstructed Story to `session_story` at close + backfill so a killed agent's method survives transcript pruning (`core/method_capture.py`); re-couples the `capture_session_method` call into #367's close handler.
- **Cross-session run grouping** — `tokenjam.run_id`/`parent_session_id`, `/api/v1/runs`, `setup_harness` MCP tool (`core/runlink.py`, `api/routes/runs.py`).
- **Per-subagent cost** — `sub_agent_id` on spans + subagent right-sizing analyzer.
- **On-demand LLM-distilled Map titles** via local `claude` CLI (`core/distill.py`, no API key).

## Migrations
`13` run_id + parent_session_id · `14` sub_agent_id on spans · `15` session_story table. (12 intentionally skipped — reserved by #370.)

## Cross-PR seams for merge (resolvable)
- Shares `SessionRecord.cache_write_tokens` (model field, default 0 here) and `core/backfill.py` with #370 — trivial conflicts when both land. Per-session cache-write totals read 0 until #370's migration 12 provides the column.
- The `#18` `session_token_cost_rollup` primitive lands in #370; this PR's detail cost uses stored values until then.

## Tests
**1658 passed, 0 failed, 1 skipped**; `ruff check` clean; `index.html` module `node --check` OK; lens string-grep regression suite green.

## Not browser-dogfooded
[...]

PR #374, opened 2026-07-02
feat(onboard): zero-token statusline for Claude Code/Codex; reserve MCP for SDK (#59)

anshss:
## Move Claude Code / Codex onboarding OFF the in-loop MCP, ONTO a zero-token statusline

**Why (the measured rationale).** Delivering tj as an in-loop **MCP** server is a per-turn quota burden on Claude Code / Codex **subscription power users**: a measured A/B showed tj-MCP-in-loop consumed **+36% model-weighted quota** vs a no-tj control (Claude's own subagent-inclusive cost). The entire CC-usage ecosystem (ccusage, usage-monitor, ccstatusline, sniffly) is deliberately out-of-band for exactly this reason; tj's `mcp__tj__*` was the one in-loop outlier.

**The line this PR draws.** The MCP is now an **SDK / API-only** feature (tj sits in the request path there for real-time enforcement / policy / budgets). Claude Code & Codex subscription users get tj **out-of-band** via a zero-token statusline plus the existing CLI reports and OTel telemetry ingest. `tj mcp` still works for anyone who wants it — onboarding just no longer defaults CC/Codex users into it.

### Deliverables

1. **`tj statusline` subcommand** (`cli/cmd_statusline.py`, registered in `cli/main.py`). Reads Claude Code's statusline JSON on stdin, locates the session JSONL (`transcript_path`, else globs `~/.claude/projects/**/<session_id>.jsonl`), and prints one line — model + total tokens + **re-read share** (cache-read ÷ total tokens over the session's assistant turns, deduped by message id) + an actionable `/compact` nudge past warn (70%) / crit (88%) thresholds. **Fail-safe** (any error → minimal line or nothing, always exit 0 — a broken statusline must never break the terminal), **fast** (pure stdlib, no network / no DB), **zero model tokens**.

2. **Onboarding wires the statusline, not the MCP (Claude Code).** `tj onboard --claude-code` writes `"statusLine": {"type": "command", "command": "<abs>/tj statusline"}` into `~/.claude/settings.json`, **idempotently + non-destructively** — a human-authored or third-party (e.g. ccstatusline) `statusLine` is a contract we never clobber (detect → warn → skip). The `claude mcp add tj` call is removed. Existing OTel telemetry env wiring is kept.

3. **Codex.** Codex CLI has no statusline / status-hook surface today, so onboarding stays fully out-of-band via OTel, **retires** any previously tj-registered `[mcp_servers.tj]` block (only its own managed block, never a user's same-named server), and documents the state. No MCP-by-default fallback — the burden does not regress onto Codex.

4. **MCP reframed as SDK-only** across `cli/cmd_mcp.py`, `mcp/server.py` `setup_project`, `cli/cmd_doctor.py` (MCP *absence* for a coding agent is now `info`, not a `warning`; a new statusline-wiring check nudges CC users instead), plus README and docs.

### Evidence

**Real session** (`~/.claude/projects/**/*.jsonl`, 137.6M-token session):
```
◆ opus-4-8  137.6M tok  🕳️ re-read 96%  → /compact to reclaim quota
```

**Bad input** (garbage / empty / non-dict JSON / binary / missing transcript) → minimal line `◆ ?` (or model-only), **exit 0**, no traceback in every case.

**Onboard in a temp HOME** → `settings.json` gets the `statusLine` key pointing at `tj statusline`, **no** `claude mcp add`, no `~/.claude.json`, no MCP references anywhere in `~/.claude`. A pre-existing foreign `statusLine` (ccstatusline) is left exactly intact and env keys are preserved.

**Checks:** `make test` → 1421 passed, 2 skipped (22 new unit tests). `make lint` and `make typecheck` clean.

Resolves #59. Do not merge without review.


anilmurty:
Reviewed this closely given it changes the default integration for every new Claude Code user — approving. I especially like that you led with the measured +36% quota rationale; reserving in-loop MCP for the SDK/enforcement path and moving CC/Codex out-of-band (statusline + OTel) is the right line, and it matches how the whole CC ecosystem (ccusage, ccstatusline) already operates.

The safety properties hold up under adversarial testing, not just reading:
- **Fail-safe is bulletproof** — binary / garbage / empty / non-dict stdin and missing transcripts all exit 0 with a minimal line, no traceback. A broken statusline can't break the terminal.
- **settings.json editing is non-destructive + idempotent** — a foreign `statusLine` (ccstatusline or human-authored) is detected and skipped, never clobbered, and the OTel env wiring is preserved.
- **Codex retirement only strips tj's own managed block** — a user's same-named server is left intact.
- **#371 session-lifecycle wiring survives the onboard rewrite** — the `session-end` trap and `otel-resource-attrs` export are still installed.

Two non-blocking cleanups before/around merge:

1. **Stale doc line.** `CLAUDE.md:319` still says the Codex re-onboard is a no-op "only when both `[otel]` and `[mcp_servers.tj]` are present" — but this PR changed that check to `[otel]`-only. You updated the neighboring Codex bullets; just missed this one. Please fold the fix in.

2. **Duplicated token-parsing with a subtle divergence.** `session_shares()` reimplements the four-bucket usage parse + message-id dedup that already lives in `core/backfill.py`, but uses **first-wins** dedup where backfill uses **last-wins** (to capture finalized cumulative usage). Net effect: on a mid-stream snapshot the statusline's re-read % can slightly disagree with the Cost tab's number for the same session. Small, but inconsistent numbers between the statusline and Lens read like a bug and undercut trust in the figures. A shared `_iter_assistant_usage()` helper would keep them from drifting — fine to do here or as a fast follow, your call, but let's not leave it indefinitely.

Nice work. Approving; let's get it in before the next release cut since it changes the first-run experience.

anshss:
@anilmurty can merge this

anilmurty (approved):
Re-approving after conflict resolution against #379/#380/#381 (union of statusline + loop command registration; doctor checks renumbered 13/14 to avoid collision). Full suite green locally (1806 passed), ruff clean. Statusline fail-safe + non-clobber verified in the original review. Follow-up #383 filed for the statusline/backfill dedup unification.

anilmurty (approved):
Re-approving (prior review dismissed). Conflict resolution verified: statusline + loop command union, doctor checks 13/14, 1806 tests green locally, CI green.

PR #379, opened 2026-07-02
test: cross-backend StorageBackend parity suite (#51)

anshss:
## Summary

Adds a parametrized cross-backend parity/contract suite so the `tj serve` HTTP shim (`ApiBackend`) can no longer silently diverge from the direct DuckDB backend. This is the structural follow-up behind the individually-filed serve-mode/DB-mode divergence bugs (`cache_write_tokens` dropped on daemon-fetched spans, cache columns zeroed, `get_daily_cost` returning a cumulative total, …).

`tokenjam` ships three `StorageBackend` implementations — `DuckDBBackend`, `InMemoryBackend`, and the read-only `ApiBackend` serve shim — and nothing enforced behavioral parity, so every new/changed method could diverge unnoticed.

## What's in this PR

`tests/integration/test_storage_backend_parity.py`:

- **`test_shim_matches_db`** — each faithfully-mirrored read method is run against **both** the DuckDB backend and the `ApiBackend` shim, the shim talking to a **real in-process uvicorn server** over the same DB (genuine HTTP + JSON round-trip — where shim bugs actually live; a sync `ApiBackend` can't drive the in-process ASGI shortcut). Fails on any divergence. Covered: `get_traces`, `get_trace_spans` (incl. `cache_tokens`/`cache_write_tokens`), `get_cost_summary`, `get_alerts`, `get_tool_calls`, `get_daily_cost` (seeds a prior-day span so the old cumulative bug would be caught), `get_baseline`.
- **`test_duckdb_and_in_memory_agree`** — DuckDB file backend vs `InMemoryBackend` on identical seeded data, so the other two shipped backends stay in lockstep.
- **Enforcement layer** — every `StorageBackend` protocol method must be classified as parity-covered / documented known-gap / deliberately-unimplemented / lifecycle. Adding a protocol method (`test_every_protocol_method_is_classified`) or teaching the shim a new method (`test_unimplemented_methods_have_no_silent_shim`) fails CI until it's classified — which for a read method means adding a parity assertion. This codifies "new `StorageBackend` method ⇒ parity test required".

The 4 current shim divergences are all **documented intentional stubs/degraded views** (`get_session_cost`/`get_recent_spans`/`get_completed_session_count` stubs; `get_completed_sessions` synthesized from `/api/v1/status`), recorded in `SHIM_KNOWN_GAPS` so they're visible instead of silent.

Also codifies the classification rule in `CLAUDE.md` → "StorageBackend parity (serve shim vs DB)".

## Tests / Verification

- `pytest tests/integration/test_storage_backend_parity.py` → **18 passed**.
- **Guard verified to bite:** temporarily reintroducing the `cache_write_tokens` drop in `_dict_to_span` turns `test_shim_matches_db[get_trace_spans]` red (`25` vs `None`); reverting turns it green.
- Full `pytest tests/integration/` → **342 passed**.
- `ruff check` clean; `mypy` clean on the new file.
- Wired into CI automatically — lives under `tests/integration/`, which `ci.yml` already runs.

## What's NOT in this PR

The 4 known shim gaps are documented, not fixed — closing them is separate work (see related #61 / #63). This PR guards against *new* divergences and makes the existing gaps explicit.

Closes #51

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -239,6 +239,12 @@ For `GET /api/v1/drift`, if `agent_id` is missing, return `JSONResponse(status_c
 
 Integration tests use `httpx.AsyncClient` with `httpx.ASGITransport(app=app)` against `InMemoryBackend`. Synthetic alert tests use `unittest.mock.MagicMock` for the DB — you must explicitly set up `db.get_recent_spans.return_value` before calling `engine.evaluate()`, and silence channels with `engine.dispatcher.channels = []`.
 
+### StorageBackend parity (serve shim vs DB)
+
+`tj` ships three `StorageBackend` implementations: `DuckDBBackend`, `InMemoryBackend` (tests), and `ApiBackend` — the read-only HTTP shim the CLI uses when `tj serve` holds the DB write-lock. The shim reconstructs objects from JSON, so it repeatedly diverged from the DB backend *silently* (e.g. dropped `cache_write_tokens` on daemon-fetched spans, cache columns zeroed, `get_daily_cost` returning a cumulative total). That whole class of bug (#51) is now guarded by `tests/integration/test_storage_backend_parity.py`, which runs each faithfully-mirrored read method against both the DuckDB backend and the `ApiBackend` shim (over a **real in-process uvicorn server**, not the ASGI shortcut a sync `ApiBackend` can't use) and fails on any divergence.
+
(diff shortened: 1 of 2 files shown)


anilmurty:
Approving — this is exactly the right structural fix for the serve-shim divergence class, and I like that it tests over a real uvicorn server rather than the ASGI shortcut, since that JSON round-trip is where the shim bugs actually live. The self-maintaining enforcement is the best part: `test_every_protocol_method_is_classified` means a future `StorageBackend` method can't dodge parity coverage, so this doesn't rot into a one-time sweep. Verified the guard bites on the `cache_write_tokens` projection and that `get_daily_cost` catches the cumulative bug via the 3-day-back seed.

Two non-blocking notes: `get_tool_calls` parity only compares the count, not field values — a touch weaker than the others if you want to tighten it later; and the live-server thread is a small potential flake vector under heavy CI, though the timeouts are generous. Neither gates merge. Nice work closing this hole for good.

anilmurty (approved):
Approving per review — cross-backend parity suite, tests over real HTTP, self-maintaining enforcement. No blockers.

PR #55, opened 2026-05-08
refactor: complete ocw→tj/tokenjam naming migration

anshss:
## Summary

- Sweeps all remaining `ocw` CLI references across 52 files — docs, examples, incidents, tests, and source
- Renames Prometheus metric prefix `ocw_*` → `tj_*` in `tokenjam/api/routes/metrics.py`
- Updates TypeScript SDK OTel attribute namespace `ocw.*` → `tokenjam.*` and default service name `ocw-ts-sdk` → `tj-ts-sdk`
- Updates Python import paths `from ocw.sdk.*` → `from tokenjam.sdk.*` in all examples
- Updates config/path references (`ocw.toml` → `tj.toml`, `~/.ocw/` → `~/.tj/`) in docs

## Test plan

- [ ] `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` — all 293 tests pass
- [ ] `cd sdk-ts && npm test` — TypeScript SDK tests pass with updated semconv assertions
- [ ] `ruff check tokenjam/` — no lint errors introduced
- [ ] Grep for remaining `ocw` refs (excluding intentional guard in CLAUDE.md): `git grep -i '\bocw\b' -- ':!CLAUDE.md'`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #381, opened 2026-07-02
Fix #55: self-heal recorded-but-unlanded migrations (silent ingest drop → blank Status)

anshss:
A live `~/.tj/telemetry.duckdb` had migration 7 recorded in `schema_migrations` but the `request_params` / `request_tools` columns it adds absent from `spans`. Every ingest of a token-bearing record then hit a DuckDB Binder Error and was silently dropped — surfacing to the user as a blank/stale Status page with active sessions showing no data.

## Summary
- `run_migrations` no longer trusts the recorded version set alone: it now reconciles the additive columns the code depends on on every open, so a recorded-but-unlanded migration self-heals.
- New `tj doctor` "Schema integrity" check WARNs when the live schema is missing an expected column; `tj doctor --repair` reconciles it.

## Root cause
`run_migrations` keys purely on the version INTEGER. If a version was recorded-applied under an older or renumbered definition (this branch renumbered migrations during the main merge — PR #306's 6/7-vs-8–13 reconciliation), the *current* SQL for that version never re-runs and its `ADD COLUMN` silently never lands. `ADD COLUMN IF NOT EXISTS` only helps while the version is still *unrecorded*.

## Fix
- `EXPECTED_ADDITIVE_COLUMNS` single-sources the additive columns later migrations append to `spans`/`sessions`.
- `ensure_expected_columns(conn)` re-issues `ADD COLUMN IF NOT EXISTS` for that set regardless of recorded versions — idempotent (pre-checked + `IF NOT EXISTS`), returns the columns it added. `run_migrations` calls it on every open and logs a WARN naming any columns it had to reconcile.
- `missing_expected_columns(conn)` is the read-only counterpart powering the doctor check.
- `cmd_doctor` gains `_check_schema_integrity` (level `warning` + `repair_action: heal_schema`) and a `heal_schema` branch in `_attempt_repairs`.
- Maintenance note in `CLAUDE.md`: when appending a migration that adds a code-read/written column, add it to `EXPECTED_ADDITIVE_COLUMNS`.

## Tests / Verification
- `tests/unit/test_schema_self_heal.py` (10 tests): simulates the recorded-version/missing-column state (drop migration 7's columns while leaving version 7 recorded) and asserts `missing_expected_columns` detects it, `run_migrations`/`ensure_expected_columns` self-heal on next open (idempotent), and — the user-visible regression — a real span carrying `request_params` ingests through `IngestPipeline` instead of being dropped on a Binder Error. Also proves the raw insert *does* raise without the heal.
- `tests/integration/test_cli.py` (+2): drives the real `tj doctor` CLI against a `DuckDBBackend` reproduced in the #55 state — asserts the "Schema integrity" WARN (exit 1) and that `--repair` reconciles the columns and a follow-up run reports OK.
- Full CI layers green locally: `pytest tests/unit/ tests/synthetic/ tests/integration/` → 1729 passed, 1 skipped. `ruff` + `mypy` clean on changed files.
- End-to-end reproduction: opened a DB, dropped migration 7's columns with version 7 recorded, reopened through `DuckDBBackend` → self-heal WARN logged, columns reconciled, `missing_expected_columns` → `[]`.


anilmurty:
Approving — this is the right fix at the right layer. I cross-checked `EXPECTED_ADDITIVE_COLUMNS` against the full `MIGRATIONS` list and it's complete: every additive column from migrations 4–14 is covered, including the run_id/parent_session_id/sub_agent_id/cache_write_tokens set from the #306 renumbering that caused this class of bug. The self-heal is idempotent and DuckDB-correct, and the tests reproduce the exact #55 state and prove ingest recovers. Good catch that `run_migrations` trusting the version integer was the root cause.

Two follow-ups, neither blocking: (1) the heal covers columns but not renumbered `CREATE TABLE` migrations — I verified those tables (session_labels/session_story/loop) aren't in the ingest hot path so they'd error visibly rather than silently drop, but let's extend the self-heal (or a doctor check) to missing tables too (filing an issue). (2) `tj doctor --repair` still exits 1 after a successful heal since the check list isn't re-evaluated — minor, matches the existing `rebuild_spans` pattern. Merging into the release since it hardens the upgrade path.

anilmurty (approved):
Approving per review — migration self-heal, EXPECTED_ADDITIVE_COLUMNS complete against the full migration list, idempotent + DuckDB-correct, tests reproduce the #55 state. Follow-up #382 filed for missing-table healing.

anilmurty (approved):
Re-approving after branch update (migration self-heal, EXPECTED_ADDITIVE_COLUMNS verified complete). CI green.

Issue #238, opened 2026-06-23
Backfill session-count reporting is confusing (new-only summary + progress-vs-table mismatch)


anilmurty:
Surfaced during a real first-run walkthrough. Two related counting problems make `tj onboard --claude-code` / `tj backfill claude-code` confusing.

## 1. Summary reports newly-added only → reads as "it barely worked"
Backfill is idempotent (deterministic span IDs), so re-running against a DB that already has history counts only *new* sessions. On a machine with a prior backfill, `tj onboard --claude-code` printed:

```
Backfilled: 1 sessions, over 46 days, $383 total
```

…while the DB actually held **13 sessions / 24.5k spans / 46 days**. A first-run user reads "1 session" as a failure. The summary should report the **total in-window** state, e.g. `1 new (12 already present) · 13 total over 46 days`, not new-only.

## 2. Progress counter unit ≠ table records
The progress line counts up to **"363 of 363 sessions … from 5 projects"**, but only **13 `SessionRecord`s** land in the `sessions` table:

```
✓ Backfilled 363 of 363 sessions, over 46 days, from 5 projects
-- sessions table: 13 rows
```

So "session" means two different things (per-conversation-file in the progress counter vs. grouped `SessionRecord` in the table). Either the progress label is wrong (it's counting conversations/files, not sessions), or the SessionRecord grouping is collapsing 363→13 more aggressively than intended — worth confirming which, since 13 sessions for 24.5k spans / $4.4k over 46 days looks low.

## Acceptance
- Backfill/onboard summary reports new / existing / total clearly.
- The progress "session" unit matches what lands in the `sessions` table (or is relabeled to what it actually counts, e.g. "conversations").
- A regression test pins the reported counts against the table (see the companion test-gap issue).

## Not a bug (ruled out during investigation)
Plan-tier stamping itself is correct — a clean backfill with a `max_5x` config stamps every session `max_5x`. This issue is purely about **count reporting**, not plan resolution.

PR #387, opened 2026-07-03
Fix #60: flag delegating sessions with unaccounted subagent turns in weighted quota

anshss:
Claude Code sessions that delegate to **Task subagents** were under-counted by tj's model-weighted quota metric. The 2026-07-02 quota A/B (arm C) measured tj at ~half of Claude Code's subagent-inclusive `result.total_cost_usd`: the parent transcript held only **5 of 23** assistant turns — the rest were subagent turns.

## Summary
- Fix a mis-attribution in the weighted-quota metric (`core/context_diagnostic.py`): `_load_turns` selected `agent_id` but bound it to `TurnComposition.sub_agent_id`, so the metric never actually read which turns were subagent turns (and `heaviest_turns[].sub_agent_id` in the JSON payload silently emitted `agent_id`). Now selects the real `sub_agent_id` column.
- Add the **partial-accounting flag** the ticket asks for: when a session delegated (a `Task` tool span in the parent) but its subagent transcript is missing, the weighted quota is silently low. New `subagent_turns` / `delegating_sessions` / `unaccounted_subagent_sessions` fields + a `subagent_accounting_partial` property + a note surfaced by `tj context`.

## Root cause & fix
tj recomputes the weighted quota from on-disk turns. Subagent turns live in separate `subagents/agent-*.jsonl` files, which prior work (backfill walking + `sub_agent_id` + session-total reconcile) already ingests — so the *capture* half of the ticket had landed, and the subagent turns were already summed into the weighted totals. What remained:

1. **Attribution was wrong** — `_load_turns` never read the `sub_agent_id` column. Fixed.
2. **No honesty flag for the unaccounted case** — a delegation whose subagent turns are missing made the number silently low. Now detected via the parent's `Task` tool span and flagged as a **lower bound** (CLAUDE.md Rule 14) instead of reported as a complete total.

## Tests / Verification
- `tests/unit/test_context_diagnostic.py`: sidechain/subagent fixture covering both halves —
  - a delegating session with captured subagent turns → subagent turns counted in the weighted quota, **not** flagged;
  - a delegating session with a missing subagent transcript → flagged partial, listed in `unaccounted_subagent_sessions`, note surfaced;
  - fields round-trip through the `--json` payload.
- `make_tool_span` gains optional `session_id` / `start_time` / `sub_agent_id` (backward-compatible) so a `Task` delegation span can be placed in a session/window.
- **End-to-end on the real arm-C A/B session** (`ingest_claude_code` → `compute_context_diagnostic`): the diagnostic now includes all **60** subagent turns and its output-token total matches ground truth exactly (**30939**), with no false partial flag.
- Full `tests/unit/` + `tests/synthetic/` (1396 passed, 1 skipped) and `tests/integration/` (324 passed) green. `ruff` + `mypy` clean on touched files.

## What's NOT in this PR
- No change to the backfill *capture* path — walking `subagents/*.jsonl`, `sub_agent_id` stamping, and session-total reconcile already landed in prior work (#238/#245/#294) and are verified above.
- Detection relies on the parent's `Task` tool span as the delegation signal; a subagent spawned outside the on-disk transcript with **no** parent-side signal at all is genuinely undetectable and is not fabricated.

Closes #60


anilmurty (approved):
Approving per review — fixes the agent_id/sub_agent_id attribution bug + honest lower-bound flag; 18 factory tests pass. Merging for a clean board.

Issue #68, opened 2026-05-28
Pre-release testing findings (v0.3.0 sprint) — bugs and playbook fixes


anilmurty:
Running list of findings surfaced while walking through `tests/manual-pre-release-testing.md` against the v0.3.0 sprint work on `task-5-docs-sweep`. Filed proactively so we can keep updating as the playbook run continues.

### 1. `tj onboard --reconfigure` bare-path early-returns
The `--reconfigure` flag is honored inside `_onboard_claude_code` and `_onboard_codex`, but the top-level `cmd_onboard` early-returns with "Config already exists. Use --force to overwrite." when called without `--claude-code` / `--codex`, regardless of `--reconfigure`.

**Fix:** Make the bare-onboard path honor `--reconfigure` (bypass the existing-config check) or explicitly document that `--reconfigure` only applies to integration-specific flows.

**Where:** `tokenjam/cli/cmd_onboard.py` ~ line 39–43

---

### 2. SDK silently drops spans on HTTP 401
When the SDK's `HttpTransport` gets a 401 from a running `tj serve` (e.g., secret mismatch), it logs a single line — `tj serve returned 401 on span export` — and the spans are dropped. The user can easily miss the message; no fallback to direct DuckDB write; no counter exposed via `tj doctor`.

**Repro:** Have `.tj/config.toml` and `~/.config/tj/config.toml` with different `ingest_secret` values, then run any example script while `tj serve` is up. Watch the spans never appear in `tj status`.

**Fix options:**
- Fall back to direct DB write when HTTP push fails (preferable for local-first ergonomics)
- Raise visibility: count dropped spans, surface in `tj doctor`
- At minimum, change the warning to ERROR level and include the secret-fingerprint mismatch hint

**Where:** `tokenjam/sdk/transport.py`

---

### 3. `tj drift` doesn't surface baselines that demonstrably exist (HTTP API fallback)
After running `examples/alerts_and_drift/drift_demo.py` (12 baseline + 1 anomalous session) and observing `drift_detected` fire successfully, `tj drift` reports "No drift baselines found." The detector clearly built and used a baseline (otherwise the alert wouldn't fire), but the CLI doesn't surface it.

`tj doctor` correctly identifies other agents as "Collecting baseline: sensitive-demo (0/10), budget-demo (0/10)" but doesn't mention `drift-demo` — suggesting its baseline IS built but `tj drift` can't read it.

[...]

PR #388, opened 2026-07-03
fix(mcp): route get_optimize_report through serve when DB is locked (#61)

anshss:
## Ticket #61

`get_optimize_report` threw `IO Error: Could not set lock on file` on **every** call whenever a `tj serve` was running.

### Root cause
When `tj mcp` reaches or spawns a `tj serve`, the MCP is initialised with `_serve_url` set and `_ro_conn = None`. In that state `get_optimize_report` called `open_db(_config.storage)`, which opens a **read-write** DuckDB connection. DuckDB's file lock is exclusive, so while the serve process holds the write lock the open fails with a raw lock exception — and `open_db` also returns a fresh backend per call, self-conflicting across repeated calls.

Verified empirically: while a live `tj serve` holds the lock, **both** a read-write `open_db` *and* a read-only attach throw `Could not set lock on file`. So the ticket's "read read-only" suggestion is infeasible — the only lock-free path is going through the serve process.

### Fix
Route the optimize computation through the serve HTTP API (`GET /api/v1/optimize`) via the existing `ApiBackend.fetch_optimize_report`, exactly as the CLI does under daemon mode (#68 §12). The direct-DB branch no longer opens a competing read-write connection: it reuses the single read-only connection when present, otherwise returns the graceful *"Optimize requires a direct database connection"* guidance instead of a raw lock exception.

### Tests
- `test_get_optimize_report_serve_holds_lock_routes_over_http` — holds a real DuckDB write lock, asserts serve-mode routes over HTTP and returns the report (no lock error).
- `test_get_optimize_report_no_connection_returns_graceful_message` — asserts the no-connection state returns the graceful message, not a lock error.

### Validation (real end-to-end loop)
Brought up a **real `tj serve`** on a seeded temp DB (separate process, holding the write lock), then drove the real MCP init path (`init(ro_conn=None, config, serve_url=...)`) and called `get_optimize_report`:

| Component | Result |
|---|---|
| Control: `open_db(read-write)` under live serve | **LOCK ERROR** — `Could not set lock on file` (reproduces the bug) |
| `get_optimize_report` (serve-mode, the #61 state) | **PASS** — returned full report (findings: cache, script, reuse, trim…), no lock error |

- `pytest tests/unit/` → 1349 passed, 1 skipped
- `ruff` + `mypy` on `server.py` → clean


anilmurty (approved):
Approving per review (critical rules clean, CI green). Merging for a clean pre-release board.

PR #390, opened 2026-07-03
fix(#63): make `tj context` render while `tj serve` is running

anshss:
## Summary

Fixes #63 — `tj context` (the launch-hero command) refused to run while `tj serve` daemon was up, telling users to stop the daemon on the exact command the launch drives them to.

- **Root cause:** the diagnostic reads the raw `attributes` column for recurring-inclusion detection, which the API shim never exposed row-by-row. DuckDB refuses a concurrent read-only connection alongside the write lock, so a client-side workaround isn't available.
- **Fix:** route the compute through the daemon (mirrors `/api/v1/reuse/clusters` #154 and `/api/v1/optimize` #68):
  - new `GET /api/v1/context` computes the diagnostic + plan-tier `framing` server-side (`tokenjam/api/routes/context.py`)
  - new `diagnostic_from_dict` inverse of `diagnostic_to_dict` in `tokenjam/core/context_diagnostic.py` so the CLI reconstructs the dataclass and renders identically
  - `cmd_context.py`: on `ApiBackend` (no `.conn`), fetch + render via `_render_via_serve`; on a truly non-conn/non-Api backend, still surface a clean error
  - `ApiBackend.fetch_context_diagnostic` HTTP client shim

## Test plan

`tests/integration/test_context_serve.py` (all 4 pass):
- [x] endpoint computes diagnostic server-side — recurring inclusion from raw `attributes` surfaces
- [x] `tj context` renders through in-process ASGI-wired `ApiBackend` (the exact daemon-is-running state) — old refuse-to-run error is gone
- [x] `tj context --json` returns identical payload through serve
- [x] a non-conn/non-Api backend still errors cleanly, no traceback

Existing `tests/unit/test_context_diagnostic.py` (18 tests) still green.


anilmurty (approved):
Approving per review — tj context serve-fallback mirrors the optimize/reuse pattern, parity + honesty preserved, both paths tested. Merging for a clean board.

PR #401, opened 2026-07-03
feat: tj resume-brief — hand a resuming session its prior method, ~40–60% fewer turns

anshss:
## What it does

`tj resume-brief` reconstructs a compact brief — **task · progress · what-was-tried/dead-ends · where-it-left-off · working files** — from the session tj *already persists*, and (via a SessionStart/PostCompact hook) re-injects it so a continuing or post-compaction session resumes instead of re-investigating.

It reuses tj's `build_session_story` + `build_method_spine` for the method trail and reads the durable `session_story` snapshot (`core/method_capture`, which survives Claude Code's transcript prune), falling back to a live transcript path. **Zero in-loop cost** — deterministic, <1s, out-of-band; ~400 injected tokens.

Two reliability fixes found while evaluating the prototype:
- **Task extraction skips slash-command-only openers** (`/clear`, `/model`) — it takes the first *substantive prose* ask, not the literal first message.
- **The brief is scoped to the last ask/phase** (reusing `build_session_asks`) so progress/files don't sprawl across unrelated earlier work in long multi-task sessions.

## How we tested it + the efficiency result

Real headless A/B: a **cold** session vs a **brief-equipped** session, each continuing the *same interrupted task*, with a fresh fixture per trial — measuring turns / cost / task-success / investigation-steps-before-fix, n=4/arm.

- **Concept fixture** (bug with a non-obvious cause): cold **8.0 turns / $0.361** → brief **4.75 turns / $0.250** = **−41% turns, −31% cost**; correctness 4/4 both arms.
- **Hard fixture** (a genuinely non-derivable, project-specific quirk that forced cold agents to trace 6 files + run an instrumentation probe — 9 investigation steps): cold **14.0 turns / $0.522** → brief **5.5 turns / $0.332** = **−61% turns, −36% cost**; correctness 4/4; the **instrumentation probe was eliminated** (4/4 cold → 0/4 brief). The benefit scales with how expensive the lost knowledge is.

Honest bound: this is a *bounded efficiency gain on a continuation* — correctness held throughout, it is not a rescue. It targets the seams where Claude Code loses method: post-compaction, cross-session, and ephemeral subagents.

## Tests added

- `tests/unit/test_resume_brief.py` — synth from fixture sessions: slash-opener skip, ask-scoping (progress/files scoped to the last ask), TodoWrite pending/in-progress extraction, interruption capture, tried/dead-end surfacing, and fail-soft on malformed input.
- `tests/unit/test_onboard_resume_hook.py` — SessionStart hook wiring: idempotency, non-destructive merge (foreign SessionStart hooks preserved), stale-entry update-in-place, and the uninstall path.
- `tests/unit/test_transcript.py` — new case locking the real-shaped `{"todos":[...]}` TodoWrite payload (ticket #67): pending/in-progress statuses + `activeForm` fallback survive parsing.

`make test` (1941 passed, 2 skipped), `make lint`, and `make typecheck` all green.

### DIFF
--- a/tests/unit/test_cmd_resume_brief.py
+++ b/tests/unit/test_cmd_resume_brief.py
@@ -1,0 +1,81 @@
+"""CLI-level tests for `tj resume-brief --from-hook`.
+
+The SessionStart-hook path reads stdin JSON (session_id / transcript_path per
+Claude Code's contract) instead of the global mtime scan, so a concurrent
+session in a different project can't cross-leak its brief. Uses Click's
+CliRunner (canonical pattern from test_cmd_hook.py).
+"""
+from __future__ import annotations
+
+import json
+from unittest.mock import patch
+
+import pytest
+from click.testing import CliRunner
+
+from tokenjam.cli.main import cli
+from tokenjam.core.config import TjConfig
+
+
+@pytest.fixture
+def runner():
+    return CliRunner()
+
+
+def _config(tmp_path) -> TjConfig:
+    cfg = TjConfig(version="1")
+    cfg.storage.path = str(tmp_path / "tj.duckdb")
+    return cfg
+
+
+def _invoke(runner, cfg, args, stdin=""):
+    with patch("tokenjam.cli.main.load_config", return_value=cfg):
+        return runner.invoke(cli, args, input=stdin)
+
+
+def test_from_hook_uses_stdin_transcript_path(runner, tmp_path):
+    """--from-hook prefers stdin transcript_path over mtime scan."""
+    # A hand-built transcript file with a plausible session id
+    projects = tmp_path / "projects" / "proj-a"
+    projects.mkdir(parents=True)
+    tpath = projects / "sess-abc.jsonl"
+    tpath.write_text(json.dumps({"type": "user", "message": {"content": "Do the thing"}}) + "\n")
+
+    payload = json.dumps({
+        "session_id": "sess-abc",
+        "transcript_path": str(tpath),
+        "source": "resume",
+    })
+    result = _invoke(runner, _config(tmp_path), ["resume-brief", "--from-hook"], stdin=payload)
+    assert result.exit_code == 0
+    # We don't assert on brief content (that's the synthesizer's contract) —
+    # only that the command completed cleanly on a real transcript path.
+
+
+def test_from_hook_with_empty_stdin_exits_silently(runner, tmp_path):
+    """No usable signal on stdin → exit 0 with no output (never guess)."""
+    result = _invoke(runner, _config(tmp_path), ["resume-brief", "--from-hook"], stdin="")
+    assert result.exit_code == 0
+    assert result.output == ""
+
+
+def test_from_hook_with_malformed_json_exits_silently(runner, tmp_path):
+    """Malformed stdin JSON must not raise; degrade to no-op."""
+    result = _invoke(runner, _config(tmp_path), ["resume-brief", "--from-hook"], stdin="{not json")
+    assert result.exit_code == 0
+    assert result.output == ""
+
+
+def test_from_hook_with_stdin_missing_keys_exits_silently(runner, tmp_path):
+    """Valid JSON but no session_id / transcript_path → no-op (no cross-leak)."""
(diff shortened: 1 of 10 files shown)


anilmurty:
Reviewed — this is a strong one, and it's squarely the prevention side of the reuse/fan-out waste the `reuse` analyzer detects. One blocking fix and a few smaller items, then it's ready.

**Blocking: `--last` guesses the session by global mtime instead of reading the hook's stdin.** Claude Code passes JSON on stdin to SessionStart hooks (`session_id`, `transcript_path`, `source`), but `cmd_resume_brief --last` picks `max(mtime)` across ALL projects' transcripts. With any concurrent session — exactly the fan-out scenario this feature is for — the hook can inject the wrong session's brief, from a different project's private work. Fix: hook path reads stdin JSON (prefer `transcript_path`/`session_id`), keep `--last` as the manual fallback. That also fixes the resume case where the new session's own fresh transcript is the newest file, and it opens the door to per-subagent briefs later.

Smaller items (non-blocking, your call what lands in this PR):
- `resume-brief` isn't in `no_db_commands`, so with the daemon running `open_db` falls back to ApiBackend (no `.conn`) and the durable-snapshot path silently no-ops — the snapshot read only works when the daemon is DOWN, which inverts the PR body's description. Fail-soft holds, so fine to ship with a noted limitation + follow-up issue for an API endpoint (I'll file it).
- `test_malformed_inputs_never_raise` has `assert ... != "" or True` — the `or True` makes it vacuous; drop it or assert no-raise.
- `select_task_prompt`'s final ternary has identical branches.
- "ticket #67" auto-links to an unrelated docs item in this repo — clarify the reference.

On the headline: the −41%/−61% measurement is real and the body states its bounds honestly — good. One rule for when this graduates to README/docs/release notes: keep the framing "measured on two continuation fixtures, n=4/arm" (or land the eval harness) — the PR-title strength shouldn't appear in docs without the reproducibility behind it.

Fix the stdin item and I'm ready to merge.

anilmurty (changes requested):
noted in previous comment

anshss:
Pushed **eb0ce99** addressing this review.

**Blocking (anilmurty) — fixed**
- Added `--from-hook`: reads Claude Code's SessionStart stdin JSON (`session_id` / `transcript_path` / `source`) and briefs that session. Rewired the SessionStart hook to `tj resume-brief --from-hook` (idempotent update replaces the stale `--last` wiring in place via the existing `resume-brief` marker). `--last` preserved as the manual operator fallback. Empty / malformed / missing-keys stdin → silent exit 0 (never guess).

**Correctness (Greptile P1 / anilmurty) — fixed**
- `select_task_prompt`: else branch now returns `""` so `_is_substantive_prompt` actually gates the story-task fallback. Added `test_story_fallback_skips_slash_only_task`.

**Smaller items — fixed**
- `test_malformed_inputs_never_raise`: dropped the vacuous `or True`; now asserts return type + no-raise.
- Onboard "skipped" message reflects the real cause (malformed `settings.json`), not "custom hooks present" (custom hooks are preserved, they'd land in `written`).
- Promoted `_read_records` → public `read_records` in `core/transcript.py`; updated import in `cmd_resume_brief.py`.
- Stray "ticket #67" comment in `test_resume_brief.py` cleaned up.

**Deferred with follow-up (anilmurty's own call)**
- `resume-brief` in `no_db_commands` — leaving as-is; fail-soft snapshot path already degrades when the daemon holds the DB. Awaiting the API-endpoint follow-up.

All 1467 unit tests pass locally. CI running.

anilmurty (approved):
Re-reviewed the fix — the blocker's fully closed. The new `--from-hook` flag makes stdin authoritative (reads transcript_path/session_id off the SessionStart payload, returns early if there's nothing usable), so it can never fall through to the global-mtime scan — the cross-project wrong-session issue is gone at the root. All four nits addressed, load-bearing test added, CI green. Shipping default-ON: it's additive and fail-soft, so the worst case is a few hundred wasted tokens, not broken behavior. If you can land the eval harness as a fast-follow to grow the n, that'd be great. Nice work @anshss — this is the prevention side of the reuse-at-fan-out story. Merging.

PR #409, opened 2026-07-07
fix(ingest): attach session-less OTLP spans to trace's session

### LINKED ISSUE #326: [ingest] Raw-OTLP cost spans missing session.id mint a throwaway UUID instead of attaching to their trace's session
## Problem
On the raw-OTLP ingest path (`/api/v1/spans`), `session.id` is optional on a span.
A fan-out harness emits the zero-cost `invoke_agent` marker span **with** a `session.id`,
but the cost-bearing `gen_ai.llm.call` spans carry only an `agent_id` and a shared
`trace_id` (no `session.id`).

`_resolve_session` then mints a fresh throwaway session UUID for each session-less cost
span, so they never accumulate onto the marker span's session row. The denormalized
per-session token/cost columns in the `sessions` aggregate stay `0` for these sessions,
even though the cost data is present.

## Where
- `otlp_parsing.py` → `_resolve_session`
- `core/ingest.py`

## Proposed fix
Before minting a new UUID for a span with no `session.id`, look up an existing session via
the span's `trace_id` (a session-bearing span on the same trace) and attach to it — so the
`sessions` aggregate self-heals at ingest time.

## Done when
A raw-OTLP cost span sharing a `trace_id` with a session-marker span accumulates onto that
session at ingest, and the denormalized `sessions` token/cost columns match the values from
the read-time `session_token_cost_rollup`.

tarun73:
## Summary

This PR addresses the issue where raw-OTLP cost spans that lack a `session.id` but share a `trace_id` with an `invoke_agent` marker span were incorrectly assigned throwaway session UUIDs. The `_resolve_session`
method now performs a trace ID lookup, successfully attaching these session-less cost spans to the existing marker's session.

## Related issue

Closes #326

## Checklist

@anilmurty can you please review


anilmurty (approved):
Thanks @tarun73 — clean fix, and it lines up exactly with #326. Resolving via the trace_id sibling before minting a UUID is the right call, and I appreciate that you added it to the protocol, both backends, and the parity test rather than just the DuckDB path.

I especially like that `test_cost_span_before_marker_gets_own_session` documents the ordering behavior honestly instead of papering over it — the read-time `session_token_cost_rollup` stays authoritative for that case, so no cost data is actually lost.

One tiny non-blocking follow-up (separate PR is fine, don't hold this one on it): the "Session Continuity" section of CLAUDE.md still only mentions `conversation_id`-based attribution — worth a one-liner now that trace_id-sibling attribution exists.

Approving. Thanks again for the contribution.

PR #410, opened 2026-07-07
[feature] Onboarding first-signal verification: tj onboard --verify, tj ping, doctor silent-case check (#80)

### LINKED ISSUE #417: Onboarding has no "did it work?" verification step — silent misconfiguration is invisible
Two onboarding personas can finish `tj onboard` "successfully" yet never emit a single span, with no signal that anything is wrong:

- **SDK**: the SDK is deliberately fail-open. A typo'd `agent_id`, a missing `patch_anthropic()`/`@watch()` call, or a dead daemon drops spans silently — onboarding completes, nothing tells the user telemetry isn't flowing.
- **Claude Code / Codex**: these only start sending OTLP telemetry after a restart. If the user never restarts, ingest never begins, and there's no first-signal check to catch it.

In both cases the failure is invisible until someone notices `tj status`/the dashboard is empty much later, with no clue why.

## What's needed
A concrete "did it work?" verification step:
- A way to emit a labeled test span through the real export path and confirm it was actually delivered (not just attempted) — proving SDK wiring without running a full agent.
- A post-setup check that polls for the first real span and reports confirmed/not-confirmed with the likely per-persona cause.
- `tj doctor` should be able to diagnose the onboarded-but-zero-spans case after the fact.

Filed to correct PR #410, which previously referenced this gap as "#80" (a stale/incorrect reference — the real #80 was `tj tokenmaxx`, already shipped).

anshss:
Two onboarding personas could finish setup "successfully" yet never emit a single span, with no signal that anything was wrong. This adds a concrete "did it work?" step to both paths plus an after-the-fact doctor diagnosis.

## Summary
- **`tj ping`** — emits one clearly-labeled test span through the real `@watch()`/`record_llm_call` export path, with a local proof-of-life exporter that confirms interception even when the daemon is down, then reports where the span was delivered (HTTP daemon / local DuckDB) and exits non-zero if it couldn't be stored.
- **`tj onboard --verify`** — after setup (SDK, `--claude-code`, `--codex`), polls for the first span and reports confirmed / not-confirmed with the likely per-persona cause. Interactive runs without the flag offer the same via a prompt.
- **`tj doctor` "Onboarding signal" check** — diagnoses the onboarded-but-zero-spans case with the per-persona cause.
- New shared primitive `tokenjam/core/onboard_verify.py` (`poll_for_first_span`, `open_read_backend`, `not_confirmed_cause`) + SDK delivery-mode tracking (`bootstrap.get_mode()`).

## The two silent personas (#80)
- **SDK** is deliberately fail-open: a typo'd `agent_id`, a missing `patch_anthropic()`, or a dead daemon drops spans silently. `tj ping` gives a self-contained "the patch IS intercepting" proof without needing to build a real agent (à la Sentry's wizard test event / OTel console-exporter proof-of-life).
- **Claude Code / Codex** only start telemetry after a restart; if the user never restarts, ingest never begins. The verify poll waits for that first post-restart span; doctor flags the silent state after the fact.

## DuckDB single-writer safety
`onboard` is a `no_db_command` (no injected DB), so the verify runner resolves its own read path via `open_read_backend`, which tries the running daemon **first** (a plain HTTP GET, no lock) and only opens the file directly when no daemon answers. The poller never competes with `tj serve` for the write lock. `tj ping` is likewise a `no_db_command` and lets the SDK resolve its own HTTP-vs-direct path.

## Design note: doctor check is info, not warning
The "Onboarding signal" check reports at **info**, not warning. Doctor has no onboarding timestamp, so it can't distinguish "just onboarded seconds ago" (expected empty) from "onboarded long ago and silent" (the failure). A warning would false-alarm every fresh install and flip a clean `tj doctor` to exit 1 (breaking the existing exit-0-when-clean contract). The info line still carries the actionable per-persona cause, and the check goes green (ok) as soon as any span exists. It complements the existing "Live-span freshness" check, which owns the spans-exist-but-stale case.

## Tests / Verification
- `tests/unit/test_onboard_verify.py` — poll primitive (immediate/delayed confirm, clean timeout, read-error surfaced, agent_id passthrough), per-persona cause strings, `open_read_backend` daemon-first / direct-fallback / locked-error resolution.
- `tests/unit/test_ping.py` — end-to-end ping against a live in-process TracerProvider: interception proof, http/direct/failed delivery reporting, exit codes, `--json`, custom `--agent`.
- `tests/unit/test_onboard_verify_cli.py` — the runner's confirmed/timeout/unavailable output + the prompt gate.
- `tests/unit/test_doctor_first_signal.py` + `tests/integration/test_cli.py` — persona detection, info-when-silent (exit 0 preserved), ok-once-a-span-exists, API-fallback skip.
- Full suite green: 2011 passed, 1 skipped. `ruff` + `mypy` clean on all touched files.
- Live-smoked: `tj ping` printed the interception proof and wrote to the local DB (exit 0); `tj onboard --verify` (no daemon) polled 60s and reported the SDK cause gracefully; `tj doctor` ran clean.

## What's NOT in this PR
- `cli/cmd_onboard.py` (1,893 lines) was **not** refactored — the verify wiring is additive per the ticket's "add minimally" note (a separate ticket splits that file).
- No pricing entry for the `tj-ping-test` model — ping intentionally emits a 1-in/1-out token span; it falls back to default rates (a harmless "No pricing data" notice). Adding a zero-cost pricing row for the test model is a possible follow-up.

Closes #417


anilmurty (commented):
Thanks @anshss, this is a genuinely thoughtful feature. The wired-and-receiving vs configured-but-silent distinction is exactly the onboarding gap that bites people, and I like that `poll_for_first_span` confirms actual arrival and takes no DB write lock. Good test coverage too.

Two things before I merge:

**1. Issue reference is off.** "Closes #80" points at the `tj tokenmaxx` command (already shipped), not this work. Can you repoint it at the right onboarding-verification issue, or file one? As-is this would close #80 by mistake.

**2. `tj ping` exit code.** The interception proof and the copy are honest. But in HTTP mode the exit is 0 based on `mode`, even though `force_flush()` is wrapped in a swallowing try/except, so a silently-failed delivery still exits 0. The docstring says scripts/CI can gate on it, so could you either confirm delivery or note that exit-0 in HTTP means "attempted", not "confirmed"? (`--verify`'s poll is the real confirmation, so this is just tightening `ping`.)

Non-blocking nits Greptile also caught: `since` is typed `datetime` but the tests pass `None` (want `datetime | None`), and `_exit_for_mode` leans on `ctx.exit()` raising (works, just fragile style).

Fix 1 and 2 and it's good to go. Nice work.

anshss:
Thanks for the review — both addressed in 16253d3.

**1. Issue reference.** Filed #417 to describe the actual gap this PR closes (no "did it work?" verification step, both silent-failure personas). PR body updated to `Closes #417`.

**2. `tj ping` exit code.** You were right that `force_flush()` succeeding didn't mean the daemon actually stored the span — `TjHttpExporter.export()` swallows network/auth failures into a logged `SpanExportResult.FAILURE` instead of raising. HTTP mode now polls the daemon's read API for the just-emitted span (reusing `onboard_verify`'s `open_read_backend`/`poll_for_first_span`, ~8s timeout) before exiting 0; on timeout it exits non-zero with "span emitted but not confirmed received — is `tj serve` healthy?" Direct-mode's synchronous write is unchanged — re-querying there would contend with this process's own write connection for no real benefit. Extended `tests/unit/test_ping.py` to cover confirmed / unconfirmed / confirm-error cases.

Also took both nits: `since` is now `datetime | None`, and `_exit_for_mode`/`_exit_for_confirmation` is a single `ctx.exit(0 if confirmed else 1)` instead of relying on `ctx.exit()` raising.

Full suite green (2015 passed, 1 skipped), ruff + mypy clean on touched files.

anilmurty (approved):
Thanks @anshss — verified the fixes landed: `tj ping` now confirms delivery via `_confirm_delivery` / `_exit_for_confirmation` in HTTP mode (no more exit-0-without-delivery), the `since` typing is fixed, and the issue reference is corrected to #417. CI green across 3.10/3.11/3.12 + test-ts. Approving.

Issue #211, opened 2026-06-22
[Lens viz] Cost-by-component + recoverable-waste overlay (differentiator)


anilmurty:
The differentiated viz that depends on our optimize analyzers' attributed waste — so it is NOT subsumed by the Analytics explorer.

## What
Split the window's cost into components (input / output / cache-read / cache-write) as a stacked bar, and overlay each analyzer's **attributed recoverable waste** (downsize / cache / script / trim / reuse) per component.

## Why (cost value)
We have the analyzers to attribute waste by component; a generic activity dashboard can't. This is the cost-WASTE viz that makes Lens distinctively a cost-optimization tool.

## Data / screen
Optimize (or Overview) screen. Per-span token-component costs + the 5 analyzers' `estimated_recoverable_*` outputs.

## Constraints
uPlot stacked bars (cheap). Plan-tier framing (subscription → token-share; "estimated recoverable" never "saved" — Critical Rule 14).

## Acceptance
- Stacked cost-by-component chart with the recoverable-waste overlay, registry-driven (new analyzers appear automatically).
- Endpoint surfaces the component split + per-analyzer recoverable; tests; framing + caveat language verified.

Effort: MEDIUM. Standalone regardless of build path.

PR #423, opened 2026-07-08
[feature] Add summarize analyzer (Track A recoverable signal)

dankellysh:
This PR registers a new summarize optimize analyzer that surfaces static
prompt files worth summarizing and reports the per-call prompt-token reduction
available. It carries the #111 recoverable-savings contract, so the Overview
waste band and the /cost/components overlay pick it up registry-driven — no UI
change in this PR.

This is the read-only cost signal for the shipped tj summarize feature
(#354–357): the analyzer runs the same catalog-default, read-only scan
(core/summarize/candidates.list_candidates) the CLI/MCP already use, and sums
each file's est_tokens_saved.

How
- New core/optimize/analyzers/summarize.py: @register("summarize") ->
  SummarizeFinding (+ SummarizeCandidate), on the recoverable contract.
- runner.py: append "summarize" to ANALYZER_ORDER and add its from_dict
  constructor.
- Tests: sum / drop-zero-savings / empty scan / scan-error-swallowed /
  round-trip.

Notes for review
1. New side-effect class in the registry: unlike the other analyzers (which
   read ctx.conn telemetry), this one reasons over the filesystem. It runs the
   same read-only scan the tj summarize CLI uses, but at catalog-default scope
   only (a handful of known prompt files/locations), so it's cheap for the
   polling Overview. The scan function also supports repo/recursive discovery
   and re-scan, but this analyzer deliberately doesn't pass those (see What's
   NOT in this PR). It's try/except-guarded — a filesystem hiccup returns an
   empty finding, never breaks the optimize report.
2. Scan cwd: the scan reflects the process cwd + ~/ globals. Correct for the
   tj optimize CLI; under the long-running daemon it reflects the daemon's cwd
   (a known limitation, flagged not fixed here). For v1, non-catalog locations
   are surfaced via the tj summarize CLI (which takes path/recursive/repo);
   wiring that breadth into the analyzer is deferred.

What's NOT in this PR
- No API routes, no UI — analyzer / cost-signal only.
- No outbound calls, no writes — the scan is read-only.
- No dollar/amortized figure (deferred until a telemetry-ranked path lands).
- Catalog-default scope only — repo/recursive discovery and re-scan exist in
  the scan/CLI but aren't wired into the analyzer (deferred).

The API (reads/apply), the outbound run surface, and the Lens UI follow as
separate PRs.

### DIFF
--- a/tests/unit/test_optimize_summarize.py
+++ b/tests/unit/test_optimize_summarize.py
@@ -1,0 +1,100 @@
+"""Track A summarize analyzer — filesystem-derived recoverable finding.
+
+The analyzer reasons over the summarize scan (filesystem), not telemetry, so the
+scan is monkeypatched to a controlled ScanResult. Asserts the #111 recoverable
+contract: tokens summed from candidates, usd deliberately None (tokens-only), an
+explicit basis, and a clean report_to_dict/report_from_dict round-trip.
+"""
+from __future__ import annotations
+
+from datetime import timedelta
+
+import pytest
+
+from tokenjam.core.config import TjConfig
+from tokenjam.core.db import InMemoryBackend
+from tokenjam.core.optimize import build_report
+from tokenjam.core.optimize.runner import report_from_dict, report_to_dict
+from tokenjam.core.summarize.candidates import Candidate, ScanResult
+from tokenjam.utils.time_parse import utcnow
+
+
+@pytest.fixture
+def db():
+    backend = InMemoryBackend()
+    yield backend
+    backend.close()
+
+
+def _window():
+    return utcnow() - timedelta(days=30), utcnow() + timedelta(hours=1)
+
+
+def _cand(path: str, saved: int, *, is_prompt: bool = True, scope: str = "repo") -> Candidate:
+    return Candidate(
+        path=path, prose_words=saved * 3, total_chars=saved * 12,
+        protected_blocks=0, est_tokens_saved=saved, pricing_mode="api",
+        scope=scope, is_prompt=is_prompt,
+    )
+
+
+def _patch_scan(monkeypatch, cands: list[Candidate]) -> None:
+    result = ScanResult(candidates=cands, root=".", recursive=False,
+                        globals_checked=0, walk_capped=False, note="")
+    monkeypatch.setattr(
+        "tokenjam.core.summarize.candidates.list_candidates",
+        lambda **kw: result,
+    )
+
+
+def _run(db) -> object:
+    since, until = _window()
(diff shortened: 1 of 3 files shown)


anilmurty:
Thanks @auspexlabs. Two things before this can merge, both quick:

1. **CI is red** — root cause: this is the first analyzer that reasons over the *filesystem* instead of the query window, so on an empty telemetry window it still scans the repo's CLAUDE.md/AGENTS.md and leaks into the `/cost/components` recoverable overlay, breaking `test_components_empty_window_is_safe` (a real invariant from #211). Fix: window-guard the analyzer — early-return an empty finding when the window has no qualifying telemetry. That's also more honest: a dead window shouldn't show a recoverable figure. Please don't edit the test to allow-list summarize.
2. **Missing the baked-in honesty caveat (Rule 14)** — every other savings finding carries its caveat as a dataclass default (e.g. `MODEL_DOWNGRADE_CAVEAT`), but `SummarizeFinding` has none, so the overlay emits an empty caveat. Add a `SUMMARIZE_HONESTY_CAVEAT` constant ("structure guaranteed; meaning may change") as the field default so it can't be dropped, and restore it in the round-trip constructor.

While you're in here: have the finding also **emit `reduction_pct` / `avg_reduction_pct`** — the UI in #426 is computing those in JS and should read them from the analyzer instead (see my #426 note).

dankellysh:
All three in e833f3d:

1. Window-guard (CI): the analyzer now early-returns an empty finding when the
   telemetry window has no tokens (ctx.summary.total_tokens == 0) — a dead window
   shows no recoverable figure, and the filesystem scan is skipped entirely.
   test_components_empty_window_is_safe passes; I didn't touch the test, and added
   a unit test asserting the scan doesn't run on a dead window.
2. Honesty caveat: added SUMMARIZE_HONESTY_CAVEAT ("Structure is guaranteed;
   meaning may change — review each rewrite before applying.") as the
   SummarizeFinding.caveat default (like MODEL_DOWNGRADE_CAVEAT) + restored in the
   round-trip constructor, so the overlay carries it.
3. reduction %: the finding emits reduction_pct (token-weighted aggregate) and
   avg_reduction_pct, and each candidate carries a per-file reduction_pct — all on
   the shared CHARS_PER_TOKEN, so #426 renders them instead of computing in JS.

anilmurty (approved):
Thanks @auspexlabs — verified: the window-guard clears the CI failure (empty window → empty finding), the honesty caveat is now a dataclass default (can't be dropped), and reduction_pct is computed server-side. CI green. Merging.

PR #424, opened 2026-07-08
[feature] Add summarize API surface (reads + guarded apply/undo)

dankellysh:
Adds the FastAPI surface the Lens Summarize screen reads from and applies
through — the in-process bridge to the shipped tj summarize core (#354–357).
Every route calls core directly (like /optimize → build_report); nothing shells
out to tj, and there are no outbound LLM calls in this PR.

Routes (all /api/v1, Depends(require_api_key), sync def so filesystem/DB work
runs in Starlette's threadpool):
- GET /summarize/capabilities — which run engines work on THIS host (UI greys
  out dead paths).
- GET /summarize/candidates — the read-only scan (list_candidates).
- GET /summarize/staged — staged rewrite records (all, or one by path).
- GET /summarize/backups — applied files that still have a backup (the Undo
  surface), each with a computed undoable flag + reason.
- POST /summarize/apply — apply staged rewrite(s) via core apply_staged.
- POST /summarize/undo — restore from backup via core undo.

How
- New api/routes/summarize.py + one include line in api/app.py.
- core/summarize/backup.py: add list_backups() (the undo surface) and record
  est_tokens_saved in the backup meta, so the Optimize box can report applied
  savings (the read-only scan drops applied files, so their saving is otherwise
  lost). Backward-compatible — older backups without the field read as 0.
- core/summarize/apply.py: pass est_tokens_saved through to backup.save on apply.
- Tests: 10 integration tests; core is monkeypatched (no real writes/outbound)
  to prove the routes wire to core and shape the payloads.

Notes for review
1. Lens's first file-writing action. apply/undo. Both call core (apply_staged /
   undo), never re-implement the guards, so every write keeps the shipped
   guarantees: owner-check + content-hash drift-refuse + symlink-refuse + gzip
   backup. Default is DRY-RUN (go=false, returns the plan, writes nothing); the
   UI sends go=true only on an explicit per-file Apply.
2. Backup meta gains a field. backup.save now records est_tokens_saved in
   *.meta.json (additive; readers default to 0). list_backups recomputes
   undoable from the live file (exists / not-symlink / content still matches what
   apply wrote) — the same conditions undo enforces, surfaced up front so the UI
   can explain why a row can't be undone (e.g. "changed since apply — undo would
   lose newer edits"; also file-missing / symlink / unreadable).
3. These handlers do blocking filesystem/DB I/O, so they're plain sync def —
   FastAPI offloads sync routes to a threadpool, keeping the event loop free.

What's NOT in this PR
- No outbound LLM calls. The run/prep/check endpoints (which make tj serve an
  outbound caller for the first time) are a separate, design-gated PR.
- No UI — API only.
- No new config, no telemetry/spans.

The analyzer (cost signal) is its own PR; the outbound run surface and the Lens
UI follow.

### DIFF
--- a/tests/integration/test_summarize_api.py
+++ b/tests/integration/test_summarize_api.py
@@ -3,10 +3,14 @@ staged / backups + apply / undo.
 
 Core scan + staging + apply are exercised by their own unit tests; here we prove
 the routes wire to core, shape the payloads, and (capabilities) reflect the host
-honestly. list_candidates / session / apply / backup are monkeypatched for
-determinism — no real writes.
+honestly. Most tests monkeypatch list_candidates / session / apply for
+determinism; the real-core apply tests at the end drive `POST /apply` end-to-end
+against a temp file (no mocks) to close that seam — the drift/owner/symlink/backup
+guards must hold through the route, not just in the unit tests.
 """
 from __future__ import annotations
+
+import re
 
 import httpx
 import pytest
@@ -22,6 +26,9 @@ from tokenjam.core.config import (
 from tokenjam.core.db import InMemoryBackend
 from tokenjam.core.ingest import IngestPipeline
 from tokenjam.core.summarize.candidates import Candidate, ScanResult
+
+PROSE = "Always act carefully and never drop a required step when you respond. " * 30
+_MARKER_RE = re.compile(r'<tj-keep id="\d+"[^>]*?(?:/>|>.*?</tj-keep>)', re.DOTALL)
 
 
 @pytest.fixture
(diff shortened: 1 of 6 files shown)


anilmurty:
Thanks @auspexlabs — this one's clean, no changes needed. I went through the apply/undo guards specifically: dry-run is the default, the write path enforces owner + content-hash-drift + symlink checks with a gzip backup before any overwrite (skips-with-reason, never forces), and undo refuses on drift with a 409. That's exactly the posture I want for anything that touches a user's prompt files.

Ready on its own — it'll land as part of the stack (order noted in #426). One optional fast-follow, not a blocker: an integration test driving `POST /apply {go:true}` through the route against a real temp file, asserting a symlink/drift is skipped — closes the seam you flagged where the route tests monkeypatch core out.

dankellysh:
Appreciate the review! Added the optional fast-follow you mentioned in d46b5ab:
a real integration test driving POST /apply {go:true} against a temp file
(no mocks) — happy-path (file rewritten + gzip backup + staging cleared) and
drift (file edited after staging → skipped "changed since check", never
overwritten). Closes the seam where the route tests monkeypatched core out.

anilmurty (approved):
Thanks @auspexlabs — the real-core write-path integration test closes the seam (drift/owner/symlink/backup all exercised through the route), plus the backup-preflight hardening. Merging.

anilmurty (approved):
Thanks @auspexlabs — the real-core write-path integration test closes the seam (drift/owner/symlink/backup through the route) and the backup-preflight hardening is in. Merging.

dankellysh:
> Thanks @auspexlabs — the real-core write-path integration test closes the seam (drift/owner/symlink/backup through the route) and the backup-preflight hardening is in. Merging.

Very welcome!

PR #429, opened 2026-07-08
fix(summarize): guard file reads against non-UTF-8 + correct /check missing-file semantics

dankellysh:
Follow-up to the summarize stack (#423–#426). Greptile's post-merge review flagged two real exception-handling bugs on `main`, both the same class — a too-narrow / wrong-exception catch. This fixes the whole class across the module (every `read_text` / `read_bytes` / `json.loads` in `backup.py` + `session.py`).

**1. Non-UTF-8 / corrupt crashes `/summarize/backups` and `/undo` (from #424).** `list_backups`, `recorded_output`, and `load_original` (used by `undo`) read via `read_text(encoding="utf-8")` / `gzip.decompress().decode()` but caught too narrow — `UnicodeDecodeError` is a `ValueError` (not `OSError`), and `load_original` also had an unguarded `KeyError` on the meta — so one non-UTF-8 / truncated / corrupt applied file or backup 500s the response instead of a clean `undoable=false` / 409. Broadened every backup read to catch `(OSError, UnicodeDecodeError, json.JSONDecodeError, KeyError)`.

**2. `/summarize/check` `FileNotFoundError→404` was dead code (from #425).** `check()` guards missing files with `is_file → SummarizeRefused`, so a missing file is **409** (its documented "vanished since prep" contract), never `FileNotFoundError`. The route's 404 catch never fired, and its test *mocked* `check` to raise `FileNotFoundError` — green while asserting an impossible path. Removed the dead catch.

**3. Same non-UTF-8 gap in the check/prepare reads.** Hardened `session.check` and `_read` (used by `prepare`) so `UnicodeDecodeError → SummarizeRefused` (→ 409); `check` also maps its `FileNotFoundError` race (vanish between `is_file` and the read) → `SummarizeRefused`. `prepare` still raises `FileNotFoundError` for a genuinely missing file, so `/prep` and `/run` stay **404** for missing — only the encoding case is remapped.

**4. Staged-record readers hardened.** `session.list_staged` / `read_staged` now catch `(OSError, UnicodeDecodeError, json.JSONDecodeError)`, so a corrupt / unreadable / non-UTF-8 staged record is skipped (or returns `None`) instead of crashing `/summarize/staged`.

**Why 409 (not 404) for `/check` on a missing file:** `check` is the second phase (after `prep`); a file gone by then is a conflict with the prepped state, which `check` already models as `SummarizeRefused` (409). The real goal — no 500 — is met; literal 404-parity with `/prep`/`/run` doesn't fit, since those are first-read ops.

**Also:** updated the `SummarizeRefused` docstring, which previously undersold how broadly the exception is now used.

**Tests** are real-input (real missing file, real non-UTF-8 bytes, corrupt staged record, corrupt backup meta/blob), not mocked exceptions.

### DIFF
--- a/tests/integration/test_summarize_run_api.py
+++ b/tests/integration/test_summarize_run_api.py
@@ -189,14 +189,41 @@ async def test_run_refused_when_outbound_disabled(db, config):
     assert "allow_outbound_run" in r.json()["detail"]
 
 
-async def test_check_missing_file_returns_404(client, monkeypatch):
-    # File vanished between prep and check → 404, parity with /prep and /run.
-    def gone(config, path, summary, source_hash, **kw):
-        raise FileNotFoundError(path)
-    monkeypatch.setattr("tokenjam.core.summarize.session.check", gone)
+async def test_check_missing_file_returns_409_real(client, tmp_path):
+    # check() guards missing files with is_file → SummarizeRefused, so a REAL missing
+    # file is 409 (its documented "vanished since prep" contract) — not the 404 an
+    # earlier mocked test wrongly asserted. (Contrast /prep + /run below, which DO 404
+    # via prepare's FileNotFoundError — verified per-function, not assumed.)
     r = await client.post("/api/v1/summarize/check",
-                          json={"path": "./gone.md", "summary": "s", "source_hash": "abc"})
-    assert r.status_code == 404
+                          json={"path": str(tmp_path / "gone.md"), "summary": "s", "source_hash": "abc"})
+    assert r.status_code == 409
+
+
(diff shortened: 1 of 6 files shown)


anilmurty (approved):
Thanks @auspexlabs — clean fix for the whole non-UTF-8/corrupt-read class (UnicodeDecodeError is a ValueError, not OSError — good catch), and you correctly found my #425 `/check`→404 ask was dead code: check() already models a missing file as SummarizeRefused/409 ("vanished since prep"), so 409 is right and 404 is for the first-read ops. The real-input tests (actual non-UTF-8 bytes, corrupt gzip/meta) are exactly right. Merging.

PR #435, opened 2026-07-09
Fix #365: decompress gzip OTLP bodies on ingest

### LINKED ISSUE #365: Ingest rejects gzip-compressed OTLP bodies with a bare 400
## Problem
OTLP/HTTP exporters (the OTel Collector's `otlphttp` exporter and most SDK exporters) gzip the request body **by default**. TokenJam's ingest returns `HTTP 400` on every gzip'd request (collector logs `responded with HTTP Status Code 400 … Dropping data`). A real user has to know to set `compression: none`, which they won't — so TokenJam is incompatible with default-configured OTLP clients out of the box. Found wiring a standard OTel Collector into `tj serve`.

## Repro
OTel Collector `otlphttp` exporter without `compression: none` → 400.

## Fix
Honor `Content-Encoding: gzip` on the ingest endpoint (decompress before parsing).

Related DX: the total-failure 400 currently returns no reason — worth returning a small JSON error (`{"error": "could not parse OTLP body: <reason>"}`) while keeping the partial-success = 200-with-counts contract. Best fixed alongside the native `/v1/traces` route (companion issue) in one PR.

tarun73:
OTLP/HTTP exporters (OTel Collector `otlphttp`, most SDK exporters) gzip request bodies by default. The ingest endpoint called `request.json()` directly which doesn't handle `Content-Encoding: gzip`, causing a 400 on every default-configured client connection — forcing users to set `compression: none` manually, which they won't know to do.

## Summary
- Add `_read_otlp_body()` helper that transparently decompresses gzip before JSON parsing
- Apply to `POST /api/v1/spans` and `POST /v1/logs`; `POST /v1/traces` inherits automatically
- Replace the bare `"Invalid JSON body"` 400 with `{"error": "could not parse OTLP body: <reason>"}` so the failure reason surfaces in collector logs
- 7 new integration tests; all 2024 existing tests still pass

## Gzip decompression — two paths

1. **`Content-Encoding: gzip` header present** — spec-compliant path, handles OTel Collector default
2. **Gzip magic bytes `\x1f\x8b` without the header** — sniff fallback for exporters that compress but omit the header (goes slightly beyond the issue, covers misconfigured clients)

## Tests / Verification
- 7 new integration tests in `tests/integration/test_api.py` (header path, sniff path, corrupt gzip × 2, plain JSON regression, `/v1/traces`, `/v1/logs`)
- Manually verified against a live `tj serve` — confirmed 400s on `main`, all pass on this branch

## What's NOT in this PR
- `x-gzip` alias support — not used by any known OTel exporter, deferred

## Related issue
Closes #365 

## Checklist

- [X] CLAUDE.md updated (if architecture changed)
- [X] Test spans use `tests/factories.py` (not raw `NormalizedSpan`)

@anilmurty  please review

** Let me know if the Gzip magic bytes is a stretch and not needed here. will remove it

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -2,6 +2,8 @@
 from __future__ import annotations
 
 import asyncio
+import gzip
+import json
 
 import pytest
 import httpx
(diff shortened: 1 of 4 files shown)


anilmurty (approved):
Thanks @tarun73 — clean fix for a real out-of-the-box breakage (default OTLP clients gzip bodies, so a stock OTel Collector got a 400 until you knew to set compression:none). Verified the whole thing: single-sourced `read_otlp_body` in `_body.py` used by spans + otlp (traces/logs inherit), the gzip-bomb guard (32MB decompress cap), `-> Any` return type, and the public helper name — so all three of Greptile's flags (P1 gzip-bomb + the two P2s) are already addressed in the current head. 7 good integration tests, CI green. Merging.

PR #439, opened 2026-07-09
fix(uninstall): remove claude() shell wrapper + unwire output_cap hook (#117)

anshss:
## Summary

`tj onboard --claude-code` writes TWO things into `~/.zshrc`: (1) a `# tj harness observability` OTEL export block, and (2) a `# tj per-terminal naming` … `# end tj per-terminal naming` block defining a `claude()` shell function that calls `tj otel-resource-attrs` and `tj session-end`. `tj uninstall` removed block (1) but had NO code path removing block (2) — so after `tj uninstall` + `pip uninstall tokenjam`, the leftover `claude()` function errors (`tj: command not found`) on every `claude` launch.

- Added `_unwire_claude_wrapper()` in `cmd_onboard.py` — the counterpart to `_install_claude_wrapper()`. Removes the `# tj per-terminal naming` … `# end tj per-terminal naming` block from `~/.zshrc` / `~/.bashrc` via exact marker-delimited removal (not a fragile regex on the block body), idempotent, preserves everything else.
- Wired `_unwire_claude_wrapper()` into `cmd_uninstall.py`.
- Also wired in the existing-but-never-called `_unwire_claude_output_cap_hook()` so an opted-in PostToolUse cap hook is cleaned up on uninstall too.
- Left the OTEL-export block's marker-matching logic untouched (owned by #118).

## Test plan

- [x] New unit tests in `tests/unit/test_uninstall_wrapper.py`: `_unwire_claude_wrapper()` removes the block written by `_install_claude_wrapper()`, cleans `.bashrc` when present, is idempotent/no-op when absent, preserves surrounding rc content, and an end-to-end `cmd_uninstall` CLI test (fixture `~/.zshrc` with both onboard-written blocks + a wired cap-output hook) asserting no `tj ` references or `claude()` function remain after uninstall.
- [x] `python3 -m pytest tests/unit/ -k "uninstall or onboard or wrapper" -q` → `144 passed, 1498 deselected`
- [x] `python3 -m pytest tests/unit/ -q` → `1641 passed, 1 skipped`
- [x] `python3 -m pytest tests/integration/test_cli_uninstall.py tests/integration/test_cli.py -q` → `85 passed`
- [x] `ruff check` clean on changed files

**Merge order**: Part of the CC install/remove cluster — merge #117 → #118 → #121; this is #117 (foundation, merge first).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #440, opened 2026-07-09
feat(uninstall): consented --purge package removal + runner-aware hint (#121)

anshss:
## Summary
- Extends `tj uninstall`'s existing opt-in package removal (`--remove-package`, #430) with a `--purge` alias and full runner detection, making `npx tokenjam uninstall [--purge]` the complete, symmetric counterpart to `npx tokenjam onboard`.
- Adds `uv tool install` detection alongside the existing pipx detection — `--purge` now auto-runs `uv tool uninstall tokenjam` for uv-tool installs, and the default (non-purge) closing hint prints the exact matching command (`pipx uninstall` / `uv tool uninstall` / `pip uninstall`) instead of only ever guessing pipx-or-pip.
- Adds `_is_ephemeral_runner()` detection (verified empirically against real `uv`/`pipx` installs — see code comments for the exact paths) so an ephemeral invocation (`uvx`/`pipx run`, including via the `npx tokenjam` wrapper's runner fallback) makes `--purge` a safe no-op with an explanation, rather than guessing at a removal command for a package that was never persistently installed.
- Documents install/remove as a symmetric pair in `docs/claude-code-integration.md`.

Resolves #121.

## Test plan
- [x] `python -m pytest tests/unit/ -k "uninstall or purge or runner" -q` → 20 passed
- [x] `python -m pytest tests/unit/ tests/integration/ -q` (full suite) → 2048 passed, 1 skipped (pre-existing)
- [x] `ruff check` clean on all touched/new files
- [x] Manually exercised the CLI output for the ephemeral + `--purge` no-op path

**Merge order:** CC install/remove cluster — merge #117 → #118 → #121; this is #121, MERGE LAST (depends on #117 + #118 for a fully-clean uninstall).

### DIFF
--- a/docs/claude-code-integration.md
+++ b/docs/claude-code-integration.md
@@ -111,15 +111,15 @@ investigation.
 
 ## Uninstalling
 
-```bash
-# Remove all TokenJam data, config, daemon, MCP registration, and env vars from every onboarded project:
-tj uninstall --yes
+Set up and remove are a symmetric one-command pair:
 
-# Then remove the package itself (tj uninstall intentionally skips this):
-pip uninstall tokenjam -y
+```bash
+npx tokenjam onboard --claude-code   # one paste to set up
+npx tokenjam uninstall [--purge]     # one paste to remove
 ```
 
-`tj uninstall` cleans up everything set by `tj onboard --claude-code`:
+`npx tokenjam uninstall` (equivalently `tj uninstall` if you have a persistent install on PATH) cleans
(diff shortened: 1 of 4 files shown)


anshss:
Superseded by #442, which folds this branch together with the other two per-ticket PRs (#439 #441 #440) into a single PR to `main`, plus the Greptile review fixes. Closing in favor of #442. Branch retained for reference.

PR #449, opened 2026-07-09
refactor(cli): unify DB access behind a DataAccess seam; fix tokenmaxx/quota-audit under a running daemon

anshss:
Unifies the CLI's DuckDB-vs-daemon data access behind one explicit `DataAccess` interface, and fixes the three analytics commands that broke while `tj serve` was running — the daemon onboarding installs by default.

## Summary

- Introduce `cli/data_access.py`: one `DataAccess` interface with two implementations — `DirectDataAccess` (direct DuckDB connection) and `ServeDataAccess` (routed through a running `tj serve`, which owns the connection). A single choke point, `resolve_data_access(ctx)`, resolves direct-vs-serve, so commands never duck-type the backend with `hasattr(db, "conn")`.
- Migrate `tj context`, `tj tokenmaxx`, and `tj quota-audit` onto the seam.
- `tj tokenmaxx` now works with the daemon up by reusing the existing `/api/v1/context` endpoint (it shares `compute_context_diagnostic` with `tj context`) — no new endpoint.
- `tj quota-audit` now works with the daemon up via a new `GET /api/v1/quota-audit` endpoint plus a canonical `audit_to_dict` / `audit_from_dict` round-trip pair in `core/optimize/types.py`.

## The problem

`cli/main.py` opens a DuckDB backend and, when the DB write-lock is held by a running `tj serve`, falls back to a read-only HTTP shim (`ApiBackend`). Commands then branched on `hasattr(db, "conn")` (direct DuckDB) vs `hasattr(db, "fetch_*")` (shim) to pick a code path. `tj tokenmaxx` and `tj quota-audit` required a direct connection and raised "needs a direct database connection … stop the daemon" whenever the daemon was up — i.e. from the moment onboarding installs it, on two of the most Claude-Code-relevant commands. The lock was mitigated ad hoc in four places (`fetch_cost_compare`, `fetch_optimize_report`, `fetch_reuse_clusters`, `fetch_context_diagnostic`), each paired with its own `hasattr` branch; the two backends drifted silently, and tokenmaxx/quota-audit had no serve path at all.

## Design: why route through the daemon (not a read-only attach)

DuckDB permits only one writer OR many readers across processes. A concurrent read-only connection alongside the serve write-lock is impossible in this DuckDB version — I verified this empirically before committing to the design (a throwaway script: one process holds a read-write connection like `tj serve`; a second process attempts `duckdb.connect(path, read_only=True)`):

```
[main] writer pid=... holds RW lock on .../probe.duckdb
[main] RW connect FAILED (expected): IOException: IO Error: Could not set lock on file ...
[main] RO connect FAILED: IOException: IO Error: Could not set lock on file ...:
       Conflicting lock is held in .../Python.framework/...
RESULT: RO_ATTACH_BLOCKED
```

So the fix routes the compute through the daemon, which owns the direct connection — the same pattern `/api/v1/context` (#63), `/api/v1/reuse/clusters` (#154) and `/api/v1/optimize` (#68) already use.

## Empirical verification: all three commands succeed under a real running daemon

Started a **real `tj serve`** holding the write lock against a seeded temp DB, then ran the three commands through it (full script + output attached in the review notes). Abridged:

```
=== daemon up (pid ..., port ...), holding the DuckDB write lock ===
=== proof: a direct CLI open of the same DB is locked out ===
duckdb.duckdb.IOException: IO Error: Could not set lock on file ... (daemon holds it)

########## tj context --since 30d (daemon up) ##########      ---- exit code: 0 ----
########## tj tokenmaxx --since 30d (daemon up) ##########    ---- exit code: 0 ----
########## tj quota-audit --since 30d (daemon up) ##########  ---- exit code: 0 ----
```

Each rendered its full card (context composition, the QuotaSink efficiency card, the Opus quota audit with spot-check sessions), and `--json` works for all.

## Tests / Verification

- `tests/integration/test_tokenmaxx_serve.py` — tokenmaxx renders through a running daemon (CLI-through-shim over the in-process ASGI app), `--json`, and a clean error when neither a live conn nor a daemon is reachable.
- `tests/integration/test_quota_audit_serve.py` — the new `/api/v1/quota-audit` endpoint computes the audit server-side; CLI-through-shim; `--json`; clean error path.
- `tests/integration/test_data_access_parity.py` — a parity guard: `DirectDataAccess` and `ServeDataAccess` produce byte-identical serialized diagnostic/audit + framing over the same seeded DB (the silent-drift protection the old `hasattr` seam lacked).
- `tests/integration/test_context_serve.py` (existing) still passes after routing `tj context` through the seam.
- Full suite green: **2316 passed, 1 skipped**. `ruff check tokenjam/` and `mypy tokenjam/` both clean.

## What's NOT in this PR (deliberately scoped)

- The other `hasattr(db, "conn")` sites in `cmd_cost` / `cmd_optimize` / `cmd_status` etc. are left as-is: those already have working daemon-up fallbacks (`fetch_cost_compare` / `fetch_optimize_report`) and degrade gracefully rather than erroring. This PR targets the commands that were actually broken plus the seam that unifies them; folding the cost/optimize family into `DataAccess` is a natural follow-up but would touch dense, unrelated framing code and widen the diff.
- No read-only-attach path — empirically impossible in this DuckDB version (see above).


anshss:
Bundled into #451 (combined onboarding first-touch batch) — merging there as one; this branch's commits and review fixes are included verbatim.

PR #452, opened 2026-07-10
fix: surface progress during tj quickstart's ingest

anshss:
\`tj quickstart\` — the zero-install, zero-config \`npx tokenjam\` demo — goes silent for its entire ingest. On a large \`~/.claude\` history that's tens of seconds of dead cursor before any output: indistinguishable from a hang at exactly the moment a brand-new user's trust is most fragile.

## Summary
- Prints an honest status line before ingest starts: \`Reading your last 30 days of Claude Code history (most-recent 300 sessions)…\`
- Wires the existing streaming per-session progress counter (\`backfill_progress\`, shipped for \`tj onboard --claude-code\`'s backfill) into quickstart's ingest call — reused, not rebuilt.
- \`--json\` stays byte-for-byte clean on stdout: both the status line and the counter route to the stderr console when \`--json\` is set, instead of contaminating the machine-readable payload or going fully silent for scripted callers.

## Root cause
\`ingest_claude_code(db, root=root, since=since_dt, max_sessions=max_sessions)\` was called in \`cli/cmd_quickstart.py\` with no \`progress=\` callback and no output printed before it returned — the first \`console.print\` of anything happened only after the (potentially tens-of-seconds) ingest finished.

\`tj onboard --claude-code\` hit the identical problem on its own backfill and already solved it: a pre-backfill status line plus a shared streaming counter module (\`tokenjam/cli/backfill_progress.py\`) that renders a live Rich progress line on a TTY and degrades to periodic plain prints (every 100 sessions) off a TTY. This PR wires that same machinery into quickstart instead of reinventing it.

## What changed
- \`tokenjam/cli/backfill_progress.py\`: \`backfill_progress()\` gained an optional \`console\` override parameter (default unchanged — \`tj onboard\`'s call site is untouched). This lets a caller redirect the counter to a different console (quickstart needs stderr under \`--json\`) without forking the shared module.
- \`tokenjam/cli/cmd_quickstart.py\`: added \`_describe_window\`/\`_pre_ingest_status\` helpers for the honest pre-ingest line, and wired \`count_claude_code_sessions_in_scope\` + \`backfill_progress\` around the \`ingest_claude_code\` call, choosing the stdout or stderr console based on \`--json\`.

## Tests / Verification
- New unit tests in \`tests/unit/test_quickstart.py\`:
  - the pre-ingest status line prints before the quota-composition render, with the correct window/cap phrasing
  - the status line correctly omits the cap phrasing under \`--full\`
  - \`--json\`'s stdout parses as pure JSON with **zero** extraneous lines (\`result.stdout\`, not the mixed \`result.output\`), while the status line still shows up on \`result.stderr\`
  - a large (250-session) history shows the streaming counter advancing (\`Backfilling 100/250…\`, \`Backfilling 200/250…\`)
- \`make test\`: 2347 passed, 2 pre-existing skips
- \`ruff check tokenjam/\`: clean
- \`mypy tokenjam/\`: clean (190 source files)
- Manual verification against a synthetic 400-session history: the fast/default (non-\`--json\`) path shows the status line + advancing counter within ~1s of launch, through to the render. \`tj quickstart --json --root ... > out.json 2> err.log\` confirmed \`out.json\` is a single valid JSON line with zero progress contamination, while \`err.log\` carries the status line and counter.

## What's NOT in this PR
- No change to the underlying ingest performance itself — this is purely a feedback/UX fix (the ingest was already bounded by the existing 300-session first-run cap, #13).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,10 @@ classifiers = [
 ]
 
 dependencies = [
-    "click>=8.1",
+    # >=8.2: CliRunner separately captures stdout/stderr by default from this
+    # version on (result.stdout / result.stderr always populated); tests that
+    # assert stdout stays clean of stderr-routed output rely on this.
+    "click>=8.2",
     "rich>=13.0",
     "tomli>=2.0; python_version < '3.11'",
     "tomli-w>=1.0",
--- a/tests/unit/test_quickstart.py
+++ b/tests/unit/test_quickstart.py
@@ -250,6 +250,67 @@ def test_quickstart_no_logs_is_graceful(tmp_path):
     assert "No Claude Code logs" in result.output
 
 
+# ── Pre-ingest progress: ingest was previously the ONE silent stretch in the
+# whole command (~40s dead cursor on a large history, nothing printed until
+# after it returned). An honest status line now lands before ingest starts,
+# and the shared streaming counter (`backfill_progress`) advances per session
+# through to render. `--json` must stay byte-for-byte clean on stdout.
+
+def test_quickstart_prints_pre_ingest_status_before_render(tmp_path):
+    root = _fixture_root(tmp_path)
+    result = _invoke_quickstart(["--root", str(root), "--since", "90d"])
+
+    assert result.exit_code == 0, result.output
+    assert "Reading your last 90 days of Claude Code history" in result.output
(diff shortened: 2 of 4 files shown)


anshss:

PR #453, opened 2026-07-10
Shared premium-tier predicate: make Fable visible to quota audit, subagent right-sizing, and the downgrade map

anshss:
## Problem

Premium-tier awareness was hardcoded to an `"opus"` substring in three separate places, so **Fable — the model tier above Opus — was invisible** to every consumer:

- the **quota audit** skipped Fable sessions entirely (never counted, never flagged);
- **subagent right-sizing** never flagged a Fable-powered subagent doing trivial work as `over_powered`;
- the **downgrade map** had no Fable entry — and was independently **stale for Opus**, stopping at `claude-opus-4-7` with `claude-opus-4-8` missing.

## Fix

One shared tier predicate, consumed by all three sites — no more per-analyzer substrings.

- **New `core/model_tiers.py`** — `model_tier()` / `is_premium_tier()` classify a model id into a family tier via a lowercased-substring match that tolerates version suffixes, `YYYYMMDD` dates, `[1m]` context tags, and Bedrock provider prefixes. `PREMIUM_TIERS = {fable, opus}`; the next premium family is a one-line add here.
- **Quota audit** (`model_downgrade.py`) resolves premium membership through the shared predicate; Fable sessions are now counted in the denominator and flagged as reclaim candidates.
- **Subagent right-sizing** flags `over_powered` for Fable spawns, not just Opus.
- **`DOWNGRADE_CANDIDATES`** gains `claude-fable-5 → claude-sonnet-4-6` and the missing `claude-opus-4-8 → claude-haiku-4-5` (each drops two tiers — the aggressive step is justified because these are only proposed for structurally tiny, Sonnet-shaped sessions).
- **Copy/docs** name both tiers (Opus/Fable) in the audit render, the onboarding next-steps line, the serve error label, and the CLI reference.

The audit's `audit_opus_quota` name and the `opus_*` serialization fields are kept for API stability but now denote the whole premium tier; docstrings say so.

## Tests

Fixture-backed coverage across all three consumers:
- `test_model_tiers.py` — classification incl. dates / `[1m]` / Bedrock prefixes.
- quota audit counts + flags a Fable session **and** an Opus 4.8 session (mixed corpus).
- subagent analyzer flags an `over_powered` Fable spawn.
- downsize proposes candidates for Fable **and** Opus 4.8.

`make test` (2365 passed, 2 skipped), `make lint`, `make typecheck` all green.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -129,7 +129,7 @@ Key flags: `--session`, `--last` (default when `--session` is omitted), `--json`
 
 ### `tj quota-audit`
 
-Retroactive audit of your Opus quota: which past Opus sessions were structurally Sonnet-shaped (small input/output, few tool calls)? Reports the percent of Opus quota reclaimable, example sessions to spot-check, and an optional tuned routing-config export. Quota-share framing, never a dollar "saving" claim. Needs a direct DB connection (can't run against a live `tj serve`).
+Retroactive audit of your premium (Opus/Fable) quota: which past premium-tier sessions were structurally Sonnet-shaped (small input/output, few tool calls)? Covers Fable (the tier above Opus) as well as Opus. Reports the percent of premium quota reclaimable, example sessions to spot-check, and an optional tuned routing-config export. Quota-share framing, never a dollar "saving" claim. Needs a direct DB connection (can't run against a live `tj serve`).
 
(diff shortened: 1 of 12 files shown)


anshss:

PR #454, opened 2026-07-10
Reframe retrospective quota-audit headline: misallocated, persona-gated

anshss:
The retroactive premium-tier quota audit rendered "~N% of your Opus quota is reclaimable". But retrospective quota is already **spent** — nothing is reclaimed. The audit's real product is a *behaviour mirror*: how much premium (Opus/Fable) quota was **misallocated** to Sonnet-shaped sessions, so the user adopts better routing habits next window.

**Stacked on PR #453 (`premium-tier-predicate`) — merge #453 first, then this.** When #453 merges to main, GitHub retargets this PR to main.

## Summary
- Rename the headline field `percent_quota_reclaimable` -> `percent_quota_misallocated` across the `OpusQuotaAudit` dataclass, `audit_to_dict` / `audit_from_dict`, the analyzer fill, the CLI, and the JSON/API payload.
- **Deprecated alias (not a clean rename).** The old key shipped publicly in **0.5.4** (verified live on PyPI), so `audit_to_dict` keeps emitting `percent_quota_reclaimable` as a documented deprecated alias for one release; `audit_from_dict` reads the new key with a fallback to the old. The alias is emitted on **both** the direct and serve paths, so the byte-identical `DataAccess` parity contract still holds.
- **Persona-gate the headline** via the existing `framing` plan-tier machinery:
  - subscription / local -> measured mirror + habit nudge ("that share stays available for hard problems next window — `/model sonnet` for mechanical work, right-size subagents").
  - api-billed -> measured mirror + honest retrospective, already-billed dollar counterfactual ("same work at the suggested tiers ≈ \$X vs your \$Y actual"). This replaces the old undifferentiated "Implied API value" calibration line.

## Release-timing decision
`percent_quota_reclaimable` is present in tag **v0.5.4**, which is the current latest release on PyPI. So it **has** shipped publicly → I kept the deprecated alias rather than doing a clean removal. It should be dropped one release later (0.7.x).

## Scope guard — deliberately untouched (honest as-is)
- `tj savings` "reclaimed" (measured avoided tokens)
- `estimated_recoverable_usd/tokens` in optimize findings (forward-looking, persona-contracted)
- the statusline "→ /compact to reclaim quota" (forward action in a live session)
- `cmd_optimize._reclaimable_share` / `cmd_tokenmaxx` `reclaimable` (different surfaces: DowngradeFinding / context-diagnostic)

The onboarding next-steps line already read "how much Opus/Fable went to Sonnet-shaped sessions" — consistent with the new framing, so no change needed.

## Tests / Verification
- Renamed `test_percent_quota_misallocated_*`; updated field/key assertions across `test_opus_quota_audit.py`, `test_quota_audit_serve.py`.
- Added per-persona CLI tests: subscription sees the habit nudge (no dollars); api sees the already-billed dollar counterfactual (no nudge). Both assert the headline says "went to Sonnet-shaped sessions" and never "reclaimable".
- **Parity:** `tests/integration/test_data_access_parity.py::test_quota_audit_parity` passes, and is strengthened to assert the new key is present and the deprecated alias mirrors it byte-for-byte.
- JSON / serve tests assert both the new key and the deprecated alias are emitted with the same value.
- `make test` (2367 passed, 1 skipped e2e), `make lint` (ruff clean), `make typecheck` (mypy clean, 191 files).

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -129,7 +129,7 @@ Key flags: `--session`, `--last` (default when `--session` is omitted), `--json`
 
 ### `tj quota-audit`
 
-Retroactive audit of your premium (Opus/Fable) quota: which past premium-tier sessions were structurally Sonnet-shaped (small input/output, few tool calls)? Covers Fable (the tier above Opus) as well as Opus. Reports the percent of premium quota reclaimable, example sessions to spot-check, and an optional tuned routing-config export. Quota-share framing, never a dollar "saving" claim. Needs a direct DB connection (can't run against a live `tj serve`).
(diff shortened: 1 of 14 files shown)


anshss:

PR #455, opened 2026-07-10
perf(backfill): columnar bulk-append cuts full-history ingest ~11x

anshss:
Full-history Claude Code backfill fed DuckDB one row at a time, so a ~5–6 GB history took ~8 minutes. The bottleneck is Python↔DuckDB row marshalling, not the engine — this switches the ingest to a dependency-free columnar bulk-append.

## Summary
- Add `DuckDBBackend.bulk_insert_spans`: writes a batch once as newline-delimited JSON and lets DuckDB's native `read_json` scan it in a single vectorized `INSERT..SELECT`. No pandas/pyarrow dependency (DuckDB reads JSON natively), so the lean base install is unchanged.
- `ingest_claude_code` now batches spans across sessions and flushes them: ONE set-based existence check per flush (replacing the per-session `IN`-list) + ONE bulk-append, instead of a dedup query + `executemany` per session.
- `--reingest` keeps the per-session path (its per-span attribute overlay is not a bulk op) but routes its inserts through the same columnar append.

## Root cause
DuckDB inserts 100k rows from pure SQL in ~0.07s, but the same rows via `executemany` take ~190s — the row-at-a-time binding across the Python↔DuckDB boundary is the wall, not the columnar engine or the index. Bigger multi-row `VALUES` statements don't help (parsing a 50k-parameter statement is itself pathologically slow); the fast, dep-free path is handing DuckDB a file it can scan.

## The fix in detail
- **Columnar append**: `bulk_insert_spans` streams the batch to a temp NDJSON file (outside the write lock), then holds the lock only for one `INSERT INTO spans SELECT … FROM read_json(?, …)`. JSON columns are written as nested objects (not pre-stringified) and read back with an explicit `read_json` column schema, so `attributes`/`events`/`request_params`/`request_tools` land as JSON values — not double-encoded strings. Timestamps round-trip as ISO-8601 → `TIMESTAMPTZ`.
- **Idempotent**: a `WHERE NOT EXISTS` anti-join skips any `span_id` already present, so a span another writer (or an earlier backfill) inserted mid-flight is skipped rather than raising a PRIMARY KEY conflict. Deterministic span IDs and skip-existing-on-re-run are unchanged.
- **Batching**: one existence check + one append per flush; session upserts stay per-file in order (so `started_at` is set by the first-processed file exactly as before), and the post-loop passes (session-total recompute, method snapshots, stale-scheme reconciliation) are untouched. The per-session progress callback still fires once per session.
- **Safety**: if a bulk flush raises, it degrades to the per-session path (the `INSERT..SELECT` is atomic, so nothing partially landed) — preserving the `files_failed` accounting.

## Performance
Synthetic 400-session / 32k-span history through the real ingest path:
- Row-at-a-time `insert_span`: ~56s
- New columnar batch ingest: ~5.1s (**11x**)

The remaining 5s is now dominated by unchanged Python JSONL parsing, so a real multi-GB / ~500k-span history extrapolates to well under 2 minutes.

## Tests / Verification
- `tests/unit/test_bulk_insert_spans.py` (new): pins `bulk_insert_spans` byte-for-byte against the per-row `insert_span` (unicode, nested JSON, request params/tools, NULLs) and proves idempotency + empty no-op.
- `tests/unit/test_backfill.py` (new cases): a tiny flush target forces multiple batch flushes and sessions split across boundaries; asserts the resulting spans + session rows are identical to a single-flush ingest, and that a re-run inserts nothing (idempotency across the boundary).
- `tests/integration/test_storage_backend_parity.py`: classify the new write method.
- Full suite: `make test` → 2348 passed, 2 skipped. `ruff check tokenjam/` and `mypy tokenjam/` clean.

## What's NOT in this PR
- No new runtime dependency (deliberately dep-free; `orjson` for the JSON encode was considered but skipped to keep the base install lean — the JSON encode is not the bottleneck).
- The live in-process / HTTP ingest path (`insert_span`) is unchanged — this targets only the batch backfill hot path.

### DIFF
--- a/tests/integration/test_storage_backend_parity.py
+++ b/tests/integration/test_storage_backend_parity.py
@@ -101,6 +101,7 @@ SHIM_KNOWN_GAPS = {
 # implementing one, ``test_unimplemented_methods_have_no_silent_shim`` fails so
 # the new method cannot land without parity coverage or an explicit gap entry.
 SHIM_NOT_IMPLEMENTED = {
+    "bulk_insert_spans",
     "close_session_by_id",
     "close_sessions_by_instance",
     "count_traces",
--- a/tests/unit/test_backfill.py
+++ b/tests/unit/test_backfill.py
@@ -1448,6 +1448,34 @@ def test_bulk_batch_flush_boundary_matches_single_flush(tmp_path, monkeypatch):
         chunked.close()
 
 
+def test_bulk_progress_counts_increase_monotonically(tmp_path):
+    """Regression: in the bulk path the progress callback must observe
+    `spans_ingested` climbing per session — not sit at flat zero until a single
+    end-of-run flush, which would make a live backfill display (onboard) show 0
+    the whole run then jump. Uses the DEFAULT (large) flush target so every
+    insert is deferred to ONE flush at the end — the exact scenario that exposed
+    the bug — and asserts the callback still saw increasing counts."""
+    _multi_session_tree(tmp_path, n_sessions=6)  # 6 sessions × 2 spans each
+
+    observed: list[int] = []
+
+    def _capture(parsed, result):
+        observed.append(result.spans_ingested)
+
+    db = InMemoryBackend()
+    try:
+        ingest_claude_code(db, root=tmp_path, progress=_capture)
+        # One callback per session; counts advance 2 per session (LLM + tool),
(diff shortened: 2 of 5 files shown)


anshss:

PR #461, opened 2026-07-10
fix(daemon): never write an ephemeral uv/pipx cache path into launchd/systemd

anshss:
## Summary
When onboarding runs through a throwaway `uvx --from tokenjam tj onboard` / `pipx run --spec tokenjam tj onboard` session and the persistent-install offer is declined (or the session is non-interactive), the daemon installer fell back to a `tj` path inside uv's ephemeral tool-archive cache (`~/.cache/uv/archive-v0/<hash>/...`).
`uv cache prune` / `uv cache clean` (routine maintenance, also run by some CI/cleanup tools) deletes that path outright, which silently kills the daemon on the next `launchctl`/`systemctl` load and freezes the daemon on whatever version was resolved at onboard time, independent of the wrapper's freshness logic.
The daemon installer now detects an ephemeral `tj` path and instead points `ProgramArguments`/`ExecStart` at the stable `uvx`/`pipx` shim itself (e.g. `uvx --from tokenjam tj --config ... serve`), so launchd/systemd re-resolves `tj` through uv/pipx on every start rather than a path that can vanish out from under it. If no durable entrypoint exists at all, it warns and skips the daemon install rather than silently writing a cache path.

## Test plan
[x] `uv run pytest tests/unit/test_onboard_daemon.py tests/unit/test_onboard_ephemeral_guard.py -q` — 35 passed, including new coverage: direct-`tj` path preferred when not ephemeral, `uvx` shim fallback when `tj` resolves into `archive-v0`, `pipx run` shim fallback, `None` (skip + warn) when no durable entrypoint exists, and end-to-end plist/unit generation asserting no `archive-v0` substring appears.
[x] `uv run pytest tests/unit -q` — no new failures (7 pre-existing, environment-local failures reproduce identically on `main` before this change, unrelated to this fix).


anshss:
Lands via #463 (bundled batch) — stays open and will auto-mark merged when the anchor lands. Review-bot finding here (dead monkeypatch) is fixed on this branch and included in the union.

PR #462, opened 2026-07-10
Extend read timeout for heavy computed serve endpoints

anshss:
## Problem

When `tj serve` is running it holds the DuckDB write lock, so CLI commands route their queries through the daemon's REST API via `ApiBackend`. That client had a **blanket 10s read timeout** — fine for the cheap shim reads (traces/cost/status/alerts), but too tight for the *computed* endpoints that re-aggregate a user's whole history server-side.

On a large history (e.g. 3,424 sessions / 150k turns, 30d window) the daemon computes `GET /api/v1/quota-audit` in ~13s. Under the 10s ceiling that raised `httpx.ReadTimeout`, so `tj quota-audit` failed **deterministically** against a running daemon — and the bigger the history, the more certain the failure. The cheap `/api/v1/context` (0.6s) and other reads completed fine under the same daemon.

## Fix

Add a per-request **read-timeout override (60s)** on the heavy computed fetchers — `quota-audit`, `context`, `optimize`, `reuse/clusters`, `cost/compare` — while keeping the tight **10s default** for the cheap reads so a genuinely wedged daemon still fails fast.

`ApiBackend._get` gains an optional `timeout` kwarg that overrides the client's blanket timeout for a single request.
The heavy fetchers pass `timeout=_HEAVY_ENDPOINT_TIMEOUT` (60s); cheap reads pass nothing and keep `_DEFAULT_TIMEOUT` (10s).

## Validation

Real-socket before/after run against an HTTP server whose `quota-audit` handler sleeps 12s (mimicking the >10s server-side compute):

| Path | Behavior |
|------|----------|
| Blanket 10s (old) `_get("/api/v1/quota-audit")` | `ReadTimeout at 10.0s` — reproduces the bug |
| Fixed `fetch_opus_quota_audit` (60s override) | `OK in 12.0s` — completes |
| Cheap read `get_traces` on a slow path | `ReadTimeout at 10.0s` — tight ceiling preserved |

New regression test `tests/unit/test_api_backend_timeout.py` asserts the wire-level timeout httpx attaches to each request: heavy endpoints get 60s, cheap reads keep 10s.
Existing serve-integration test doubles updated to match the new `_get` signature.
`ruff` + `mypy` clean; full unit + integration suites green (one unrelated pre-existing env-artifact failure in `test_onboard_path_branch`).

### DIFF
--- a/tests/integration/test_context_serve.py
+++ b/tests/integration/test_context_serve.py
@@ -125,7 +125,7 @@ def _apibackend_wired_to(app, monkeypatch) -> ApiBackend:
     api = ApiBackend("http://daemon")
     transport = httpx.ASGITransport(app=app)
 
-    def _sync_get(path, params=None):
+    def _sync_get(path, params=None, *, timeout=None):
         async def _call():
             async with httpx.AsyncClient(
                 transport=transport, base_url="http://daemon",
--- a/tests/integration/test_data_access_parity.py
+++ b/tests/integration/test_data_access_parity.py
@@ -75,7 +75,7 @@ def _serve_access(app, monkeypatch) -> ServeDataAccess:
     api = ApiBackend("http://daemon")
     transport = httpx.ASGITransport(app=app)
 
-    def _sync_get(path, params=None):
+    def _sync_get(path, params=None, *, timeout=None):
         async def _call():
             async with httpx.AsyncClient(
                 transport=transport, base_url="http://daemon",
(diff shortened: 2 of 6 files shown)


anshss:
Lands via #463 (bundled batch) — stays open and will auto-mark merged when the anchor lands. Review-bot finding here (read-leg-scoped timeout) is fixed on this branch and included in the union.

Issue #377, opened 2026-07-02
Window token-totals omit cache-creation (write) tokens, so cost-comparison tokens_delta understates volume


anshss:
**Severity:** Low
**Where:** `tokenjam/core/cost.py` (`WindowTotals.total_tokens`, ~L133-145) + `tokenjam/core/db.py` (`get_window_cost_totals` SQL, ~L1550-1564)

**Observed:** A recent change split cache tokens into separate read (`cache_tokens`) and write/creation (`cache_write_tokens`) buckets and updated every rollup to carry writes — except the window token-total used by the SDK cost-comparison view. `WindowTotals.total_tokens` computes `input + output + cache_tokens` and the `get_window_cost_totals` SQL sums only `cache_tokens`, never `cache_write_tokens`. As a result `tokens_delta` / `tokens_delta_pct` in the cost-comparison view slightly understate token volume for any window containing cache-creation tokens. **Dollars are unaffected** — per-span `cost_usd` already prices cache-write tokens correctly; this is a token-count display discrepancy only. This is the one place that still sums cache reads without writes while every other aggregation carries both.

**Fix direction:** Include `cache_write_tokens` in `WindowTotals.total_tokens` and in the `get_window_cost_totals` SQL sum, matching how the session rollups (`db.py:604, 1070`) already carry both read and write buckets.

**Done when:** `total_tokens` / `tokens_delta` include cache-creation tokens; a test covers a window containing `cache_write_tokens` and asserts the delta reflects them.

PR #468, opened 2026-07-10
Include cache-write tokens in optimize-path token totals

anshss:
## Problem

Two optimize-path token totals summed input + output + cache-read tokens
but silently omitted cache-write (cache-creation) tokens:

- `summarize_window` (`tokenjam/core/optimize/runner.py`) — its `total_tokens`
  feeds the reclaimable-share denominator used by every optimize analyzer's
  reclaimable-% figure.
- `analyze_model_downgrade` (`tokenjam/core/optimize/analyzers/model_downgrade.py`)
  — its `window_total_tokens` feeds the model-downgrade analyzer's
  candidate-token share.

For any window with meaningful cache-creation volume, both totals understated
the true token volume, which inflated the derived percentages.

This is the same class of bug as public issue #377, which PR #450 fixed for
the cost-comparison window path (`WindowTotals` / `get_window_cost_totals`).
That PR did not touch these two optimize-path call sites, so the gap
persisted there.

## Fix

Add `COALESCE(SUM(cache_write_tokens), 0)` to `summarize_window`'s token SQL
sum, and add `cache_write_tok` to `window_total_tokens`'s accumulation in
`analyze_model_downgrade`, matching the pattern PR #450 established for
`get_window_cost_totals`.

## Other sites noted (not fixed, out of scope for this change)

While verifying no other token-total site has the same gap, I found two more
that also omit cache-write tokens from a "total tokens" figure:

- `analyze_model_downgrade`'s `candidate_tokens` accumulation (same file,
  just below the fixed `window_total_tokens` line) — sums input+output+cache
  for candidate sessions only, still missing cache-write.
- `workflow_restructure.py`'s per-cluster `cluster_tokens` aggregate
  (`SUM(input_tokens + output_tokens + cache_tokens)`) — same gap pattern.

Flagging these for a follow-up rather than expanding this PR's scope.

## Test plan

- Added `test_summarize_window_total_includes_cache_write_tokens` and
  `test_downgrade_window_total_tokens_includes_cache_write_tokens` to
  `tests/unit/test_optimize.py`, asserting a cache-write-heavy window's
  totals include the cache-creation volume.
- `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` —
  2432 passed, 1 pre-existing unrelated failure (an onboarding test that
  fails locally due to a real `~/.config/tj/config.toml` on this machine;
  reproduces identically on `main` with no changes applied), 2 skipped.
- `ruff check` clean on all changed files.

### DIFF
--- a/tests/unit/test_optimize.py
+++ b/tests/unit/test_optimize.py
@@ -100,6 +100,51 @@ def test_summarize_window_total_includes_cache_read_tokens(db):
     assert s.total_tokens == 1000 + 200 + 5000
 
 
+def test_summarize_window_total_includes_cache_write_tokens(db):
+    """Window header total must also include cache-CREATION tokens, not just
+    cache-read, so a cache-write-heavy window's reclaimable-share denominator
+    (cmd_optimize._reclaimable_share) isn't inflated."""
+    start = utcnow() - timedelta(days=2)
+    db.insert_span(make_llm_span(
+        model="claude-opus-4-7",
+        provider="anthropic",
+        input_tokens=1000,
+        output_tokens=200,
+        cache_tokens=500,
+        cache_write_tokens=8000,
+        cost_usd=0.030,
+        session_id="cache-write-heavy",
+        start_time=start,
+    ))
+    since = utcnow() - timedelta(days=30)
+    until = utcnow() + timedelta(hours=1)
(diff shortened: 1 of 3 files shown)


anshss:

PR #469, opened 2026-07-10
fix(onboard): daemon install resolves tj via interpreter sibling, not PATH

anshss:
## Problem

The background-daemon installer (\`_install_launchd\`/\`_install_systemd\`) resolves the \`tj\` binary to bake into the daemon's ProgramArguments via \`_resolve_tj_binary()\`, which used to prefer \`shutil.which("tj")\` first, falling back to the interpreter sibling only if PATH resolution failed.

If an older/other \`tj\` happened to be earlier on PATH at the exact moment \`tj onboard\` installed the daemon (a real-world PATH-shadow scenario — confirmed elsewhere in this codebase for a VS Code integrated terminal ordering a stale pip \`tj\` ahead of a freshly-installed uv-tool shim), the daemon unit got permanently pinned to the wrong absolute binary in its plist/unit file. The daemon then silently runs a stale \`tj serve\` indefinitely — surviving even after the shadowing PATH entry is later removed, since the plist/unit already has the wrong path baked in.

## Fix

Flip \`_resolve_tj_binary()\`'s priority to match \`_current_tj_binary()\` (used elsewhere for the same PATH-shadow problem): prefer the PATH-independent sibling next to the running interpreter (\`sys.executable\`'s directory), falling back to \`shutil.which("tj")\` only when no sibling exists.

The final fallback intentionally still returns the *constructed* sibling path rather than a bare \`"tj"\` (unlike \`_current_tj_binary\`'s last resort) — \`_daemon_program_args\`'s ephemeral uv/pipx-cache detection needs a real path shape (containing \`/uv/\` or \`/pipx/\`) to route the daemon through a durable uvx/pipx shim instead of a cache path that \`uv cache prune\` could delete.

## Test plan

- Added regression tests in \`tests/unit/test_onboard_path_resolution.py\` mirroring the existing \`_current_tj_binary\` coverage (prefers sibling over a shadowing PATH tj; falls back to \`which\` when no sibling; falls back to the constructed sibling path, not bare \`tj\`, when nothing resolves).
- Updated two tests in \`tests/unit/test_onboard_daemon.py\` whose fixtures had been implicitly relying on the old which-first ordering (one ran against the real dev venv, where the sibling \`tj\` actually exists, so it silently depended on \`which\` losing).
- \`make test\`: 2433 passed, 2 skipped, 1 pre-existing failure unrelated to this change (\`test_bare_onboard_sdk_choice_falls_through_to_generic\` — fails identically on unmodified \`main\`; an isolation bug that hits the real \`~/.config/tj/config.toml\` on this machine).
- \`make lint\` / \`make typecheck\`: clean.

### DIFF
--- a/tests/unit/test_onboard_daemon.py
+++ b/tests/unit/test_onboard_daemon.py
@@ -97,11 +97,35 @@ class TestTjBinaryResolution:
     matches inside `/python3`). The unit is written pointing at a binary that
     doesn't exist; `launchctl load` still returns 0, so onboarding reports
     success while `tj serve` never launches.
+
+    Also regression for a PATH-shadow bug: `_resolve_tj_binary` used to
+    prefer `shutil.which("tj")` over the interpreter sibling, so an
+    older/other `tj` earlier on PATH at the moment `tj onboard` installs the
+    daemon got permanently baked into the launchd/systemd unit — surviving
+    even after the shadowing PATH entry was later removed. It must now
+    prefer the PATH-independent sibling next to the running interpreter,
+    the same priority `_current_tj_binary` uses.
     """
 
-    def test_prefers_tj_on_path(self, monkeypatch):
+    def test_prefers_interpreter_sibling_over_a_shadowing_path_tj(self, monkeypatch, tmp_path):
+        """A `tj` earlier on PATH must never win over the sibling next to the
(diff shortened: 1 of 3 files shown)


anshss:

PR #470, opened 2026-07-10
Print a repair hint instead of a raw traceback for a dangling editable install

anshss:
## Problem

If \`tokenjam\` is installed with \`pip install -e .\` (or via a tool that does this under the hood) and the source checkout it points at is later removed (e.g. a deleted git worktree), the \`tj\`/\`tokenjam\` CLI degrades: \`tokenjam\` resolves as an empty Python namespace package (no \`__file__\`, no real \`cli\` subpackage on disk), so every subcommand dies with a bare traceback:

\`\`\`
Traceback (most recent call last):
  File ".../bin/tj", line 5, in <module>
    from tokenjam.cli.main import cli
ModuleNotFoundError: No module named 'tokenjam.cli'
\`\`\`

Nothing in that traceback tells the user the actual cause (a dangling editable install) or how to fix it.

## Fix

- Added \`tokenjam/cli/entrypoint.py\`, a small module registered as the \`tj\`/\`tokenjam\` console scripts. It tries the real import; on \`ModuleNotFoundError\` it checks specifically for the namespace-package signature (\`tokenjam.__file__ is None\` and no contributing path holds a real \`cli/main.py\`) and prints a one-line hint instead of letting the traceback through:

  > \`tj's install points at a source path that no longer exists. Reinstall with \`pipx reinstall tokenjam\` or \`pip install -e <checkout>\`.\`

  Any other \`ModuleNotFoundError\` (e.g. a genuinely missing dependency) still propagates normally — this only catches the specific dangling-editable case.
- The tricky part: the console-script stub that pip generates does a hard, unguarded top-level import of whatever module the entry point names. If that target module lives entirely inside the vanished editable source tree, there's nothing left on disk to run a guard from. So \`entrypoint.py\` is also added to \`[tool.hatch.build.targets.wheel.force-include]\`, which makes hatchling physically copy it into site-packages even for an editable install (the same mechanism already used to ship the \`incidents/\` demo data) — so it's guaranteed to still be present and importable even after the rest of the source is deleted.

## Test plan

- Added \`tests/unit/test_cli_entrypoint.py\`: unit tests simulating the namespace-package condition via fake module objects (dangling case detected, benign namespace package with real \`cli/main.py\` not misdetected, real install not misdetected, detector never raises), plus tests for \`main()\`'s hint/exit-1 path, its re-raise of unrelated \`ModuleNotFoundError\`s, and its normal success path.
- Manually verified end-to-end outside the test suite: built a real editable install in a scratch venv, confirmed \`tokenjam/cli/entrypoint.py\` is physically present in site-packages, ran \`tj --version\` (works), deleted the source checkout, ran \`tj --version\` again — now prints the one-line hint and exits 1 instead of a traceback (confirmed for another subcommand too).
- \`make test\`: 2436 passed, 2 skipped, 1 pre-existing unrelated failure (\`test_bare_onboard_sdk_choice_falls_through_to_generic\`, fails identically on a clean checkout of this branch before my changes — caused by a real \`~/.config/tj/config.toml\` on the test machine leaking into an isolated-filesystem test, unrelated to this change).
- \`ruff check\` and \`mypy\` clean on the new/changed files.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,8 +72,8 @@ mcp       = []
 bloat     = ["llmlingua>=0.2"]
 
 [project.scripts]
-tj = "tokenjam.cli.main:cli"
-tokenjam = "tokenjam.cli.main:cli"
+tj = "tokenjam.cli.entrypoint:main"
+tokenjam = "tokenjam.cli.entrypoint:main"
 
 [tool.pytest.ini_options]
 testpaths = ["tests/unit", "tests/synthetic", "tests/agents", "tests/integration"]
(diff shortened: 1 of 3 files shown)


anshss:

Issue #477, opened 2026-07-12
Add `tj optimize --validate`: measure a finding's actual token/cost/quality delta on your own recorded calls


anilmurty:
## Problem

The analyzers report **estimated** recoverable (heuristic). But a user rightly wants to know whether a specific recommendation *actually* saves on **their** usage — and whether quality survives — **before** acting on it. Today there's no way to measure that; the number stays an estimate.

## Proposal

`tj optimize --validate <finding>` re-runs the candidate against the baseline on a small sample of the user's **own recorded calls** (using their own API key) and reports the **measured** token/cost delta plus a quality-preservation check. It turns "estimated recoverable" into "measured on your data."

## Scope (v1)

- `downsize` and `summarize`/`trim` findings — the ones where "does it still work?" is the open question.
- `cache` is deterministic (savings are computable without an A/B) — out of scope here.

## Mechanism

1. Pull the recorded prompts for the finding's sessions. **Requires prompt capture on** — off by default (privacy), so v1 serves the **subset of users who've enabled capture**; it's an opt-in power feature, not universal. Fail with a clear, actionable message (and the capture-config hint) when it's off.
2. **Sample** K calls (default small, e.g. 5–10; cost-bounded). Uses `TJ_ANTHROPIC_API_KEY`; print the estimated cost and require confirmation before spending.
3. Run each sampled call **two ways** through the real provider — baseline (as recorded) vs candidate (downsize → the cheaper model; summarize → the trimmed prompt).
4. Measure **actual** input/output/cache tokens + cost for both.
5. **Quality signal:** exact-match for deterministic/tool outputs; content-presence for summarize (did the key facts survive); an LLM-judge or embedding-similarity for open-ended (later). Report `preserved N/K`.
6. Output, e.g.: `measured over 5 calls: 40,120 → 24,050 tokens (−40%), $0.61 → $0.37, quality preserved 5/5`.

## Honesty (Rule 14)

Output framing is **"measured on a sample of your usage"** — never a guarantee, and never "certified"/"guaranteed savings" (that word is reserved for a different, later layer). The quality line is the point: a saving that keeps the same result. Report the sample size prominently so it can't read as a fleet-wide guarantee.

## Reuse

[...]

PR #479, opened 2026-07-12
[feature] Add verbosity analyzer (#478)

### LINKED ISSUE #478: Add a `verbosity` analyzer: surface high output-token spend (the output-side lever the other analyzers miss)
## Problem

All five analyzers target the **input** side — prompt size, model choice, caching, repeated shapes. **None looks at output tokens.** An agent that generates verbose prose burns output tokens (often the pricier per-token side), and that waste is completely invisible today.

## Proposal

A new self-registering `verbosity` analyzer that surfaces agents/sessions with **high output-token spend relative to the task**, as review candidates — the same "here's a candidate to look at" model as the other analyzers.

## Detection heuristic (v1 — tune on real data)

Flag `(agent, model)` / sessions where output tokens run high relative to a baseline:
- output tokens well above the **median for the same `(tool, arg-shape)` / task shape** — **preferred**: it's the only signal grounded in like-for-like tasks;
- high **absolute output spend** for a `(provider, model)`;
- high **output:input ratio** — **weakest signal**, use with caution: a legitimately long answer to a short prompt is not waste.

## Framing (Rule 14) — this is the least-grounded analyzer, treat it that way

Unlike `trim` (which removes *provably-redundant input*), **output length ≠ waste**: a terse answer can drop information the task needed, and "wasteful output" is task-dependent. So verbosity's recoverable estimate is inherently **softer** than the input-side analyzers. Concretely:

- "predicted high-verbosity output — **review before constraining a response**." A candidate, never "you're wasting X."
- Carries the recoverable-savings contract fields (`estimated_recoverable_usd` / `estimated_recoverable_tokens`, `estimate_basis="heuristic"`), recoverable = output above a sane baseline priced at output rates — **but framed more conservatively than peers.** Prefer not to assert a hard dollar figure until it's been measured (see #477); a measured number is the honest one for this lever.

## Remedy (surfaced, not applied)

Recommend an output-brevity constraint — a terse system-prompt snippet and/or a `max_tokens` cap — for the user to apply. This pairs with #477 (`--validate`), which is **load-bearing here, not optional**: a brevity constraint can even be net-negative once its own overhead is counted, so measure before claiming.

## Implementation

Follows the established analyzer pattern: drop a file under `tokenjam/core/optimize/analyzers/`, decorate `@register("verbosity")`, operate on `db.conn`, append to `ANALYZER_ORDER` if ordering matters. `cmd_optimize`'s positional choices auto-derive — no edits there.

## Acceptance criteria

- [ ] `tj optimize verbosity` runs and surfaces candidates with the honest candidate framing.
- [ ] Self-registers — appears in `tj optimize`'s choices with no edits to `cmd_optimize`.
- [ ] Result dataclass carries the recoverable-savings contract fields; `report_to_dict`/`report_from_dict` round-trip it.
- [ ] Baseline defaults to the per-`(tool, arg-shape)` median (the most defensible signal); the recoverable figure is framed conservatively (softer than peer analyzers).
- [ ] Human-readable + `--json` output.
- [ ] Tests using the synthetic span factory (`tests/factories.py`) for detection + framing.

## Out of scope

- Actually applying the terse constraint (that's a remedy the user — or, later, the enforcement plane — does).
- Any fleet-scale enforcement.

## Related

- #477 — `tj optimize --validate` (measure a candidate's real savings; verbosity findings feed straight into it, and are the ones that most need measuring before a $ claim).
- Adjacent idea for a separate issue: **tool-schema bloat** — oversized tool/function JSON schemas are a real input-side cost and are now visible via captured `request_tools` (#209); a `trim`-adjacent finding.

anilmurty:
Adds the output-side cost lever the existing analyzers miss. Every current analyzer targets the INPUT side (prompt size, model choice, caching, repeated shapes); none looks at OUTPUT tokens, so verbose generation burns output tokens (the pricier per-token side) invisibly.

## Summary
- New self-registering `@register("verbosity")` analyzer (`tokenjam/core/optimize/analyzers/output_verbosity.py`).
- Detection leads with the most defensible signal: the per-`(tool, arg-shape)` task-shape **median** of output tokens. A session is a candidate only when its output exceeds `HIGH_VERBOSITY_MULTIPLE` (2.0×) its like-for-like cohort median, and only for cohorts of at least `MIN_COHORT_SESSIONS` (5).
- The output:input ratio is carried as a **weak, descriptive-only** field — never the flag (a long answer to a short prompt is not waste).
- Carries the #111 recoverable-savings contract: `estimated_recoverable_usd` = output above the cohort median priced at OUTPUT rates, `estimated_recoverable_tokens`, `estimate_basis`, `estimate_confidence="heuristic"`. Round-trips through `report_to_dict` / `report_from_dict`.
- Human-readable renderer + `--json`. Self-registers, so `cmd_optimize`'s positional choices auto-derive (no edits to its dispatch beyond the renderer/label wiring every wave-2 finding needs).

## Design rule honored (least-grounded analyzer)
Output length is not waste the way provably-redundant input is (`trim`), so verbosity is framed the softest of any analyzer:
- Framing is candidate-only: "predicted high-verbosity output — review before constraining a response." Never "you're wasting X" — the mandatory `VERBOSITY_HONESTY_CAVEAT` says output length is not waste and is carried as a dataclass default so no surface can drop it (Rule 14).
- The recoverable figure is an explicit **soft upper bound**; `estimate_basis` states a brevity constraint can be net-negative once its overhead is counted and points at measurement (`--validate`, #477) as the honest number.

## Remedy (surfaced, not applied)
Recommends an output-brevity constraint — a terse system-prompt snippet (`remedy_snippet`) plus a suggested `max_tokens` cap (the cohort median baseline) — for the user to apply and then **measure**. Nothing is written or applied.

## Detection heuristic + baseline (chosen)
- **Baseline:** per-task-shape median output tokens, where a session's task shape is the ordered `(tool_name, arg_shape)` tuple of its tool calls (the same signature the `script` analyzer builds).
- **Flag:** `session output_tokens > 2.0 × cohort median`, cohorts of ≥5 sessions only.
- **Recoverable:** `(output_tokens − cohort median)` per candidate, priced at that session's dominant model's output rate.

## Tests / Verification
- `tests/unit/test_optimize_verbosity.py` (12 tests, all via `tests/factories.py` per Rule 8): median-baseline detection, no-flag-when-near-median, sub-cohort-size guard, ratio-is-descriptive-not-the-flag, the recoverable contract (output-above-baseline at output rates), empty-window contributes nothing, honest-caveat framing, remedy surfaced not applied, `report_to_dict`/`report_from_dict` round-trip, Click-choice + renderer wiring, and CLI render across all four pricing modes.
- `ruff check tokenjam/` — clean. `mypy tokenjam/` — clean (194 files). `pytest tests/unit tests/synthetic` — 2056 passed; integration optimize/mcp/cli — 136 passed.

## What's NOT in this PR (deliberately deferred)
- Actually applying the terse constraint (out of scope per #478 — a remedy the user or the enforcement plane applies).
- The measured-savings path via `tj optimize --validate` (#477) — verbosity findings feed straight into it; until then the $ figure is an explicit soft estimate.
- A Lens Optimize-tab renderer for verbosity (the CLI + `--json` + registry-driven Overview waste band pick it up; no UI JS was added).

Closes #478

### DIFF
--- a/docs/optimize/verbosity.md
+++ b/docs/optimize/verbosity.md
@@ -1,0 +1,96 @@
+# Verbosity
+
+Product name: **Verbosity**. Internal/CLI name: `verbosity`.
+
+```bash
+tj optimize verbosity
+```
+
+The other analyzers target **input** and model choice. Verbosity is the one
+that looks at **output** — sessions burning tokens on answers that run long for
+the work they did. It flags a session only when its output is a clear outlier
+against sessions doing the *same shape* of work, then surfaces a brevity remedy
+for you to measure.
+
+## The weakest-grounded analyzer — treated that way
+
+Output length is not waste. A terse answer can drop information the task needed,
+and "too verbose" is task-dependent — so verbosity's recoverable figure is the
+least defensible of any analyzer. Two consequences are baked in:
+
+- The baseline is the **per-`(tool, arg-shape)` median** — the only signal
+  grounded in like-for-like tasks. A high output:input ratio is carried as a
+  descriptive field only; it never flags a session on its own (a legitimately
+  long answer to a short prompt is not waste).
+- The recoverable figure is a **soft upper bound**, and `--validate` (measuring
+  a brevity constraint on your own calls) is load-bearing, not optional, before
+  anyone claims a dollar saving.
+
+## What it flags
+
+Sessions are grouped into cohorts by their task-shape signature — the
+`(tool_name, arg_shape)` sequence, the same signature the [Script](script.md)
+analyzer uses. For each cohort:
+
+| Step | Rule |
+|---|---|
+| Cohort must be real | At least `MIN_COHORT_SESSIONS` (5) sessions share the signature, else its median is noise and the cohort is skipped. |
+| Baseline | The cohort's **median** output tokens. |
+| Flag | A session whose output exceeds `HIGH_VERBOSITY_MULTIPLE` (2.0×) the cohort median. |
+
+Thresholds live as module constants in
+`tokenjam/core/optimize/analyzers/output_verbosity.py`
+(`MIN_COHORT_SESSIONS`, `HIGH_VERBOSITY_MULTIPLE`, `MAX_EXAMPLES`).
+
+It reads aggregate token counts and the tool-call shape — no prompt/completion
+content capture required.
+
+## Output
+
+The finding reports, for the window:
+
+- `total_candidates` — the true number of flagged sessions
+- `candidates` — the top `MAX_EXAMPLES` (5) by over-baseline output; when more
+  were flagged the CLI shows "showing top 5 of N" so the headline never
+  under-reports
+- per candidate: `output_tokens`, `baseline_output_tokens` (the cohort median),
+  `over_baseline_tokens`, `over_baseline_multiple`, and the descriptive-only
+  `output_input_ratio`
+- `suggested_max_tokens` — the cohort baseline, an advisory `max_tokens` cap to
+  review (never applied)
+- `sessions_examined` / `cohorts_examined`
+
+Rendering follows the same plan-tier-aware convention as the rest of
+`tj optimize`: `api` plans see the dollar figure, subscription/local/unknown
+plans see the over-baseline token figure instead.
+
+## Estimate basis / confidence
+
+`estimated_recoverable_tokens` is the over-baseline output summed across flagged
+sessions; `estimated_recoverable_usd` prices it at **output** rates.
+`estimate_confidence` is `"heuristic"` and `estimate_basis` reads:
+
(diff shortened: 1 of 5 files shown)


anilmurty:
Addressed both Greptile comments in efcabf9:
- **P1 (silent truncation):** added `total_candidates` (set before the display slice, round-tripped through `report_to_dict`/`from_dict`); the renderer now shows the true flagged count with a "showing top 5 of N" note when capped.
- **P2 (rounding drift):** `over_baseline_tokens` now uses `int(round(...))` to match `baseline_output_tokens`.

Added a regression test for the truncation + round-trip. ruff/mypy clean; 13 verbosity tests pass.

PR #482, opened 2026-07-13
feat: cut what your agents cost by ending their repeated mistakes

anshss:
> ## Anchor PR: merge this one
>
> This PR contains the **complete change** and is the single merge that lands it. It has been split into 12 concern-scoped PRs below purely to make review tractable. Each slice is stacked on the previous one, so its diff shows only that slice.
>
> **Do not merge the slices individually.** Merging this anchor (with a merge commit, not a squash) brings every commit in the stack into `main` in one action. Slice CI may be red where a branch references code that lands in a later slice; that is an artifact of ordering by concern rather than dependency. CI on this anchor is authoritative.
>
> | # | PR | Slice |
> |---|----|-------|
> | 1 | #510 | scope `tj stop`/`reset` to the invoking install's daemon |
> | 2 | #511 | extract shared token-accounting, clustering, transcript-cache helpers |
> | 3 | #512 | `[optimize]` analyzer thresholds config |
> | 4 | #513 | ⚠️ capture prompt text and tool inputs by default (behavioral) |
> | 5 | #514 | `--json` as a global flag |
> | 6 | #515 | four cost analyzers |
> | 7 | #516 | `relearn` detector with reversible apply |
> | 8 | #517 | cost-proposal inbox |
> | 9 | #518 | CLI + local web API surface |
> | 10 | #519 | ⚠️ remove dead hook/savings/output-cap surfaces (deletions) |
> | 11 | #520 | Review-inbox UI |
>
> **#521 (docs / public copy) is deliberately NOT part of this anchor.** It is based on `main` and merges independently, so merging this anchor does not bring any README/docs change into `main`. Anchor + #521 together reconstruct the full original tree (verified identical tree SHA).
>
> The split preserves every commit's content; this anchor is byte-identical to the code stack tip.

## What this does

Adds the detect and propose and apply surface that turns TokenJam's analyzers into reviewable, reversible fixes. Two things are new:

- A cross-session recurring-failure detector (`relearn`) that finds the failures a coding agent silently re-hits across unwatched sessions, clusters them into signatures, and turns each into a stored proposal a human applies behind a dry-run gate. Every apply is snapshotted, git-committed when the target is tracked, and reversible.
- A cost-proposal inbox that adapts each cost analyzer's finding into a typed card, surfaced in both the web Review inbox and the terminal, so findings that used to be invisible outside the web UI now have a real CLI path.

Alongside those, four new cost analyzers land (dead MCP weight, repeat-context, per-agent model downsizing, batch-API placement), the previously CLI-invisible analyzer findings get real terminal renderers, the recoverable-dollar headline is made honest about analyzer overlap, three dead surfaces are retired (the `hook`, `savings`, and typeable `quickstart` commands, plus the `output_cap` module), and the new write endpoints are hardened with an unconditional local-write token.

Scope note: this PR stops at apply. An earlier iteration on the branch carried a verify/receipts layer that measured whether a fix's recurrence or cost actually dropped afterward; it was removed as unbounded (commit `c0316ab`), so `cost_verify.py`, `relearn_verify.py`, `relearn_otel_verify.py`, and `receipts.py` are not in the final tree. A few `verify`-named fields survive as inert storage scaffolding that nothing reads.

### New analyzers (`tokenjam/core/optimize/analyzers/`)

- **`relearn.py`** — the cross-session recurring-failure detector. Per session it builds the method spine, treats every errored tool step as a failure episode, and clusters episodes into signatures via validated regex families (wrong-cwd Read, edit-before-read, blocked sleep-chain, stale-read race, string-not-found, deferred-tool-cold, command-not-found, malformed Read offset, WebFetch domain-block), with everything else falling to a generic bucket then a bounded distill pass over the residual generics through the local `claude` CLI. Clusters already codified in a reachable `CLAUDE.md`/`learnings.md` are dropped. Surviving clusters recurring across three or more distinct sessions become proposals with a conservative token estimate, a target intervention rung, and a project-vs-user scope. Never raises. Registered as analyzer `relearn`.
- **`deadweight.py`** — MCP dead-weight plus an always-injected context-tax table. It enumerates configured MCP servers (read-only) and counts real tool invocations; a server present across enough sessions with zero invocations is dead weight paying a schema tax for no return. A deferred/ToolSearch listing lowers the estimated per-turn tax rather than charging the full schema cost. Registered as `deadweight`.
- **`context_resend.py`** — measures repeat context (`repeat_share = 1 - max(prompt_size) / sum(prompt_size)`, token-weighted), independent of whether `cache_control` was ever enabled. This is the structural gap `cache_efficacy` misses: cache efficacy reads 0.0 when caching was never turned on even if identical content is re-sent every turn. Registered as `resend`.
- **`batch_placement.py`** — flags cadence-regular, non-interactive workload groups as Batch-API (50% discount) candidates, using a conservative structural signal (low coefficient of variation on inter-start gaps, no human turn after the first model call). Advise-only: the card states that batch adoption is an architectural change, not a config flip. Runs inside the `downsize` lane; surfaced under the `placement` name in `tj optimize`.
- **`downsize_agents.py`** — per-agent price arithmetic for the downsize card: each agent's real token mix (input, output, cache read, and cache write) priced at the current model and repriced at the proposed model through `core.pricing`/`core.cost`, with a 30-day projection. It does not measure whether the cheaper model gives an equivalent answer, and the card says so.

### Proposal and apply machinery (`tokenjam/core/optimize/`)

- **`relearn_proposals.py` / `relearn_store.py`** — the detector's clusters ARE the proposals. Each gets a stable `rp_`-prefixed ID that is a pure function of its signature, so it survives recomputes. Apply can only name a stored proposal; a caller cannot hand-build cluster content and aim a write wherever it likes. The full-corpus detector result is cached to `relearn_cache.json`, computed by a `tj serve` background job on a fresh DuckDB connection, so no HTTP request pays the full scan.
- **`relearn_apply.py`** — the apply stage. Routes an approved proposal to an intervention rung: rung 1 appends a marked section to a `CLAUDE.md`; rung 2 writes a new `.claude/skills/<slug>/SKILL.md`; rungs 3-5 write an enforcement hook plus a staged `settings.json` patch, always DISABLED until a human calls enable. Every write takes a pre-image snapshot, is git-committed when inside a repo, and is recorded to an `applied_fixes.json` ledger; `revert_applied_fix` restores the pre-image. Nothing writes unless the caller passes `go=True`; the default is a dry run returning the would-be diff.
- **`relearn_otel.py`** — a second extraction path for agents that leave no transcript (plain OTel services): it reads failing spans from the `spans` table into the same failure-episode shape, so clustering and proposal building are reused. These clusters are advise-only with no apply target; only non-coding agents are read here, since coding agents already come through the transcript path.
- **`cost_proposals.py` / `cost_apply.py` / `model_apply.py`** — adapts each cost analyzer's finding into a typed `CostProposal` so the inbox can list it beside the relearn proposals. Most cards are advise-only (the fix lives in the user's own application code; the printed snippet IS the fix); a minority are apply-capable. `cost_apply.mark_applied` records the user's change as a `core.loop` `Expectation` marker plus a `cost_applied.json` ledger entry (no code write of its own), and `revert_applied` undoes the record. `model_apply` performs the two deterministic model-id writes: `agent_model` rewrites the `model:` key in a `.claude/agents/<name>.md` frontmatter, and `model_swap` replaces one exact model-id string with another in a registered git repo; both inherit the snapshot/commit/revert machinery.
- **`accounting.py`** — shared token/dollar primitives: `four_type_token_sum_sql` (input + output + cache-read + cache-write, always) and call-identity dedupe so a call seen by both the live and backfill paths is counted once. **`clustering.py`** extracts the grouping/threshold/masking plumbing that `relearn`, `plan_reuse`, and `workflow_restructure` had each hand-rolled.

### Shared core primitives

- **`atomic_write.py`** — the atomic-write primitive relocated to a neutral module, reused by `summarize.apply` and `relearn_apply`; refuses a symlinked target for every caller.
- **`transcript_cache.py`** — an opt-in per-file transcript parse cache keyed on `(path, size, mtime)`, the same staleness signal the relearn and deadweight analyzers already trust. `core.transcript.read_records` still defaults to always-reparse unless a `cache_dir` is passed.
- **`server_state.py`** — a `tj serve` state/PID file so `tj stop` (and through it `reset`/`uninstall`) scopes to THIS install's daemon via `Path.home()`, instead of a bare `pgrep -f` that would match every `tj serve` on the machine.

### CLI

- **New `tj relearn` group** (`cmd_relearn.py`, `relearn_write_verbs.py`, `cost_proposal_verbs.py`) — the terminal path a headless or display-less install previously lacked. Verbs: `list`, `apply <proposal_id>` (dry run; writes only with `--go`), `enable <fix_id> --yes`, `revert <fix_id>`, `eval-case <proposal_id>` (emits the clustered evidence as JSON for your own eval tooling), plus the cost verbs `cost-proposals`, `cost-apply`, `cost-mark-applied`, `cost-revert`. Every verb is a thin wrapper over the same core functions the API route calls, so CLI and UI cannot drift into two meanings of "apply".
- **Global `--json` fix** (`json_option.py`) — `tj --json <cmd>` and `tj <cmd> --json` are now equivalent on every command that supports JSON, instead of a local flag silently ignoring the global one.
- **`cmd_optimize.py`** — CLI renderers for findings that previously lived only in the web inbox: context-resend, relearn, deadweight, cache root-cause candidates, the cache-recommend `cache_control` snippet, trim source-file provenance, and the batch-placement card. `tj optimize placement` is accepted and resolves to running `downsize`.
- **Removed** — `tj hook`, `tj savings` (with `core/output_cap.py`, `cli/cmd_savings.py`, and `core/savings_log.py` trimmed to its one surviving helper), and `quickstart` as a typeable command. Bare `npx tokenjam` now routes to the zero-install report through the `TJ_NPX_ZERO_INSTALL_REPORT` env var the npm wrapper sets, with no public subcommand name.

### API

- **`routes/relearn.py`** — `GET /relearn/proposals` and `/cost-proposals`, `POST /relearn/apply`, `/{id}/enable`, `/{id}/disable`, `/{id}/revert`, `GET /applied`, `POST /refresh`. Reads the warm cache and never computes the finding inline on a request; `/apply` names a stored proposal and refuses caller-supplied cluster content; a refused write becomes HTTP 409.
- **Local write auth** (`deps.py::require_relearn_write_auth`, `app.py`) — a random per-process token minted in `create_app`, embedded only in the same-origin UI HTML as `<meta name="tj-write-token">` and echoed back as `X-TJ-Local-Token`. It guards every mutating relearn endpoint unconditionally, independent of `api.auth.enabled`, so a default (auth-disabled) local install is not left open to a cross-origin CSRF POST at `127.0.0.1`.
- **`routes/cost.py`** — the recoverable headline is now honest about overlap: `total_recoverable_usd` is labeled a gross ceiling across overlapping analyzers (`recoverable_additive: false` plus an overlap note), and a single `largest_recoverable_*` is surfaced as the one standalone-honest claim.

### Config (`core/config.py`)

- New `[optimize]` block (`OptimizeConfig`) exposing per-analyzer sensitivity thresholds, each defaulting to the module constant it replaces so an unset section changes nothing: `min_sessions_deadweight` (default 5, lowered from 10), `cache_efficacy_threshold` (0.30), `min_recurring_sessions` (3), `min_sessions_for_cadence` (5), `trim_significance_threshold` (0.40), and related cluster/cohort/prefix/cost floors.
- New `[loop]` block (`LoopConfig`) with `transcript_path`, so a Claude Agent SDK app can point the detector at its transcript root.
- `capture.prompts` and `capture.tool_inputs` now default to **true** (previously off), because cache-recommend, trim, and the clustering analyzers are dark without prompt text and tool-call arguments.

### UI (`tokenjam/ui/index.html`)

Review inbox split into Improve and Observe lenses; Optimize detail cards for context-resend, batch-placement, and the cost rollups, each linking to its Review-inbox fix; a tri-state select-all in the Review inbox; a Sessions sidebar entry; budget-projection reframed as a forecast rather than recoverable waste.

## Test plan

New and expanded unit + integration suites cover each surface:

- Detector and apply: `tests/unit/test_relearn.py`, `test_relearn_apply.py`, `test_relearn_cli.py`, `test_relearn_otel.py`, `test_onboard_relearn_activation.py`, `tests/integration/test_relearn_api.py`.
- Cost proposals and applies: `tests/unit/test_cost_proposals.py`, `test_cost_proposals_model.py`, `test_cost_proposal_verbs.py`, `test_model_apply.py`, `tests/integration/test_cost_proposals_api.py`.
- New analyzers: `tests/unit/test_deadweight.py`, `test_deadweight_apply.py`, `test_batch_placement.py`, `test_context_resend.py`, `test_downsize_agent_pricing.py`, `test_optimize_clustering.py`.
- Flags, capture defaults, daemon scoping, caching: `tests/integration/test_global_json_flag.py`, `test_onboard_capture_default.py`, `test_doctor_capture_prompts.py`, `test_cmd_stop_scoping.py`, `test_transcript_cache.py`.
- Token-sum safety net: `tests/token_aggregate_guard.py` asserts any multi-column `SUM(...)` includes all four token types (input, output, cache read, cache write), the omission that has shipped repeatedly in this package.

Removed with the verify/receipts layer: `test_receipts.py`, `test_relearn_verify.py`, `test_relearn_otel_verify.py`. Removed with the retired commands: `test_cmd_hook.py`, `test_output_cap.py`, `test_savings_log.py`.

Run: `make test` (`pytest tests/`) with `mypy` type-checking.


anilmurty (commented):
Thanks @anshss — this is a lot of thoughtful work and the shape is right: detect → propose → approve → apply → verify → compound is exactly the loop we want, and the honesty discipline is genuinely well-held (every "tokens saved" carries an `est.` tag + "estimated/correlational, never a causal claim", verdicts are observation labels, nothing overclaims). I did a close review — strong draft, a few things to land before merge.

### Must-fix before merge

1. **Auth on the mutating endpoints.** `/pothole/apply`, `/enable`, `/disable`, `/revert`, `/refresh` (`routes/pothole.py`) are guarded only by `Depends(require_api_key)`, which is off by default (`api.auth.enabled=false`). These are the first HTTP endpoints in the product that write files + `git commit` (skills, hooks, `settings.json` patches) — every prior write path went through ingest-auth. On a default local install they're unauthenticated, so any local process (or a CSRF POST to `localhost:7391`) can drive apply→enable; the human-in-the-loop guarantee is a UI convention, not enforced server-side. Please gate these behind real auth regardless of the global flag, and/or constrain `target_path` to a home-anchored allowlist.

2. **Rung-1 note skips the overwrite guard.** In `pothole_apply.py` (~L801) the "refusing to overwrite a file TokenJam didn't write" check only covers rungs 2/3 (skill/hook); `RUNG_NOTE` renders + appends unconditionally. Since `target_path` is client-supplied and the note rung applies live (no enable gate), an apply with `rung=1, target_path=<any file>` appends a `<!-- tokenjam:pothole -->` block into it (into a `.py`/`.zshrc` → corruption) and git-commits it. It's backed up/revertible, but please apply the same non-TokenJam guard, or allowlist note targets to `CLAUDE.md`/`*.md`.

3. **No symlink guard on the write primitive.** The `_atomic_write` you reuse from summarize keeps its symlink/owner checks in the *callers* (`apply_staged`), not the primitive — so `_build_write_plan`/`_restore_backup` don't defend a symlinked `target_path`. Mirror the summarize guard (`if target.is_symlink(): refuse`).

4. **Config-path bug is broader than the PR notes.** Beyond `pothole_store.default_cache_path()` ignoring `--config`, the apply/ledger/backup tree in `pothole_apply.py` falls back to the real `~/.tj` when `storage.path` is `""`/`:memory:` — so InMemory-backed tests write into the real `~/.tj`, and an apply-time vs revert-time config mismatch can silently break revert (revert reads the ledger by that path). Thread `config` through and send the `:memory:` fallback to a temp dir.

### Polish (non-blocking)
- Ledger headline says "realized savings" (`index.html` ~L7512) — "realized" reads as booked actuals; swap to "estimated savings" to match the `est.` tag everywhere else. Same for the internal comment "evidence that it actually made things leaner" (~L7502) → "correlates with fewer recurrences".
- The Dashboard "Pending fixes" health tile is hardcoded `value=0 / coming soon` (`index.html` ~L4885) — wire it to `/pothole/proposals` or drop it until it's live.
- `PotholeFinding` is tokens-only (no `estimated_recoverable_usd`) — defensible for an agent-pothole, but it means it won't appear in the registry-driven Overview waste band; worth a line in the PR body confirming that's intended.

### Also
The branch is in conflict with `main` (likely `runner.py`, `index.html`, `app.py`, `test_lens_ui_regression.py`) — needs a rebase, then a re-run of the UI regression/offline tests + the three new pothole test files.

Rule-compliance otherwise checks out: self-registration (Rule 16), offline UI (Rule 18 — no external render-time URLs), `core/` purity (no cli/api imports), single-compute-path, and the UI regression guards are **updated, not weakened**. The fail-open hook templates and the disabled-by-default enable gate are exactly right. Solid direction — let's get the apply-surface hardening + rebase in and it's close.

PR #483, opened 2026-07-13
fix(helicone): handle NDJSON inputs correctly

### LINKED ISSUE #351: Helicone adapter crashes on NDJSON input (inverted detection condition)
**Severity:** Low
**Where:** `tokenjam` sub-repo — `tokenjam/core/ingest_adapters/helicone.py` (`_load_file`, ~line 248)

**Observed:** The NDJSON gate is `if text[0] == '{' and '\n{' in text and not text.startswith('{"'):`. Real NDJSON lines are compact objects starting with `{"`, so `text.startswith('{"')` is True and `not …` is False — the NDJSON-handling branch is skipped and control falls to the final `return json.loads(text)`, which can't parse a multi-object document. The exception is uncaught (the call site has no try, and `json.JSONDecodeError` is a `ValueError`, not the `RuntimeError` the CLI catches), so `tj backfill helicone --source-file dump.ndjson` aborts with a traceback — despite the docstring advertising NDJSON support. Reproduced on a 2-line NDJSON dump.

**Fix direction:** Replace the fragile prefix heuristic with the OTLP adapter's pattern: try whole-file `json.loads(text)` first, then on `JSONDecodeError` fall back to line-by-line parsing. (Equivalently, drop the inverted `and not text.startswith('{"')` clause.)

**Done when:** `tj backfill helicone` ingests a standard NDJSON dump without error; a test covers NDJSON input.

tarun73:
## Summary

This PR resolves an issue where the Helicone ingest adapter (`_load_file`) would crash with a `JSONDecodeError` when attempting to load valid NDJSON input. The previous logic used an incorrect string-matching
heuristic that effectively bypassed the NDJSON parsing block.

### Changes:
- **`helicone.py`**: Refactored the `_load_file` detection logic. Instead of relying on `text.startswith()`, the code now attempts a standard `try...except json.JSONDecodeError` on the entire file first, seamlessly
falling back to a line-by-line parser for NDJSON inputs.
- **`test_ingest_helicone.py`**: Added `test_ingest_handles_ndjson` to ensure line-by-line JSON payload formats are successfully parsed and routed correctly.

## Related issue

Closes #351

## Checklist

@anilmurty please review

### DIFF
--- a/tests/unit/test_ingest_helicone.py
+++ b/tests/unit/test_ingest_helicone.py
@@ -260,3 +260,30 @@ def test_record_to_span_cache_write_absent_leaves_field_none():
     span = _record_to_span(record)
     assert span is not None
     assert span.cache_write_tokens is None
+
+
+def test_ingest_handles_ndjson(db, tmp_path):
+    """NDJSON format (line-by-line JSON objects) is accepted."""
+    record1 = {
+        "request": {
+            "id": "req-ndjson-1",
+            "created_at": "2026-04-01T10:00:00Z",
+            "model": "claude-haiku-4-5",
+            "provider": "ANTHROPIC",
+            "prompt_tokens": 200,
+        },
+    }
+    record2 = {
+        "request": {
+            "id": "req-ndjson-2",
+            "created_at": "2026-04-01T10:00:01Z",
+            "model": "claude-haiku-4-5",
+            "provider": "ANTHROPIC",
+            "prompt_tokens": 300,
+        },
+    }
+    path = tmp_path / "dump.ndjson"
+    path.write_text(json.dumps(record1) + "\n" + json.dumps(record2) + "\n")
+    result = ingest_helicone(db, source_file=str(path))
+    assert result["records_read"] == 2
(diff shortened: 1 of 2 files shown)


anilmurty (approved):
Thanks @tarun73 — clean fix, and it matches the fix direction in #351 exactly: try a whole-file `json.loads` first, fall back to line-by-line, and the inverted `startswith('{"')` heuristic is gone. Tests cover both valid NDJSON and the malformed-line path.

Heads up that the Greptile P2 above (unwrapped per-line `json.loads`) is already handled in your current diff — you wrap it and raise a descriptive `ValueError("Failed to parse NDJSON in {path} at line {i}")` with a test asserting the line number. Greptile reviewed an earlier commit.

One small non-blocking follow-up: per #351 the CLI catches `RuntimeError`, not `ValueError`, so a genuinely malformed file still surfaces as a `ValueError` rather than a clean CLI error — but the primary bug (valid NDJSON crashing) is fully fixed, so that's later polish at most.

LGTM — ready to merge.

PR #489, opened 2026-07-15
Quantify recoverable tokens for over-powered subagents in optimize ranking

anshss:
## Problem

The subagent right-sizing analyzer flags `over_powered` / `over_provisioned` subagent spawns — one of the most actionable findings for Claude Code users — but never filled `estimated_recoverable_usd` / `estimated_recoverable_tokens`. The ranked `tj optimize` report orders findings by `estimated_recoverable_tokens / window total`, so a finding with no estimate renders unranked. The subagent analyzer was therefore **structurally locked out of the numbered slots** no matter how much flagged spend it found.

## Fix

Derive a conservative recoverable estimate for `over_powered` subagents: the token-cost delta of routing each to its cheaper same-family model (reusing `lookup_downgrade` + the pricing tables) over the **identical token mix**. It's a pure model-swap counterfactual — no token-count change, no quality claim — consistent with the analyzer's honesty discipline:

`over_provisioned` is deliberately **excluded** — its recoverable is a prompt-size reduction we can't size honestly, so it never inflates the number.
A subagent with no cheaper alternative or no pricing data contributes **nothing** rather than inventing a figure.
When nothing `over_powered` has a priced alternative, the estimate stays `None` and the finding renders unranked (honest empty state), exactly as before.
`estimate_basis` / `estimate_confidence` carry the honest framing; the CLI renders an "estimated recoverable" line and the existing caveat still governs.

Per-subagent `provider` is now captured so the cheaper model can be priced; it round-trips through the report (de)serialization.

The finding now competes for a numbered slot on its own merits. On a CC-heavy dataset with significant flagged subagent spend it lands in slot ①.

## Tests

`test_over_powered_subagent_carries_quantified_estimate` — the model-swap delta math.
`test_over_provisioned_only_subagent_has_no_estimate` — the exclusion stays honest (estimate `None`).
`test_over_powered_estimate_ranks_in_numbered_slot` — the finding earns a numbered (major) slot.

All existing optimize unit tests pass (106); the full unit + synthetic suites are green except one pre-existing, unrelated onboarding test that reads a real local config.

### DIFF
--- a/tests/unit/test_optimize_subagent.py
+++ b/tests/unit/test_optimize_subagent.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 
 from datetime import timedelta
 
+import pytest
+
 from tokenjam.core.db import InMemoryBackend
 from tokenjam.core.optimize.analyzers.subagent_rightsizing import run as run_subagent
 from tokenjam.core.optimize.runner import report_from_dict, report_to_dict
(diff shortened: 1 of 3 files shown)


anshss:

anshss:

PR #490, opened 2026-07-15
Add release script and version-lockstep CI guards

anshss:
## Summary

Cutting a release currently requires hand-bumping three package version files (`pyproject.toml`, `sdk-ts/package.json`, `npm-wrapper/package.json`) in lockstep, with no automation and no failure signal until `npm publish` fails mid-release with a confusing "version already exists" error.

Add `scripts/release.sh X.Y.Z` (also runnable as `make release VERSION=X.Y.Z`) — bumps all lockstep-required files from one input and greps the repo for any straggler still referencing the old version string.
Add a `version-lockstep` job to `ci.yml` that fails every PR/push to `main` where `pyproject.toml` and `sdk-ts/package.json` versions disagree — catches drift well before a release is ever cut.
Add a pre-publish guard to `publish-npm.yml`'s `publish-npm` job that checks `sdk-ts/package.json`'s version against the release tag *before* installing/building/publishing anything, so a mismatch fails fast instead of mid-publish.
Update `CLAUDE.md`'s release docs to point at the new script/guards instead of the old manual-bump instructions.

`npm-wrapper/package.json` is auto-synced from the release tag by CI (`publish-npm-wrapper` job) so it was already safe from this class of bug — the script still bumps it for local `npm pack` hygiene.

## Test plan
[x] Ran `./scripts/release.sh 0.5.8` locally — confirmed it bumps all three files with a minimal diff (no incidental reformatting), then reverted.
[x] Confirmed error paths: same-version input, malformed version string, no args.
[x] Simulated both new CI checks locally against matching and mismatched version states — both pass/fail as expected.
[x] Validated new/edited workflow YAML with `yaml.safe_load`.
[x] `ruff check tokenjam/` still passes.


anshss:

anshss:

PR #491, opened 2026-07-15
fix(mypy): re-enable bug-catching error codes disabled in mypy config

anshss:
## Problem

The `[tool.mypy]` `disable_error_code` list globally suppressed whole classes of real bugs across the entire repo:

```
disable_error_code = ["attr-defined", "misc", "method-assign", "call-overload", "arg-type", "assignment", "no-any-return"]
```

Because these codes were off repo-wide, "mypy clean" certified much less than it appeared. Concretely, unpacking a `tuple[...] | None` without narrowing — mypy's `"None" object is not iterable`, code `misc` — passed the typecheck, as did attribute typos (`attr-defined`), wrong argument types (`arg-type`), and bad assignments (`assignment`). The type checker was silently decorative for these bug classes.

## Change

Re-enabled `misc`, `arg-type`, `assignment`, `call-overload`, and `method-assign` by burning down **every** resulting violation (83 across ~28 files) with real fixes:

**Narrowing guards / correct annotations** for `X | None → X` sites, over-broad `object`/`Any` locals, and `Literal` params (the large majority).
**Honest `Any` typing** for the SDK integrations that monkeypatch genuinely-untyped third-party callables (`anthropic`, `openai`, `litellm`, `langchain`, `langgraph`, `gemini`, `crewai`, `autogen`) — the `self._original_* = None` initializers were forcing a `None` inferred type.
**A few documented `cast(...)`s** at untyped-dependency boundaries (DuckDB row cells, out-of-scope model signatures, an OTLP exporter's duck-typed interface), each with a reason comment.
**One targeted `# type: ignore[assignment]`** for the interchangeable grpc/http OTLP exporter import, with justification.

Only `attr-defined` and `no-any-return` remain globally disabled — now with an **inline justification** in `pyproject.toml` and a per-code burn-down note, so the remaining suppression is deliberate and documented rather than silent.

These are type annotations only; **no runtime behavior changed** (e.g. the one constant swap resolves to the same value the dataclass field defaulted to).

## Validation

`mypy tokenjam/` → **Success: no issues found in 195 source files** (with the tightened config).
`ruff check tokenjam/` → **All checks passed.**
`pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` → **2504 passed, 2 skipped.** The single remaining failure (`test_bare_onboard_sdk_choice_falls_through_to_generic`) is a **pre-existing** test-isolation issue (a real `~/.config/tj/config.toml` leaks past the test's `isolated_filesystem`); it reproduces identically on the untouched baseline and is unrelated to this change.


anshss:

anshss:

PR #492, opened 2026-07-16
Fix: add cache_write_tokens to workflow_restructure per-cluster totals

anshss:
## Summary

The workflow_restructure analyzer (the "script" finding) was omitting `cache_write_tokens` from its per-cluster token-recovery calculation. This is the third occurrence of this bug pattern in the optimize/ directory, following fixes in cache_efficacy (#450) and runner.py (#468).

## Issue

When clustering deterministic workflows, the analyzer computes `estimated_recoverable_tokens` by summing session tokens within each cluster:

```sql
COALESCE(SUM(input_tokens + output_tokens + cache_tokens), 0)
```

The `cache_tokens` field represents cache *reads* (reused context); `cache_write_tokens` represents cache *creation*. The analyzer was missing the latter, causing cache-write-heavy workloads to underreport their recoverable token savings.

## Fix

**Line 192** (`workflow_restructure.py`): Add `cache_write_tokens` to the sum, matching the pattern established by PR #450 and #468
**Verify no other similar omissions**: Confirmed runner.py already includes both cache token types correctly
**Regression test**: New test (`test_cache_write_tokens_included_in_recoverable_total`) seeds 20 sessions with cache writes and verifies the calculated total is accurate

## Testing

✅ All 14 workflow_restructure tests pass
✅ All 18 optimize module tests pass
✅ Linting passes (ruff check)
✅ Type checking passes (mypy)

## Verification

The regression test verifies that a cache-write-heavy cluster (20 sessions × 2000 tokens each including 300 cache_write per session) correctly recovers 40,000 tokens, not the underreported 34,000 that would result from omitting cache_write_tokens.


anshss:

anshss:

PR #493, opened 2026-07-16
Make the statusline /compact nudge driver-conditional

anshss:
The zero-token statusline names the top re-read driver once re-reading passes the WARN threshold, but the remedy it suggested was still purely threshold-based (`>=70%` -> "consider /compact", `>=88%` -> "/compact to reclaim quota") — identical for every driver.

That advice is factually wrong for the dominant driver class. `/compact` shrinks conversation history only, so it cannot reduce **statically re-injected** content — CLAUDE.md, `@file` reads, re-run searches — which is re-sent every turn regardless of compaction. The line could render `re-read 88% (CLAUDE.md ×14)` right next to a remedy that can't touch CLAUDE.md re-reads at all. It's also the wrong trade for a user mid-window who values session memory: compaction is lossy, quota already spent isn't refunded, and any savings are future-only.

The root enabler was that the attribution cache collapsed the classified inclusion type down to a bare label string, so the statusline consumer couldn't condition its remedy even though the diagnostic already knows the type.

## Summary
Carry `inclusion_type` through `write_attribution_cache` / `refresh_attribution_cache`; add `format_driver` returning `(label, inclusion_type)` as the single cache reader. `format_driver_label` stays a thin label-only wrapper for the resume-brief, so **label rendering is unchanged**.
Make the nudge driver-conditional:
**static driver** (`file_read` / `search`) -> a structural remedy only (`→ trim re-reads: see tj context`), **never** `/compact`;
**history-bloat driver** (`tool_output` / `prompt`), or driver unknown -> a memory-preserving fresh session first (`→ fresh session (resume-brief) or /compact`), with `/compact` as the blunt second;
**window genuinely near its limit** — derived from the last turn's context occupancy, with a `[1m]` 1M-context guard to avoid false positives — -> a direct `→ /compact now (window near full)`, where a user-chosen compaction beats a forced auto-compact.
Add `last_turn_context_tokens` to `core.usage`; the statusline reads the transcript **once** for both the re-read share and the window occupancy, preserving the single-pass, zero-token, stdlib, fail-safe contract.

## Tests
New type-conditional tests in `test_statusline.py` (each driver kind, the near-limit override, the `[1m]` guard, the pre-upgrade cache-without-type fallback); the shipped threshold tests are updated to the new copy.
`test_attribution_cache.py` covers `inclusion_type` round-trip + `format_driver`; `test_usage.py` covers `last_turn_context_tokens`.
Verified end-to-end by driving the real `tj statusline` CLI for every driver type and both window regimes. `ruff` + `mypy` clean.


anshss:

anshss:

PR #494, opened 2026-07-16
Unify coding-vs-SDK agent_id classification

anshss:
## Summary
Two independently-maintained agent_id prefix lists decided coding-vs-SDK classification: the alert engine's `is_interactive_coding_agent` (`claude-code` / `codex`, no trailing hyphen) and the optimize-persona classifier in `framing.py` (`claude-code-` only, hyphenated, no codex).
They disagreed at the margins: a bare `claude-code` agent id, and every codex session — a pure-codex usage window fell through to the SDK-facing `tokenjam-bench` call-to-action in `tj optimize` instead of the interactive-coding-agent CTA.
`framing.agent_persona_mix` now calls the alert engine's `is_interactive_coding_agent` instead of keeping its own prefix check, so alert gating, drift detection, the `/status` route, and the `tj optimize` persona CTA all agree on one classifier.

## Test plan
[x] Added parametrized tests pinning the margin cases (bare `claude-code`, bare `codex`, `codex-*`) for `is_interactive_coding_agent`.
[x] Added a regression test asserting `agent_persona_mix` agrees with `is_interactive_coding_agent` on those same margin cases.
[x] `pytest tests/unit/test_framing.py tests/unit/test_alerts.py tests/unit/test_drift.py` — 72 passed.
[x] `pytest tests/unit/` — full suite green except one pre-existing, unrelated failure caused by a stray `~/.config/tj/config.toml` on this host (reproduces identically on `main`, before this change).
[x] `ruff check` / `mypy` clean on touched files.


anshss:

anshss:

PR #496, opened 2026-07-16
test: isolate suite from the real ~/.tj / ~/.config/tj store

anshss:
The test suite could take an exclusive DuckDB lock on the developer's real `~/.tj/telemetry.duckdb` and block on the real `~/.config/tj/config.toml` fallback, because both paths expand against the actual HOME unless a test explicitly overrides them. Most tests already pin an explicit `tmp_path`, but any that forget it hold the lock for the run's duration — colliding with a concurrently running `tj serve` / CLI ("Database is locked" in either direction).

## Summary
Add a session-wide, autouse `_tj_isolated_home` fixture in `tests/conftest.py` that repoints `HOME` (and `tokenjam.core.config.SEARCH_PATHS`, a module-level constant baked from `Path.home()` at import time — an env-var change alone can't retroactively fix it) at a throwaway tmp dir for the whole test run.
Add a second autouse `_tj_guard_real_home_db` fixture that wraps `DuckDBBackend.__init__` to raise loudly if any test still resolves a db path under the real `~/.tj` or `~/.config/tj` — a backstop for a test that forgets its own isolation.
Document both in CLAUDE.md.

## Tests / Verification
[x] `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` — 2506 passed, 1 skipped
[x] Confirmed the guard fires: constructing a `DuckDBBackend` pointed at the real home's `.tj/telemetry.duckdb` raises `RuntimeError` immediately
[x] **Live concurrency check**: started a real `tj serve` holding an exclusive lock on the actual `~/.tj/telemetry.duckdb` (confirmed via `lsof`), then ran the full suite concurrently — all 2506 tests passed, no lock errors in either direction, and the real DB's mtime/size were byte-for-byte unchanged afterward
[x] `ruff check tests/conftest.py` — clean


anshss:

anshss:

PR #497, opened 2026-07-16
docs(config): mark output_cap hook measured-negative — do not enable

anshss:
## What

The `[hooks.output_cap]` PostToolUse output-trim hook ships **default-off** and is now conclusively a dead lever. This PR documents that disposition prominently so nobody enables it, without changing any behavior.

## Why

Two independent measurements killed the hook's thesis:

1. **The A/B gate failed.** Across 5 trials/arm (sonnet), the treatment arm cost **+5.6%** (median $0.833 vs $0.789) and cache-read +2.3% — run-to-run noise, no quota win. Claude Code already truncates Bash tool output to ~30 KB *before* the PostToolUse hook ever sees it, so the addressable bloat per call is tiny (~1,206 tok ≈ 0.06% of a ~2M-token cache-read footprint) and agents self-mitigate (`| tail`, `grep`, `Read`).
2. **Confirmed live 2026-07-03:** Claude Code *natively offloads* large tool output before it enters model context (a 980 KB Bash output → CC's own `tool-results/` file + a ~2 KB preview), so the hook has nothing left to trim.

So flipping `[hooks.output_cap].enabled = true` buys ~+5.6% cost for zero benefit — a latent footgun, not a real option.

## Changes (docstring/comments only — no behavior change)

Lead the `CapOutputConfig` docstring with a prominent **⚠ MEASURED NEGATIVE — DO NOT ENABLE** banner and fold in the CC-offload finding.
Strengthen the `enabled` field comment and the onboarding comment to match.
**Bug fix:** correct a stale parse comment that claimed the `[hooks.output_cap]` section is *DEFAULT-ON* while the code is (correctly) *DEFAULT-OFF*.

The hook code, config knob, and `tj hook cap-output` scaffold are all left intact and default-off — this is a disposition/documentation change only.

## Test plan

[x] `ruff check` clean on both changed files
[x] Modules import cleanly (docstring syntax verified)
[x] `pytest tests/unit/test_config.py test_output_cap.py test_cmd_hook.py test_onboard_cap_hook.py test_savings_log.py test_uninstall_wrapper.py test_config_secret_divergence.py` → 98 passed
[x] Onboard test group (`test_onboard_cap_hook / test_onboard_first_run / test_onboard_hygiene`) → 27 passed
[x] `tj hook --help` still lists the `cap-output` scaffold


anshss:

anshss:

PR #498, opened 2026-07-16
test: isolate onboard test from real config file

anshss:
## Summary

Fixes a test isolation bug where `test_bare_onboard_sdk_choice_falls_through_to_generic` would fail if the developer has a real `~/.config/tj/config.toml` file on their machine.

## The Problem

The test was using `CliRunner().isolated_filesystem()` but the config discovery logic was still finding the real home directory's config file. This happened because:

1. `SEARCH_PATHS` in `tokenjam/core/config.py` includes `Path.home() / ".config" / "tj" / "config.toml"`
2. `SEARCH_PATHS` is evaluated at module import time (not at runtime)
3. When the test ran, `find_config_file()` would find the real config and fail

## The Solution

Monkeypatch `SEARCH_PATHS` to use a temporary directory's path instead of the real home directory. This isolates the config search to the test's isolated filesystem.

## Test Results

Test passes when no real config exists
Test passes when a real config exists at `~/.config/tj/config.toml`
All 23 tests in the test file pass

### DIFF
--- a/tests/unit/test_onboard_path_branch.py
+++ b/tests/unit/test_onboard_path_branch.py
@@ -89,9 +89,11 @@ def test_bare_onboard_routes_to_selected_path(_routing_stubs, answer, key):
     assert set(_routing_stubs) == {key}
 
 
-def test_bare_onboard_sdk_choice_falls_through_to_generic(_routing_stubs, monkeypatch):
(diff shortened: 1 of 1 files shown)


anshss:

anshss:

PR #499, opened 2026-07-16
Render SDK service window cost next to the cost/min sparkline

anshss:
## Summary
`sdk_services[].window_cost` (cost summed over the same 24-minute sparkline window as `cost_per_min`) was computed server-side but never read by the dashboard.
Surfaces it as the label on the `cost / min` sparkline in the SDK services panel, e.g. `cost / min · $0.0680 last 24m`, so the per-minute chart shows its own total alongside `today_cost`.

## Test plan
[x] `pytest tests/unit/` — 1949 passed, 2 skipped
[x] `ruff check tokenjam/` — clean
[x] New static regression test (`test_sdk_services_panel_renders_window_cost`) pins the field's presence in `SdkServicesPanel`
[x] Seeded a local DuckDB with a live SDK-service span series, ran the real FastAPI app under uvicorn, and drove the actual dashboard in a headless browser — confirmed `cost / min · $0.0680 last 24m` renders next to the sparkline


anshss:

anshss:

PR #500, opened 2026-07-16
Add `tj demo --live` to replay incident scenarios into a running tj serve

anshss:
`tj demo <scenario>` injects realistic synthetic SDK-shaped spans, but only into a throwaway in-memory backend — the output never reaches the on-disk DuckDB that `tj serve` reads. So the live dashboard's SDK-services zone can only be populated by a live agent or hand-rolled OTLP posts, and the demo scenarios can't dogfood or demo that zone.

This adds a `--live` flag that replays the scenario's spans through the **real** ingest path.

## Summary
`tj demo <scenario> --live` serializes each scenario span to OTLP JSON and POSTs it to a running `tj serve`'s `/api/v1/spans`, exactly as the SDK's `TjHttpExporter` does — so the demo populates the same DuckDB the dashboard reads.
Without `--live`, behaviour is unchanged (throwaway in-memory backend).

## How it works
**`tokenjam/demo/live.py`** — a `LiveSink` that buffers the scenario's spans and POSTs them to serve in one batch, plus a `NormalizedSpan → OTLP JSON` serializer that is the inverse of `otlp_parsing.parse_otlp_span`. Because the round-trip reproduces an equivalent span, the server-side cost/alert/drift hooks fire on the receive path just as they do for real SDK traffic. Spans are snapshotted **before** local processing so captured content (e.g. `gen_ai.tool.input`) survives for server-side retry-loop detection.
**`DemoEnvironment`** auto-resolves the sink from the Click context, so every scenario picks it up with no per-scenario wiring.
**`cmd_demo`** pre-checks that a `tj serve` is reachable (fails fast with actionable guidance otherwise), runs the scenario, then flushes + reports the replay. In `--json` mode the replay summary is routed to stderr so stdout stays a single clean JSON object.

The flag is documented in `tj demo --help`.

## Tests / Verification
`tests/unit/test_demo_live.py` (10) — serializer round-trip, sink buffer/flush, Bearer-header handling, 401 / connect-error paths, serve health probe.
`tests/integration/test_demo_live.py` (4) — drives the real CLI `--live` into a real `build_default_pipeline` serve app (via ASGI transport) and asserts the spans, retry-loop alert, and recomputed cost land in the server DB; asserts `--json` keeps stdout clean; asserts fail-fast when no serve is reachable.
Full local run: `tests/unit tests/synthetic tests/integration` → 2509 passed, 2 skipped; `tests/agents` → 10 passed. `ruff check tokenjam/` and `mypy tokenjam/` clean.
**Live end-to-end:** started a real `tj serve`, ran all three scenarios with `--live`, and confirmed the dashboard's SDK-services panel (`Status` view) renders `demo-surprise-cost` ($1.37/day), `demo-retry-loop` (83.3% err), and `demo-hallucination-drift`, with Recent Alerts showing the retry-loop + failure-rate alerts.

## What's NOT in this PR
**Docs prose** (`docs/cli-reference.md` has a `tj demo` section): left for maintainer review rather than edited here. Suggested one-line addition under that section's example block:
`tj demo retry-loop --live # replay into a running tj serve (lights up the SDK-services zone)`
**The `hallucination-drift` `drift_detected` alert does not fire on the live path.** The scenario tunes drift config locally (`baseline_sessions=5`) via the demo's in-process `agent_configs`; those are server-side config, not span attributes, so they don't travel over the wire. The live daemon evaluates drift with its own config (production defaults). This is correct for the real ingest path — the spans still populate the dashboard. To see the drift alert fire live, configure `[agents.demo-hallucination-drift.drift]` in the tj config.


anshss:

anshss:

PR #504, opened 2026-07-19
Fix #345: honor TJ_CONFIG in pricing-override config discovery

### LINKED ISSUE #345: Pricing-override config discovery ignores `TJ_CONFIG`, silently dropping user rate overrides
**Severity:** Low
**Where:** `tokenjam` sub-repo — `tokenjam/core/pricing.py` (`_config_pricing_section`, ~line 117)

**Observed:** `_config_pricing_section()` calls `find_config_file()` with no argument to locate the config whose `[pricing]` section supplies rate overrides. `find_config_file()` with no override only walks the default search paths (`tokenjam.toml` → `.tj/config.toml` → `~/.config/tj/config.toml`) and never reads the `TJ_CONFIG` env var — that resolution lives only in `load_config()`. So when a user sets `TJ_CONFIG` to a non-default file, the rest of the app loads that file but the pricing loader resolves a *different* file (or none). Their `[pricing.<provider>]` / `[pricing.models]` overrides are silently ignored and costs are computed from the wrong rate table without warning.

**Fix direction:** Resolve the config path the same way `load_config` does — pass `os.environ.get("TJ_CONFIG")` into `find_config_file(...)` inside `_config_pricing_section` (or share a single discovery helper) so the `[pricing]` section is read from the same file as the rest of the app.

**Done when:** With `TJ_CONFIG` pointing at a config that carries `[pricing]` overrides, those overrides take effect in cost calculations.

sanmaxdev:
`_config_pricing_section()` called `find_config_file()` with no override, so pricing discovery only walked the default search paths and never read `TJ_CONFIG`, while `load_config()` resolves the env var first. With `TJ_CONFIG` pointing at a non-default config, the app loaded that file but the pricing loader resolved a different file (or none), silently dropping `[pricing]` rate overrides and computing costs from the wrong rate table.

## Summary
- Pass `os.environ.get("TJ_CONFIG")` into `find_config_file()` inside `_config_pricing_section()`, so the `[pricing]` section is read from the same file as the rest of the app.
- A `TJ_CONFIG` pointing at a missing file still degrades silently to packaged rates via the existing `FileNotFoundError` catch.

## Related issue
Closes #345

## Tests / Verification
- New regression tests in `tests/unit/test_pricing_override.py`: overrides take effect from a `TJ_CONFIG` file, `TJ_CONFIG` wins over search-path discovery, and a nonexistent `TJ_CONFIG` path degrades to packaged rates. Confirmed the new tests fail without the fix and pass with it.
- `pytest tests/unit/test_pricing_override.py -q` — 30 passed
- `pytest tests/unit/ -q` — 2001 passed, 1 skipped
- `pytest tests/synthetic/ tests/agents/ tests/integration/ -q` — 564 passed
- `ruff check tokenjam/core/pricing.py tests/unit/test_pricing_override.py` — clean
- `mypy tokenjam/core/pricing.py` — clean

## Checklist
- [x] Contributor docs updated (not needed, no architecture change)
- [x] Test spans use `tests/factories.py` (not applicable, config-layer change)

@anilmurty ready for review.

### DIFF
--- a/tests/unit/test_pricing_override.py
+++ b/tests/unit/test_pricing_override.py
@@ -21,6 +21,7 @@ def _isolate_pricing(monkeypatch, tmp_path):
     empty dir so a project-local tj.toml/.tj/config.toml [pricing] section
     can't leak in either (config discovery walks the cwd)."""
     monkeypatch.delenv(pricing.USER_PRICING_ENV, raising=False)
+    monkeypatch.delenv("TJ_CONFIG", raising=False)
     monkeypatch.setenv("HOME", str(tmp_path))          # POSIX
     monkeypatch.setenv("USERPROFILE", str(tmp_path))   # Windows Path.home()
     monkeypatch.chdir(tmp_path)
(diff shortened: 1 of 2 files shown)


anshss (approved):

anshss:
Thanks for this. Good catch on the asymmetry between the pricing path and load_config!

PR #511, opened 2026-07-22
refactor: extract shared token-accounting, clustering and transcript-cache helpers

anshss:
>

Extracts plumbing that several analyzers had each hand-rolled. Behavior-preserving.

- `accounting.py`: a `four_type_token_sum_sql` helper that always sums input + output + cache-read + cache-write, plus call-identity dedupe so a call seen by both the live and backfill paths counts once.
- `clustering.py`: the grouping/threshold/masking plumbing shared by the cross-session miners.
- `transcript_cache.py`: an opt-in per-file parse cache keyed on `(path, size, mtime)`.
- the atomic-write primitive relocated to a neutral module that refuses a symlinked target for every caller.

### DIFF
--- a/tests/integration/test_cost_charts.py
+++ b/tests/integration/test_cost_charts.py
@@ -196,3 +196,38 @@ async def test_components_empty_window_is_safe():
     assert d["total_cost_usd"] == 0
     assert d["recoverable"] == []
     assert "framing" in d
+
+
+@pytest.mark.asyncio
+async def test_components_never_invokes_the_full_corpus_relearn_scan(monkeypatch):
+    """`/cost/components` must not run `relearn`: its own docstring says HTTP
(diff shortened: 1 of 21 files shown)


anshss:

```
git rev-list --count origin/main..origin/refactor/shared-optimize-helpers   ->  0 commits missing
git diff --name-only origin/main...origin/refactor/shared-optimize-helpers  ->  0 files differ
```

**Why closed rather than merged:** this PR was stacked on `fix/tj-stop-scoping` rather than `main`, so its diff showed only its own slice for review. #482 carried every commit in the stack and brought them into `main` in one merge. GitHub only marks a PR merged when its head lands in *its own* base branch, and `fix/tj-stop-scoping` never received that merge, so this stayed open despite its commits being shipped. The commits are in `main` individually, with original authorship, because #482 was merged with a merge commit rather than a squash.

PR #513, opened 2026-07-22
feat: capture prompt text and tool inputs by default

anshss:
>

**Behavioral change worth explicit review.** Flips `capture.prompts` and `capture.tool_inputs` to default **true**.

Rationale: cache-recommend, trim, and the clustering miners are dark without prompt text and tool-call arguments. The consequence is that new installs begin capturing prompt content by default, which is why this is isolated in its own PR rather than folded into a larger change.

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -1998,12 +1998,16 @@ class _StubProvider:
         return self._resp
 
 
-def test_validate_capture_off_fails_with_hint(runner, db, config):
-    """Capture is off by default -> actionable failure naming the config toggle,
-    and no provider client is ever constructed."""
-    _seed_validate_span(db)  # capture flag in `config` is off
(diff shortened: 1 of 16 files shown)


anshss:

```
git rev-list --count origin/main..origin/feat/capture-prompts-by-default   ->  0 commits missing
git diff --name-only origin/main...origin/feat/capture-prompts-by-default  ->  0 files differ
```

**Why closed rather than merged:** this PR was stacked on `feat/optimize-config-thresholds` rather than `main`, so its diff showed only its own slice for review. #482 carried every commit in the stack and brought them into `main` in one merge. GitHub only marks a PR merged when its head lands in *its own* base branch, and `feat/optimize-config-thresholds` never received that merge, so this stayed open despite its commits being shipped. The commits are in `main` individually, with original authorship, because #482 was merged with a merge commit rather than a squash.

PR #515, opened 2026-07-22
feat: add four cost analyzers (context-resend, MCP dead-weight, batch placement, agent downsizing)

anshss:
>

Adds four read-only cost detectors.

- **`context_resend`**: token-weighted repeat-context share (`1 - max(prompt_size)/sum(prompt_size)`), independent of whether `cache_control` was ever enabled. Covers the gap where cache-efficacy reads 0.0 because caching was never turned on, even while identical content is re-sent every turn.
- **`deadweight`**: MCP servers configured but never invoked across enough sessions, paying a schema tax for no return, plus an always-injected context-tax table.
- **`batch_placement`**: cadence-regular, non-interactive workload groups as Batch-API candidates. Advise-only; batch adoption is an architectural change, not a config flip.
- **`downsize_agents`**: per-agent current-vs-proposed model repricing across all four token types, with a 30-day projection. It does not measure whether the cheaper model answers equivalently, and the card says so.


anshss:

```
git rev-list --count origin/main..origin/feat/cost-analyzers   ->  0 commits missing
git diff --name-only origin/main...origin/feat/cost-analyzers  ->  0 files differ
```

**Why closed rather than merged:** this PR was stacked on `feat/global-json-flag` rather than `main`, so its diff showed only its own slice for review. #482 carried every commit in the stack and brought them into `main` in one merge. GitHub only marks a PR merged when its head lands in *its own* base branch, and `feat/global-json-flag` never received that merge, so this stayed open despite its commits being shipped. The commits are in `main` individually, with original authorship, because #482 was merged with a merge commit rather than a squash.

PR #516, opened 2026-07-22
feat: relearn, a cross-session recurring-failure detector with reversible apply

anshss:
>

Adds `relearn`, a cross-session detector for the failures an agent silently re-hits.

It treats every errored tool step as a failure episode and clusters episodes into signatures via validated regex families, with a bounded distill pass over the residual generics. Clusters already codified in a reachable `CLAUDE.md`/`learnings.md` are dropped. What recurs across three or more distinct sessions becomes a stored proposal with a stable `rp_`-prefixed id that is a pure function of its signature.

Apply routes an approved proposal to an intervention rung: a note appended to `CLAUDE.md`, a new skill file, or an enforcement hook staged **disabled** until a human enables it. Every write takes a pre-image snapshot, is git-committed when the target is tracked, and is reversible. Nothing writes without an explicit go; the default is a dry run returning the would-be diff. Apply can only name a stored proposal, so a caller cannot hand-build cluster content and aim a write anywhere.


anshss:

```
git rev-list --count origin/main..origin/feat/relearn-detector   ->  0 commits missing
git diff --name-only origin/main...origin/feat/relearn-detector  ->  0 files differ
```

**Why closed rather than merged:** this PR was stacked on `feat/cost-analyzers` rather than `main`, so its diff showed only its own slice for review. #482 carried every commit in the stack and brought them into `main` in one merge. GitHub only marks a PR merged when its head lands in *its own* base branch, and `feat/cost-analyzers` never received that merge, so this stayed open despite its commits being shipped. The commits are in `main` individually, with original authorship, because #482 was merged with a merge commit rather than a squash.

PR #517, opened 2026-07-22
feat: cost-proposal inbox for reviewable, reversible analyzer findings

anshss:
>

Adapts each cost analyzer's finding into a typed `CostProposal` so findings can be listed and acted on in one place.

Most cards are advise-only, because the fix lives in your own application code and the printed snippet **is** the fix. Marking one applied records the change plus a ledger entry rather than writing code. `model_apply` performs the two deterministic model-id writes: rewriting the `model:` key in an agent's frontmatter, and swapping one exact model-id string in a registered git repo, both through the same snapshot/commit/revert machinery.

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -1222,6 +1222,57 @@ def test_optimize_json_output_includes_caveat(runner, db, config):
     assert "Candidate-flagging heuristic" in data["downgrade"]["caveat"]
 
 
+def _seed_low_cache_efficacy_window(db, agent_id="test-agent"):
+    """12 sessions shaped to trip the `cache` analyzer -- the same fixture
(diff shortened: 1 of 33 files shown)


anshss:

```
git rev-list --count origin/main..origin/feat/cost-proposal-inbox   ->  0 commits missing
git diff --name-only origin/main...origin/feat/cost-proposal-inbox  ->  0 files differ
```

**Why closed rather than merged:** this PR was stacked on `feat/relearn-detector` rather than `main`, so its diff showed only its own slice for review. #482 carried every commit in the stack and brought them into `main` in one merge. GitHub only marks a PR merged when its head lands in *its own* base branch, and `feat/relearn-detector` never received that merge, so this stayed open despite its commits being shipped. The commits are in `main` individually, with original authorship, because #482 was merged with a merge commit rather than a squash.

PR #518, opened 2026-07-22
feat: surface relearn and cost proposals in the CLI and local web API

anshss:
>

Adds the entry points for everything above.

- A `tj relearn` command group: `list`, `apply` (dry run unless `--go`), `enable`, `revert`, `eval-case`, plus the cost verbs. Every verb is a thin wrapper over the same core functions the API route calls, so CLI and UI cannot drift into two meanings of "apply".
- CLI renderers for findings that previously existed only in the web UI: context-resend, relearn, deadweight, cache root-cause, the cache-recommend snippet, trim provenance, and batch placement.
- The `/relearn/*` routes, reading a warm cache so no HTTP request pays a full corpus scan.
- **Local write auth**: a random per-process token minted at app creation, embedded only in same-origin UI HTML and echoed back as a header. It guards every mutating endpoint unconditionally, independent of `api.auth.enabled`, so a default auth-disabled local install is not open to a cross-origin POST at `127.0.0.1`.
- Reframes the recoverable-dollar headline as a gross ceiling across overlapping analyzers rather than an additive total.


anshss:

```
git rev-list --count origin/main..origin/feat/proposals-cli-api   ->  0 commits missing
git diff --name-only origin/main...origin/feat/proposals-cli-api  ->  0 files differ
```

**Why closed rather than merged:** this PR was stacked on `feat/cost-proposal-inbox` rather than `main`, so its diff showed only its own slice for review. #482 carried every commit in the stack and brought them into `main` in one merge. GitHub only marks a PR merged when its head lands in *its own* base branch, and `feat/cost-proposal-inbox` never received that merge, so this stayed open despite its commits being shipped. The commits are in `main` individually, with original authorship, because #482 was merged with a merge commit rather than a squash.

PR #519, opened 2026-07-22
refactor: remove the measured-ineffective hook, savings and output-cap commands

anshss:
>

**Deletions, called out explicitly** rather than folded into a feature PR.

Removes surfaces measured as ineffective or superseded: the `hook` and `savings` commands, the `output_cap` module, and `quickstart` as a typeable command name. Bare `npx tokenjam` still reaches the zero-install report, now through an env var the npm wrapper sets rather than a public subcommand.


anshss:

```
git rev-list --count origin/main..origin/refactor/remove-dead-surfaces   ->  0 commits missing
git diff --name-only origin/main...origin/refactor/remove-dead-surfaces  ->  0 files differ
```

**Why closed rather than merged:** this PR was stacked on `feat/proposals-cli-api` rather than `main`, so its diff showed only its own slice for review. #482 carried every commit in the stack and brought them into `main` in one merge. GitHub only marks a PR merged when its head lands in *its own* base branch, and `feat/proposals-cli-api` never received that merge, so this stayed open despite its commits being shipped. The commits are in `main` individually, with original authorship, because #482 was merged with a merge commit rather than a squash.

PR #520, opened 2026-07-22
feat: Review-inbox UI with Improve/Observe lenses

anshss:
>

Splits the Review inbox into **Improve** and **Observe** lenses, adds Optimize detail cards for context-resend, batch-placement and the cost rollups that each link to their inbox fix, adds a tri-state select-all and a Sessions sidebar entry, and reframes budget-projection as a forecast rather than recoverable waste.


anshss:

```
git rev-list --count origin/main..origin/feat/review-inbox-ui   ->  0 commits missing
git diff --name-only origin/main...origin/feat/review-inbox-ui  ->  0 files differ
```

**Why closed rather than merged:** this PR was stacked on `refactor/remove-dead-surfaces` rather than `main`, so its diff showed only its own slice for review. #482 carried every commit in the stack and brought them into `main` in one merge. GitHub only marks a PR merged when its head lands in *its own* base branch, and `refactor/remove-dead-surfaces` never received that merge, so this stayed open despite its commits being shipped. The commits are in `main` individually, with original authorship, because #482 was merged with a merge commit rather than a squash.

PR #521, opened 2026-07-22
docs: internal docs and CLI reference for the new analyzers

anshss:
> **Stack slice of #482.** Based on the anchor branch, so the diff shows only this slice.
>
> Merge order: anchor #482 first, then this PR, then the public-copy PR that sits on top of it.

Internal documentation for the new analyzers and commands. No public-facing README copy (that lives in the repositioning PR stacked on top of this one) and no product code.

- `CLAUDE.md` and `docs/**`: architecture, configuration, CLI reference, getting-started, first-hour, harness-integration, installation, alerts, and the agent capability matrix updated for the new analyzers, the `tj relearn` command group, and the `[optimize]` config block.
- `npm-wrapper/bin/tj.js`: a file-header comment correction, so it describes the zero-install report rather than the retired `quickstart` routing.


anshss:
Heads up: this PR merged into its base branch `self-improve-loop`, not into `main`. Because `self-improve-loop` had already been merged into `main` before this merge, these 20 commits never reached `main` — verified: `docs/cost-repositioning` is not an ancestor of `main`, and `main`'s `docs/cli-reference.md` still has no `resend` entry.

No work is lost; the branch still holds every commit. They are being re-landed against `main` in a follow-up PR.

PR #522, opened 2026-07-22
docs: reposition public copy around cost [IN PROGRESS]

anshss:
> **Stack slice of #482.** Sits on top of the internal-docs PR #521, so the diff shows only the public copy.
>
> This is the last PR in the chain, which is deliberate: it carries `tests/unit/test_docs_analyzer_count.py`, a guard asserting every doc states the same analyzer count as the code registry. That test can only pass once all the code and all the docs are present, so it lives here.

Public-facing copy only.

- `README.md` and `npm-wrapper/README.md` repositioned to lead with cost and to name the mechanism plainly. The install headline stays npx-first; entry copy is no longer Claude-Code-only.
- `tests/unit/test_docs_analyzer_count.py`: the doc/registry consistency guard (verified passing on this branch).

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,9 @@ ceiling), and `cache-recommend` (the Cache card's breakpoint-suggestion half). R
 
 `tj serve` runs Lens at `http://127.0.0.1:7391/`. It opens on the **Improve** lens: a **Review inbox**
 where the self-improve loop's proposed fixes land for your approval, a Dashboard that lands you on
-recurring mistakes and current health, and Status. Flip to the **Observe** lens for Traces, Cost,
+recurring mistakes and current health, and Status. The downsize, cache, and trim analyzers file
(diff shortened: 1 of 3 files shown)


anilmurty:
Thanks @anshss. One blocker before this goes in: the hero leads with `npx tokenjam` under "See it in 30 seconds, no install." I want the public README to lead with `pipx install tokenjam && tj onboard` as the primary path — that's the full product — and keep the `npx tokenjam` no-install peek as a secondary "just looking?" line underneath. Peek-first undercounts installs and hides everything that only shows up after onboard.

Heads up too: #532 rewrites this same hero/persona block in the opposite (pipx-first) direction, so the two can't both land as-is. I'd take #532's framing as the base here and fold your section/analyzer updates on top. Can you reconcile the two into one coherent README?

PR #532, opened 2026-07-23
Analyzer count correction

anshss:
## Summary
- The first-run home screen and README table showed `tj onboard --claude-code` as a distinct "recommended" setup from bare `tj onboard`, describing the latter as "generic setup for the Python SDK." Both invoke the same interactive wizard; the flag only pre-answers the wizard's first question.
- Collapses the home-screen hint to a single `tj onboard` command, described as interactive setup, plus a note to run it once per project so sessions/proposals group correctly in the dashboard.
- Reframes the README "Which path are you?" table around `tj onboard` as the one entry point, with `--claude-code` / `--codex` noted as shortcuts rather than separate paths.
- Copy only; no behavior change to the onboard command itself.

## Test plan
- [x] `pytest tests/unit/test_onboard_first_run.py` — 17 passed
- [x] `pytest tests/unit -k "onboard or home"` — 270 passed


anilmurty:
Thanks @anshss — this is the framing I want: pipx-first, with `tj onboard` as the universal entry and `--claude-code`/`--codex` as shortcuts rather than a separate "recommended" path.

Two things before merge: it says "seven analyzers" but we're at thirteen now (downsize, cache, cache-recommend, resend, script, reuse, trim, subagent, summarize, verbosity, deadweight, relearn, plus budget-projection). And it overlaps the README hero with #522, which currently goes peek-first — let's fold those two together (keeping this PR's pipx-first hero). Fix the count, rebase on main, reconcile with #522, and I'll merge.

anshss:
> Thanks @anshss — this is the framing I want: pipx-first, with `tj onboard` as the universal entry and `--claude-code`/`--codex` as shortcuts rather than a separate "recommended" path.
> 
> Two things before merge: it says "seven analyzers" but we're at thirteen now (downsize, cache, cache-recommend, resend, script, reuse, trim, subagent, summarize, verbosity, deadweight, relearn, plus budget-projection). And it overlaps the README hero with #522, which currently goes peek-first — let's fold those two together (keeping this PR's pipx-first hero). Fix the count, rebase on main, reconcile with #522, and I'll merge.

thanks for spotting that! this pr was branched off main and contained its own wording, along with the update around removal of our flag, and that’s why the seven analyzer issue happened. the idea was to keep 522, the broader readme repositioning, on hold until i got more clarity and could get this to main first. but since you already recommended keeping this pr’s framing and folding the section/analyzer updates from 522 into it, it seems like a quicker fix, so i’m doing that!

PR #537, opened 2026-07-24
npx wrapper: throttle staleness checks, gate on success, bound pin-probe cost

anshss:
## Summary

Follow-up to the wrapper version-pinning + stale-shadowing-install-warning change already merged to main. Three refinements to that change's `npm-wrapper/bin/tj.js`, all in the same file:

- The stale-shadowing-install warning now only fires when the real invocation actually succeeded (exit status 0). Previously it could fire even when the underlying CLI itself failed (bad args, etc.), piling an unrelated advisory note onto a real failure.
- Added a 24h throttle around the staleness detector's four subprocess probes (uv/pipx/pip/brew). The check is purely advisory, so a bounded delay before the nudge appears is an acceptable tradeoff for not spawning four extra processes on every single invocation.
- The pinned-resolution probe is a genuine second process spawn on top of the real invocation (needed to confirm the pinned version resolves before committing to it). Rather than risk double-running the user's real command on a resolution failure, or breaking live stdout/stderr streaming for interactive subcommands by buffering output to inspect it, the probe's cost is now bounded via a cache marker: once a given runner+version pair is confirmed to resolve, later runs skip the probe entirely, so the double-invocation cost is paid roughly once per version per machine rather than on every run.

## Test plan

- [x] `node -c bin/tj.js` syntax check
- [x] Verified a failing real invocation (bad flag, non-zero exit) does not print the staleness warning
- [x] Verified a successful run prints the staleness note once, and an immediate repeat run is silent (throttled)
- [x] Verified with a PATH shim counting real `uvx` invocations that the resolution probe is skipped once the pin is confirmed (cold cache: 3 calls; warm cache: 2 calls)
- [x] Re-verified the unpinned-fallback path still works and does not falsely cache a marker on a failed resolution

### DIFF
--- a/npm-wrapper/bin/tj.js
+++ b/npm-wrapper/bin/tj.js
@@ -40,10 +40,23 @@
  * via `uv tool install` / `pipx install` / `pip install` / Homebrew, which can
  * sit on an old version indefinitely. See `warnIfShadowedByStaleInstall`
  * below: detect-and-tell only, never mutates, never auto-upgrades.
+ *
+ * Two on-disk markers (under the cache dir from `cacheDir()`) keep both of
+ * the above cheap on repeat runs, since `npx tokenjam` is meant to be run
+ * often, not once:
+ *   - `pin-ok-<bin>-<version>`: once the pinned spec has been CONFIRMED to
+ *     resolve for a given runner+version (see `resolves()`), skip the
+ *     confirmation probe on later runs for that same exact version — see the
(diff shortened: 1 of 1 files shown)


anshss:
Closing: this was opened as an unsolicited follow-up to an already-merged change. The branch `fix/wrapper-stale-detection-followups` is left in place if the findings are worth revisiting later.

PR #540, opened 2026-07-24
fix: use constant-time comparison for API key to prevent timing attacks

marceli1404:
## Changes
- Replace != operator with secrets.compare_digest() for API key comparison

## Why
The != operator short-circuits on the first differing byte, allowing timing side-channel attacks where an attacker can statistically recover the API key character-by-character. The same file already uses secrets.compare_digest() for the relearn write token.


anilmurty:
Welcome and thanks for the contribution @marceli1404 ! - reviewing now and should be able to merge this shortly

PR #542, opened 2026-07-25
fix(optimize): source cache-recommend's prefix signal from CLAUDE.md for Claude Code data

anshss:
## Summary
`cache-recommend` hashes `gen_ai.prompt.content` looking for a prefix repeated across >=3 calls, but for Claude-Code-sourced spans that field holds the human's per-turn message — a different string on every call — so the analyzer always produced zero candidates on this data even when a genuinely stable, cacheable prefix exists.
Claude Code's on-disk session transcript never records the actual system prompt it sends the API (verified against real transcripts), so there was no captured field to source the real prefix from.
The project's `CLAUDE.md` is resent verbatim on every call (the chat-completions API is stateless — the full system prompt goes out with every request), so the backfill parser now reads it straight off disk and stamps it on the span as a new `tokenjam.system_prefix.content` attribute, gated by the same `[capture] prompts` toggle as the existing prompt content. `cache-recommend` prefers this attribute when present, falling back to the previous prompt-content hashing for spans that don't carry it (e.g. raw Anthropic API telemetry).

## Test plan
[x] `pytest tests/unit/test_backfill.py tests/unit/test_cache_recommend.py -q` — 72 passed, including 3 new tests covering the CLAUDE.md capture and the Claude-Code-persona candidate path
[x] Full unit suite: `pytest tests/unit -q` — 2928 passed, 2 skipped, no regressions
[x] `ruff check` / `mypy` on all touched files — clean
[x] Live validation against a real local Claude Code project transcript + its real `CLAUDE.md` (isolated scratch DB, no other project data touched): before this fix `cache-recommend` produced 0 candidates on this data; after, it reports 1 candidate with 474 occurrences of the project's actual `CLAUDE.md` as the shared prefix

### DIFF
--- a/tests/unit/test_backfill.py
+++ b/tests/unit/test_backfill.py
@@ -937,6 +937,68 @@ def test_capture_flags_are_independent(tmp_path):
     assert GenAIAttributes.TOOL_INPUT not in tool.attributes
 
 
+def test_capture_prompts_reads_project_claude_md_as_system_prefix(tmp_path):
+    """#272: the human's per-turn message never repeats verbatim, so
+    cache-recommend's prefix-hash needs a different, genuinely stable
+    signal. The project's CLAUDE.md is read straight off disk (it's not in
+    the transcript) and stamped as `TjAttributes.SYSTEM_PREFIX_CONTENT` --
+    identical on every assistant span for the same project, unlike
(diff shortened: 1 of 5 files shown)


anshss:

PR #543, opened 2026-07-25
Put deadweight's recoverable estimate on the same window basis as every other cost analyzer

anshss:
`deadweight` folded a fixed 90-day projection directly into its shared `estimated_recoverable_usd`/`estimated_recoverable_tokens` fields, while every other cost analyzer keeps that field strictly window-scoped (`downsize` keeps its own 30-day projection in a separate `monthly_savings_usd` field for exactly this reason — the "Recoverable-savings contract"). The inflated field wasn't comparable to the other analyzers and corrupted any sum across them; the projection also grew unboundedly as the window shrank (a 7-day window showed roughly 12.9x the true window-observed figure).

## Summary

`deadweight` now emits a pure window-scoped `estimated_tax_tokens_window` / `estimated_tax_usd_window` (renamed from the old `*_90d` fields), with no internal projection folded in.
Introduce one central, shared projection instead: `compute_projection_ratio` derives a 30-day-pace ratio from the window's own distinct active-session days (`WindowSummary.active_days`, now computed in `summarize_window`), capped at 3x and guardrailed against thin windows/session counts — below the guardrails it falls back to the observed figure, unscaled, with a disclosure string, rather than inventing a ratio from too little data.
Every cost-proposal adapter now defers its `estimated_monthly_usd`/`estimated_monthly_tokens` to this one central pass instead of self-projecting. `downsize` used to copy its own `window_days`-based projection onto the shared monthly field, which was a milder version of the same defect (a different analyzer landing on a different basis than everyone else) — it's now scaled by the same central ratio as every other analyzer.
`CostProposal` gains a `scaling` field (`"per_session"` | `"one_time"`), so a genuinely one-time saving is never rescaled by session pace. Every current analyzer's waste recurs per session/call, so the field defaults to `"per_session"`, documented inline with the reasoning.
`estimated_recoverable_rollup` carries the same ratio, plus `active_days`/`n_sessions`/`window_days` and a human-readable disclosure line, alongside its existing (unchanged) raw observed sum.

## Tests / Verification

Renamed and extended `tests/unit/test_deadweight.py` / `tests/unit/test_cost_proposals.py` for the field rename and the new central-ratio behavior (guardrail-blocked vs guardrail-passing windows, downsize's projection now matching everyone else's basis).
Full suite green: `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` — 3566 passed, 2 skipped.
`ruff check` / `mypy` clean on every touched file.
Live-verified the fix with a real on-disk transcript scenario (a configured-but-never-called MCP server across 6 sessions spanning a 7-day window): the window figure is now exactly `6 x 25,000 = 150,000` tokens, not the old ~12.9x-inflated figure. Also verified the new projection end to end: a 30-day/10-active-day/25-session window yields ratio `3.0` with the disclosure line rendering real active-day/session/window values, and a thin window correctly falls back to the observed figure with the ratio reported as unavailable.

## What's NOT in this PR

`relearn`'s own `RelearnCluster.estimated_monthly_usd`/`estimated_monthly_tokens` (a structurally separate, unbounded-history system with no fixed window) is untouched — folding it into the same central ratio needs its own design pass for what "the window" even means for an unbounded corpus scan. Filed as a separate follow-up.

### DIFF
--- a/tests/unit/test_cost_proposals.py
+++ b/tests/unit/test_cost_proposals.py
@@ -231,7 +231,7 @@ def _dead_server(**overrides):
     fields = dict(
         name="apollo", scope="project", source="/repo/.mcp.json",
         sessions_present=10, invocations=0, deferred_sessions=0, dead=True,
-        estimated_tax_tokens_per_session=25_000, estimated_tax_tokens_90d=225_000,
+        estimated_tax_tokens_per_session=25_000, estimated_tax_tokens_window=225_000,
         tax_construction="25,000 tok/session (full schema injection), cited estimate.",
         fix="Remove or project-scope the `apollo` MCP server (/repo/.mcp.json); "
             "zero tool calls across 10 session(s) in this window.",
@@ -251,7 +251,7 @@ def _deadweight_finding(dead_servers=None, tax_table=None):
     return DeadweightFinding(
         sessions_scanned=10, configured_servers=1,
         servers=dead_servers, dead_servers=dead_servers, tax_table=tax_table,
-        estimated_recoverable_tokens=sum(s.estimated_tax_tokens_90d for s in dead_servers) or None,
+        estimated_recoverable_tokens=sum(s.estimated_tax_tokens_window for s in dead_servers) or None,
         estimate_basis="dead-server 90d tax sum",
     )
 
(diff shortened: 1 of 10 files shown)


anshss:

PR #544, opened 2026-07-25
Lead the context-resend claude-code fix with a durable subagent-offload rule

anshss:
## Summary
The context re-send advisory's claude-code branch led with `/compact` — a manual, per-session action a user typically abandons a full session for rather than running.
Replace the headline claude-code fix with an apply-capable, rung-1 CLAUDE.md rule (reusing the same `_persona_gated_write_fields` write machinery `script`/`reuse`/`verbosity` already use) that instructs offloading context-heavy sub-tasks to subagents, so the repeated context never accumulates on the main thread in the first place. `/compact` is kept only as a secondary, immediate-relief note for an already-full session.
The SDK `cache_control` lever and its persona gating are unchanged.

## Test plan
[x] `pytest tests/unit/test_cost_proposals.py tests/unit/test_context_resend.py tests/unit/test_optimize_resend_render.py` — all pass, plus new coverage for the apply-capable write on `claude-code`/`mixed` personas and the unchanged `sdk`/`unknown` behavior
[x] Full `pytest tests/unit` — 2928 passed, 2 skipped (pre-existing)

### DIFF
--- a/tests/unit/test_cost_proposals.py
+++ b/tests/unit/test_cost_proposals.py
@@ -1000,16 +1000,56 @@ def _resend_finding():
 
 def test_resend_suppresses_cache_control_snippet_for_claude_code():
     # The cache_control snippet is the SDK lever. A Claude Code window never
-    # constructs the request, so it is not theirs to paste — the card must lead
-    # with /compact and show NO snippet, matching how the cache family gates.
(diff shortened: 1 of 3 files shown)


anshss:

PR #545, opened 2026-07-25
Hide no-lever analyzers from the Recoverable waste band on Claude Code windows

anshss:
## Summary
The Overview "Recoverable waste" band built a tile for any finding carrying `estimated_recoverable_usd`, with no regard for whether the window's dominant persona actually has a lever for that analyzer.
On a Claude Code window this let non-actionable findings (e.g. cache_control tuning, Batch API placement, prompt-template trimming, a blanket verbosity cap, a brittle call-pattern script match) dominate the band and crowd out findings a Claude Code user can actually act on.
The API payload already carries the window's dominant `persona`; the tile-building path now reads it and drops the no-lever analyzers entirely before ranking (no placeholder row), so the tile that leads is guaranteed to carry a real lever. Only the `claude-code` persona is filtered — sdk/mixed/unknown windows are unaffected.

## Test plan
[x] Seeded a local `tj serve` instance with a claude-code-dominant window carrying real (non-zero) `cache`, `batch placement`, and `verbosity` findings alongside a `downsize` candidate, and confirmed via the live-rendered Dashboard that the Recoverable waste band shows only `Reuse` / `Downsize` / `Summarize` / `Context resend` / `Deadweight` — the no-lever findings are gone and the top tile (`Reuse`) is actionable.
[x] Re-seeded the identical data under a non-Claude-Code (`sdk`) agent id and confirmed the same findings (`Cache`, `Batch placement`, `Verbosity`, `Cache recommend`, `Script`, `Trim`) reappear in full — the gate is persona-conditional, not a blanket removal.
[x] Verified no console errors from the change.

### DIFF
--- a/tests/unit/test_lens_ui_regression.py
+++ b/tests/unit/test_lens_ui_regression.py
@@ -2852,3 +2852,32 @@ def test_review_inbox_dollar_headline_ignores_framing_even_when_suppressed():
     assert headline_unit(priced_deadweight, founder_framing) == "dollars"
     assert headline_unit(resend_structural, founder_framing) == "tokens"
     assert headline_unit(unpriced_placement, founder_framing) == "tokens"
+
+
+# --- Recoverable-waste band drops no-lever analyzers for claude-code -------- #
+def test_recoverable_band_excludes_no_lever_analyzers_for_claude_code(html):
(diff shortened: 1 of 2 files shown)


anshss:

PR #546, opened 2026-07-25
Fix: downsize and subagent double-count the same tokens in the recoverable rollup

anshss:
The Review-inbox "estimated recoverable" headline could exceed what a session ever spent. `downsize` and `subagent` were pricing the same model swap over the same tokens, under two different signatures, so the rollup summed both.

## Summary
`analyze_model_downgrade` now scopes its candidate aggregation to the main thread (`sub_agent_id IS NULL`), so downsize and subagent partition a session instead of overlapping it.
The window-wide denominators (`total_sessions`, `window_total_tokens`) deliberately stay unfiltered, via a separate totals query — only what we CLAIM narrows, never what we compare it to.
`thinking_tokens_by_session` gains a `main_thread_only` flag so the card's informational thinking share can't exceed 1.0 against a now-main-thread-only output figure.
New end-to-end regression test at the rollup layer, plus a durable rule in `CLAUDE.md` for the next analyzer that wants to claim a recoverable figure.

## Root cause

Claude Code stores a Task subagent's turns under its **parent** session id (backfill sets `sub_agent_id` from `agentId`/`isSidechain`). `analyze_model_downgrade` aggregated with a bare `GROUP BY session_id` and no `sub_agent_id` filter, so:

A session whose aggregate looked *small* only because it consisted mostly of one lightweight premium-model dispatch passed downsize's structural thresholds (`SMALL_INPUT_TOKENS` / `SMALL_OUTPUT_TOKENS` / `SMALL_TOOL_CALLS`) and produced a per-agent card claiming the subagent's tokens.
`subagent_rightsizing` independently aggregated the SAME spans scoped to `sub_agent_id IS NOT NULL` and flagged the dispatch `over_powered`, claiming those tokens again.
`estimated_recoverable_rollup` dedupes by **exact signature string**. `cost:downsize:<agent>` and `cost:subagent` are structurally distinct, so both survived.
`MODE(model)` over the mixed population could also name a **subagent's** model as the session's dominant one, so the routing suggestion on the card was not necessarily about the main thread at all.

Measured on a seeded session that spent **4,850 tokens / $0.32** in total, the rollup claimed **9,150 tokens / $0.3195** recoverable — 1.89x the tokens the session ever spent, and essentially all of its cost.

## Why the source filter rather than a downstream subtraction

Both remedies exist in-tree, and the cache family already uses the subtraction one (`_per_agent_cache_recoverable_by_model`). The column filter is the better fit here:

It repairs the **derived shape**, not just the sum — the per-session `MODE(model)` and tool-call count now genuinely describe the main thread. A subtraction would leave the card recommending a swap for the parent agent on the strength of a subagent's tokens.
`cache` cannot use it: its generic per-`(provider, model)` efficacy row must stay window-wide by definition, so subtraction is the only option there. Here the overlap is one column away.

The `CLAUDE.md` rule added in this PR states both remedies and when each applies, so this doesn't read as a second competing scheme.

## Tests / Verification

**`tests/unit/test_rollup_subagent_downsize_dedup.py`** (new) — one session, small main thread plus a single premium-model Task dispatch. Runs the real chain `build_report` -> `cost_proposals_from_report` -> `estimated_recoverable_rollup` and asserts the token sum equals the session's real total. **Confirmed RED before the fix**: `assert 9150 == 4850`. A per-analyzer unit test structurally cannot see this overlap, which is why the pin lives at the rollup layer.
Also pins that the denominators stay window-wide, so a later "simplification" that filters them alongside the numerator can't pass silently.
**Real local loop** on a seeded DuckDB store (`tj optimize --since 30d`, then `tj relearn cost-proposals`), isolated `HOME` + `TJ_CONFIG`, same store before and after:

| | headline | rollup tokens | rollup USD | cards |
|---|---|---|---|---|
| before | `~$0.32 estimated recoverable` | 9,150 | $0.3195 | `cost:downsize:claude-code-x` 4,850 tok + `cost:subagent` 4,300 tok |
| after | `~$0.30 estimated recoverable` | 4,850 | $0.2975 | `cost:downsize:claude-code-x` 550 tok + `cost:subagent` 4,300 tok |
| session truth | — | 4,850 | $0.32 spent | — |

After the fix the two cards partition the session exactly (550 + 4,300 = 4,850) instead of overlapping it.
Full CI suite green: **3,568 passed, 2 skipped**. `ruff check tokenjam/` and `mypy tokenjam/` clean.

## What's NOT in this PR

**`audit_opus_quota` / `tj context` still count subagent turns** in their whole-session quota figures. That is correct there and left alone: those are TOTALS of what was spent, not recoverable claims, and `context_diagnostic` documents the choice explicitly. Only figures that feed the rollup need disjoint populations.
**`over_provisioned`** still carries no recoverable estimate (its remedy is a prompt-size cut that can't be sized honestly), so it can't overlap anything.
**No threshold retuning.** Narrowing the aggregation means a large session with a small main thread can now become a downsize candidate on its main thread alone. That is the correct reading of the existing heuristic on a correctly-scoped population, and the tokens are still counted once; recalibrating `SMALL_*` for the new population is a separate question.


anshss:

PR #547, opened 2026-07-25
Apply persona gating consistently to model-swap, batch, and workspace-note proposals

anshss:
## Summary
The downsize (model-oversizing) card told every persona to "switch models," even an interactive Claude Code session that has no such lever — it now gets the levers it can actually pull (routing config export, subagent right-sizing, `/compact`) instead.
The batch-placement card showed an api-billed Batch API dollar figure regardless of persona; it's now suppressed outright for an interactive coding-agent session, since there's no application code of its own to move to a batch lane.
The relearn detector's rung-1/rung-2 CLAUDE.md/skill write ignored persona entirely, unlike the other analyzers that share the same write surface (script/reuse) — it now gates the same way.
The verbosity remedy text read as an enforceable "be concise" rule despite the underlying signal being scoped to one task-shape cohort; it's reworded as a review prompt, and the write path is disabled outright for every persona (not just persona-gated), since a cohort-scoped flag written into a global file would apply blanket terseness pressure to unrelated future tasks.

## Test plan
[x] `pytest tests/unit/test_cost_proposals.py tests/unit/test_relearn.py tests/unit/test_relearn_otel.py tests/unit/test_optimize_verbosity.py` — new + existing coverage green
[x] `pytest tests/unit` — full unit suite (2941 passed, 2 skipped)
[x] `pytest tests/integration/test_cost_proposals_api.py tests/integration/test_relearn_api.py`
[x] `ruff check` / `mypy` on all touched modules

### DIFF
--- a/tests/unit/test_cost_proposals.py
+++ b/tests/unit/test_cost_proposals.py
@@ -526,6 +526,10 @@ def _verbosity_finding(**overrides):
 
 
 def test_verbosity_proposal_shape_and_apply_fields():
+    """Unlike script/reuse, verbosity NEVER offers a write — for any persona,
+    including claude-code (see `_verbosity_to_proposals`'s docstring): a
+    cohort-scoped flag written as a global CLAUDE.md rule fails the
+    no-quality-tax gate outright, regardless of who would read it."""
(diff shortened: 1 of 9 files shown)


anshss:

PR #548, opened 2026-07-25
Report write-bearing savings net of the fix's own standing cost, and bound how many permanent rules are offered

anshss:
## Summary

Every analyzer whose fix **writes a permanent artifact** into the user's agent files reported its saving **gross** of what that artifact costs to keep. A `CLAUDE.md` rule is re-sent on every future session, forever, and nothing anywhere netted the two. A rule costing ~200 tok/session to prevent a 1,500-token failure twice a month is net negative, and the product showed it as a saving.

Compounding it, there was no write budget. `relearn` had no cap, no top-N and no cross-cluster merge, so 41 live clusters meant up to 41 separately appended blocks, and the tail carried a generic `"Review examples"` placeholder as a permanent rule. Meanwhile `summarize` exists to COMPRESS exactly the files those analyzers append to, and neither side knew the other existed: the same report could recommend adding dozens of rules and compressing the file they land in.

## What changed

Two new modules own the arithmetic, and both write lanes consume them.

**`core/optimize/projection.py`** is now the single 30-day projection basis: active-day pace `r = clamp(30 / D_active, 1.0, 3.0)`, with guardrails that suppress the projection entirely on thin data. A window-scoped saving and a per-session standing cost are only comparable on one horizon; this is where that horizon is defined, and the recoverable-savings rollup is its other intended consumer.

**`core/optimize/write_budget.py`** answers four questions for any set of proposed writes:

**What does it cost to keep?** Priced by **rung**, which is the distinction that makes the whole thing correct rather than merely conservative: a rung-1 `CLAUDE.md` block is charged in full, a rung-2 skill only at its always-loaded frontmatter (the body loads on invoke), and a rung-3 hook at **zero**, because a settings file is executed, never sent to the model as prompt text.
**Is it worth its own keep?** `net = gross - standing`. A non-positive net suppresses the write and the claim.
**How many rules may be offered?** Ranked by net value and capped, with the budget sized as a share of the always-loaded agent-file footprint `summarize` measures. That is the cross-reference the loop was missing.
**Is it a real rule?** A generic placeholder never becomes a permanent artifact.

Same-family clusters collapse onto **one** block: the family's largest carries the write, siblings say they are covered by it, and only one block's standing cost is charged.

Reported figures are now net. `relearn`'s four `estimated_*` fields and the equivalent `CostProposal` fields carry the net value; the pre-net figures move to `gross_*` and stay inspectable. A suppressed write degrades exactly the way the persona gate already degrades one: advise-only, with the fix still copyable as a snippet and its own reason on the card instead of the generic workspace-less one.

Two rules that are easy to get backwards and are enforced here:

The budget is spent in **per-session** tokens while the saving is netted against the **window** total. Charging the window total against the budget would shrink it as a user's session count grows, which is backwards: a busier user is exactly who a good rule pays off for.
A **project-scoped** rule is charged only against its own project's sessions, never the whole corpus. Netting a cluster-scoped saving against a corpus-scoped cost is the same basis mismatch in a different costume.

A finding with **no token estimate** is offered untouched rather than suppressed. Reaching a "net negative" verdict from a *missing* measurement would kill real fixes for the sin of being unquantified, which is strictly worse than the bug being fixed.

## Verification

Run through the real CLI path against a real 4,693-session local corpus (36.4-day span), not a fixture:

```
tj optimize relearn summarize --since 30d --json
```

| | Before | After |
|---|---|---|
| Clusters found | 45 | 45 |
| Permanent writes offered | up to 45, uncapped | **4** (all zero-standing-cost hooks) |
| Generic-placeholder rules offered | 20 | **0** |
| Claimed recoverable | 2,638,500 tok | **1,048,500 tok** |
| Any claim exceeding its gross | n/a | **none** |

Suppressions: 21 net-negative, 20 placeholder. Worked example from the corpus, `edit_before_read` (rung-1, user-global):

```
gross saving 457,500 tok (305 occurrences x 1,500)
block size 107 tok/session
exposure 4,693 sessions (user-global: every session re-sends it)
standing cost 502,686 tok
payback ratio 0.9 -> NET NEGATIVE, claim drops to 0
```

`cwd_confusion` is the contrast: a rung-3 **hook**, zero standing prompt cost, keeps its full 735,000-token claim and its write offer.

That live run caught two modelling errors before this landed. The first pass measured the agent-file footprint at 234,220 tokens by summing every catalog file whole, including 160 KB of `SKILL.md` bodies that only load on invoke; the always-loaded model measures 26,545. And an absolute size ceiling killed the lane for every real project, since a mature `CLAUDE.md` is routinely 25k tokens and appending one 107-token rule to it is 0.4% growth. The budget is now a share of the measured footprint, with the absolute stop moved to a genuinely pathological 50k.

## Tests

`tests/unit/test_write_budget.py` — 29 new tests covering the projection basis and its guardrails, standing cost by rung, the quality floor (including that an honesty caveat is never mistaken for a placeholder), net suppression, family merge, ranked budgeting, the summarize cross-reference, and both lane integrations.
Full CI scope (`tests/unit tests/synthetic tests/agents tests/integration`): **3,597 passed, 1 skipped**.
`ruff check tokenjam/` clean; `mypy tokenjam/` clean across 219 files.

## What's NOT in this PR

The two write lanes carry separate budgets rather than a shared pool: they are built in different passes that never see each other's proposals, so a joint allocation needs a seam that does not exist yet. The per-lane caps are explicit and documented rather than implicit.


anshss:

PR #571, opened 2026-07-25
Fix dollar attribution in relearn, summarize, and resend

anshss:
Three analyzers were reporting dollar figures that were missing, on the wrong time basis, or measuring something other than what they claimed. This fixes all three, plus the shared rate-blending they needed.

## Summary
- `relearn` now prices each occurrence with its own **measured re-read tail** instead of a flat per-turn cost, and stops claiming anything for clusters whose fix is a generic placeholder.
- `summarize` now carries a **USD figure on the same window basis as its siblings**, derived from observed session-load counts, instead of a dollarless one-time token figure.
- `resend` **splits cost-of-waste from recoverable** and re-derives the recoverable figure for the offload + right-sizing lever, measured from this corpus's own `sub_agent_id` telemetry.
- Rate blending shared by the first two moved into one module and now derives rates **per token class** from the pricing table.

## relearn: an occurrence costs more than one turn

A failure's text does not cost one turn's tokens once. It lands in the conversation and is re-sent on every later call that still carries it, so an occurrence at call index *k* bills `1500 x input_rate + 1500 x tail x cache_read_rate`.

Everything is expressed in **input-token equivalents** (`1 + cache_read_ratio x tail`) so the token and dollar figures stay proportional — the recoverable contract prices one against the other, and `write_budget` nets a rule's standing cost against the token side. Inflating tokens by the raw tail while leaving dollars on the cache-read rate would have quietly defeated that netting.

Compaction is accounted for rather than assumed away: the tail stops at the first prompt-size collapse (a `/compact`, a resume, a fresh window), because past that point the failure text is no longer in what gets re-sent. Each cluster then takes the **median** of its occurrences' multipliers, not the mean — a corpus's call volume is dominated by long uncompacted harness sessions whose mean does not describe a typical occurrence. Which basis was used is stated in `estimate_basis`.

- The `GROUNDED_TOKENS_PER_OCCURRENCE = 1500` constant is **unchanged**, per the inconclusive detour-hypothesis result.
- Clusters matching no known family fall back to generic "review examples" text. Those now **claim zero on every basis** (the same verdict the write budget already reached for the write, applied to the claim so it holds for every persona). The raw observation stays inspectable on the `gross_*` fields.
- `RelearnFinding` gained a `estimated_monthly_usd` total; `RelearnCluster` gained `tail_calls_median` / `tail_multiplier` so the multiplier is never a black box.

## summarize: always-on context saves repeatedly, not once

The catalog files (`CLAUDE.md`, `.claude/rules/*`, the `~/.claude` equivalents, `AGENTS.md`) are re-sent at the head of every session that loads them and re-read on every call within one. The saving recurs.

It is now priced over the analyzed window as `reduction x sessions-that-load x (first call at input rate + later calls at cache-read rate)`, with load counts **observed from telemetry** — a global-scope file against every session in the window, a project-scope file against its own repo's sessions only (matched through the same `agent_id` → repo derivation relearn uses). A file whose loading sessions cannot be identified carries tokens and **no** dollars: never a zero, never a rate borrowed from a file that did resolve.

The CLI renderer surfaces the figure through `render_savings`, so a subscription/local plan gets the same framing every other analyzer gives it.

### Inbox visibility: decided to keep it out

`summarize` stays out of `COST_ANALYZERS`, and the reasoning is now written down where the decision lives rather than being a one-line aside:

1. It has its own dedicated review surface (the curate/diff prepare/check/apply lifecycle) that a single-card adapter cannot represent.
2. **It is the budget, not a peer.** `write_budget.measured_agent_file_tokens` reads this finding to size how much permanent rule-writing every *other* analyzer may propose. Listing it in the inbox would make the analyzer that sets the budget also compete for it, and a user dismissing the card would be dismissing the counterweight that stops the rule-writers from growing the very files it wants compressed.
3. Its saving is already visible in the Overview waste band (registry-driven off the presence of `estimated_recoverable_usd`), so it is not hidden — only absent from that one surface.

## resend: cost-of-waste and recoverable are now different fields

**(a) The split.** `cost_of_waste_usd` is an observation: what re-sent context actually cost over the window, priced per token class at the rates it really billed at (cache reads at the cache-read rate, still-uncached repeat at the input rate). Nothing projected, nothing discounted, because nothing is being claimed. It is legitimately much larger than the recoverable figure and must never be presented as a saving — multi-turn work inherently re-sends context (the floor is not zero), offload *moves* context rather than deleting it, and compaction only helps forward and costs a summarization call of its own. `estimated_recoverable_rollup` reads only the `estimated_*` fields, so the gross can never leak into the Review inbox headline; a test pins that by stripping the recoverable fields and watching the rollup go to zero while the gross is still on the card.

**(b) The recoverable figure, re-derived and measured here.** The lever is compound: offload context-heavy in-thread work to a subagent (its tool output never enters the parent context, so it is never re-billed on later parent calls) **and** right-size that subagent. Neither cancels the other.

- `saving_offload` = the re-read tail that offloading removes, computed per main-thread turn with the same compaction truncation as relearn, priced at the cache-read rate.
- `saving_rightsize` = the same offloaded material priced at the cheaper same-family model's input rate instead of the premium one.

The avoidable fraction is **no longer the inherited 0.683**. Telemetry carries `sub_agent_id`, so it is measured directly: among sessions that delegate at all, what share of context-introducing volume already runs in subagents. No delegating session in the window means no dollar claim at all and a note saying so, rather than a fraction invented to fill the gap. `AVOIDABLE_FRACTION_OF_REPEAT` survives only on the tokens/compaction claim, whose mechanism genuinely is cross-corpus.

**Disjointness (Critical Rule 27).** The whole derivation runs over main-thread spans (`sub_agent_id IS NULL`) in sessions above a context-heavy threshold. That makes it disjoint from `subagent` by column filter and from `downsize` by construction (downsize's structural gate is session input under 5K tokens; this requires an order of magnitude more). Pinned by a new end-to-end test alongside the existing downsize/subagent dedup guards.

**(c) One compound card.** The `resend` card now carries both halves of the lever: the rung-1 CLAUDE.md offload rule as the write (naming the concrete over-powered subagent resolved from the `subagent` finding), and the `.claude/agents/<name>.md` frontmatter pinning both `model:` and `reasoning_effort:` as the one-paste artifact. This consolidates what `resend` and subagent right-sizing said separately — **no new card is added**.

## Shared rate blending

`relearn` and `summarize` both need "a block of tokens sent once at the input rate, then re-read at the cache-read rate", blended over the models actually observed. That now lives once in `core/optimize/rate_profile.py`, and it derives rates **per token class from the pricing table** rather than from an all-in observed cost-per-token — a blended `$/token` cannot tell an input token from a cache read, which is exactly the distinction these figures rest on. It returns `None` rather than any default when the data cannot supply a rate.

## Tests / Verification
- `tests/unit/test_relearn.py` — the removed `_blended_dollar_rate` tests were **relocated and inverted onto `blended_rate_profile`** rather than deleted (Critical Rule 23). New: the tail multiplier raises an occurrence above its raw 1,500 tokens while the constant stays 1,500; compaction truncates the tail; the finding-level USD total.
- `tests/unit/test_context_resend.py` — the old test asserting the cache_control-adoption derivation was inverted onto the new contract: cost-of-waste is observed, separate, and never equal to recoverable; cache reads are priced at the cache-read rate; the offloadable share is measured from `sub_agent_id` and is not the 0.683 constant; recoverable is offload + rightsize; no delegating session means no claim.
- `tests/unit/test_optimize_summarize.py` — a global-scope file is priced across every session and every re-read (strictly more than a single send); the CLI surfaces the dollar figure when priced. The existing "never fabricates a dollar" test was kept and now names its inverse.
- `tests/unit/test_cost_proposals.py` — compound card carries both halves; cost-of-waste rides its own fields and is absent from the rollup.
- `tests/unit/test_rollup_subagent_downsize_dedup.py` — new disjointness guard for resend vs subagent/downsize.
- Full suite: **3652 passed, 2 skipped**. `ruff check tokenjam/` and `mypy tokenjam/` both clean.

## What's NOT in this PR
- **A two-file atomic apply.** The compound card's rung-1 write lands in CLAUDE.md, and the apply engine (`relearn_apply._build_write_plan`) writes exactly one target per apply with one backup and one undo. Teaching it to write two files transactionally is a real change to the apply/backup/undo path, which is the "reinvent the write machinery" this was meant to avoid. The agent-file model/effort pin is therefore carried as the card's one-paste artifact with its resolved path, not as a second write. Flagging this as the one deviation from the brief.
- **Three-way card consolidation.** The `subagent` card still stands on its own. It has a genuinely stronger apply path (a deterministic `model:` key rewrite) and its own disjoint span population, so suppressing it would remove a working fix. The resend card consolidates the *recommendation* by naming that same agent file; it does not absorb the card. No card count increased either way.
- **Reasoning-effort writes through `model_apply`.** `render_agent_model` still only sets `model:`. The effort key ships as copyable frontmatter.
- **VERSION / CHANGELOG.** A separate open PR owns the version bump.

### DIFF
--- a/tests/unit/test_context_resend.py
+++ b/tests/unit/test_context_resend.py
@@ -209,9 +209,15 @@ def test_estimated_recoverable_tokens_uses_avoidable_fraction(db):
     assert finding.estimated_recoverable_tokens < finding.repeat_tokens
 
 
-def test_estimated_recoverable_usd_prices_uncached_share_at_rate_delta(db):
-    """Fully-uncached session: uncached_fraction == 1.0, so usd is exactly
-    repeat_tokens x (input - cache-read rate) x AVOIDABLE_FRACTION_OF_REPEAT."""
+def test_cost_of_waste_is_observed_and_never_the_recoverable_figure(db):
+    """The gross cost of re-sent context is an OBSERVATION on its own field.
+
+    It used to be absent entirely while `estimated_recoverable_usd` carried the
+    cache_control-adoption delta. Now the two are separate quantities: the gross
+    is priced per token class at what it really billed (cache reads at the
+    cache-read rate, the still-uncached repeat at the input rate) and is never
+    the same number as the recoverable one.
+    """
     _seed_session(db, "heavy", [1000, 1000, 1000, 1000], cache_ratio=0.0,
                   provider="anthropic", model="claude-haiku-4-5")
     _seed_session(db, "pad1", [500])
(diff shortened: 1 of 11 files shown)


anshss:
This PR is now bundled into #570 — the combined branch (including 5 review-bot correctness fixes applied on top of the merge) is pushed there, and this PR will auto-flip to Merged as soon as #570 lands via a merge commit. Please don't push new commits to this branch in the meantime — a new commit here would move the head off the SHA #570 already merged in and strand this PR open.

Issue #563, opened 2026-07-25
is_serve_process fails open when neither /proc nor ps is available


anshss:
**Severity:** Low
**Model:** haiku

Where: tokenjam/ (sub-repo) — tokenjam/core/server_state.py
Observed: is_serve_process() fails open (returns True) when it cannot read the command line because neither /proc nor ps is available, so a recycled stale PID could be signalled. The comment documents this as a deliberate tradeoff, and it only affects platforms lacking both /proc and ps, which is essentially none of the supported targets.
Fix direction: corroborate identity another way before signalling — e.g. record the process start time in the state file and compare it — or fail closed on platforms where the command line is unavailable.
Done when: a recycled PID cannot be signalled on a platform without /proc or ps.
Ref: review of the tj stop scoping PR (public repo PR 510).

be-student:
I’m working on this with Codex. I’ll make process identity checks fail closed when neither `/proc` nor `ps` can establish the command line, and add regression coverage proving that a stale live PID is not selected or sent a termination signal in that case.

PR #573, opened 2026-07-25
fix(transcript_cache): fingerprint equal-size in-place rewrites

### LINKED ISSUE #560: transcript parse cache can serve stale records on an equal-size in-place rewrite
**Severity:** Low
**Model:** haiku

Where: tokenjam/ (sub-repo) — tokenjam/core/optimize/transcript_cache.py
Observed: Cache validity is keyed only on (path, size, mtime). An in-place equal-length rewrite that preserves mtime, or two same-size rewrites inside one filesystem mtime tick, would serve stale cached records. Low real-world risk today because Claude Code transcripts are append-only JSONL that grow monotonically, so worst case is one-tick-delayed invalidation rather than permanent staleness.
Fix direction: add a cheap content fingerprint (hash of first/last N bytes) alongside size+mtime as a belt-and-suspenders check.
Done when: an equal-size in-place rewrite invalidates the cache, covered by a test.
Ref: raised by the review bot on the shared-helpers PR (public repo PR 511).

AshSgDe29071999:
## Summary
- Adds a cheap head/tail content fingerprint alongside `(size, mtime)` so equal-length in-place rewrites that preserve mtime invalidate the parse cache.
- Extends unit coverage for the equal-size / same-mtime rewrite case.

Fixes #560

## Test plan
- [ ] `pytest tests/unit/test_transcript_cache.py -q`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tests/unit/test_transcript_cache.py
+++ b/tests/unit/test_transcript_cache.py
@@ -154,9 +154,33 @@ def test_concurrent_writers_never_corrupt_the_cache_file(tmp_path):
     cache_dir = tmp_path / "cache"
     cache_path = cache_dir / tc._cache_key(src)
 
-    tc._store(cache_path, src, 100, 1.0, [{"a": 1}])
-    tc._store(cache_path, src, 100, 1.0, [{"a": 2}])
+    tc._store(cache_path, src, 100, 1.0, "fp", [{"a": 1}])
+    tc._store(cache_path, src, 100, 1.0, "fp", [{"a": 2}])
 
     loaded = tc._load(cache_path)
     assert loaded is not None
     assert loaded["records"] == [{"a": 2}]
+
+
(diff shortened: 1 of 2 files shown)


anshss:
Thanks for this, and sorry to close it. Going with #572, which takes the same
approach but hashes the full file rather than a head/tail window.

Two things decided it. The 64-byte window here means an equal-size, same-mtime
rewrite in the middle of a large transcript still gets served from cache, so the
original issue isn't fully closed; the added test uses a 60-byte file, which sits
entirely inside the window and so can't catch it. Separately, `_content_fingerprint`
opens the file without an `OSError` guard, so a transcript that's unreadable or
deleted mid-scan now raises instead of degrading to `[]` as the module documents.

The instinct was right and the diff was clean. If you'd like another, there are a
few open issues labeled `good first issue`, and #563 is close to this one in spirit.

PR #597, opened 2026-07-25
lots of fixes in resend, downsize, summrize relearn analyzers and leading with avoidable waste on ui

anshss:
A figure may wear the "waste" label only to the extent it was **avoidable**. Unavoidable or un-analysed spend is cost, and cost is a different claim.

Every change here is that principle applied, or a measurement gap it exposed.

The dashboard used to lead with **$1,152.46 "estimated recoverable / mo"**. That number was wrong in three separate ways at once: it was a forward projection built from a pacing multiplier, it omitted the two largest findings the product computes, and one of its cards was pricing work nobody could have avoided. This PR replaces it with what was actually observed.

---

## What a user sees now

| | before | after |
|---|---|---|
| headline | $1,152.46 "estimated recoverable / mo" | **$5,330.32 avoidable, observed over the window** |
| cost, shown separately | not rendered anywhere | **$7,019.10**, cost wording only, never summed with the headline |
| `summarize` | absent from every surface | **$3,968.57** — a real card |
| `downsize` | $3.91 | **$516.14** |
| `relearn` | $16.86 net, four reductions deep | **$46.30 observed**, all 55 clusters priced |
| `reuse` | $30.84 | gated off for this persona |

Measured on a real 2,291-session corpus, 30-day window, `claude-code` persona.

---

### 1. Lead with what was avoidable, past tense

Both surfaces now render one figure per analyzer: `past_overspend_usd`, observed over the analyzed window, never paced. The word "recoverable" no longer appears in user-facing copy.

Where an analyzer also knows the **total** cost of the behaviour, that lands on separate `observed_cost_*` fields with cost-only wording and its own rollup key. It is never added to the headline, because the avoidable figure is a *subset* of the cost — adding them double-counts, and swapping them relabels un-analysed cost as waste.

### 2. `resend`: stop implying 94% was unavoidable

`resend` reported **$6,972.80** cost of re-sending context beside **$398.41** avoidable. Nobody wrote the claim down, but the pairing made it: that 94.3% had been shown to be unavoidable.

It had not. The two figures were computed over **different populations**. Decomposed:

| bucket | cost | sessions | why it is not in the avoidable figure |
|---|---|---|---|
| in scope | $5,172.11 | 408 | analysed |
| driver-role | **$1,787.15 (25.6%)** | 706 | ceded whole to `downsize` — analysed on *another card* |
| no lever | $13.54 | 446 | below `MIN_SESSION_CONTEXT_TOKENS`, dropped before the calculation ran |

Three ways of not being analysed. Zero findings of necessity.

`context_resend` now partitions its cost by the **same predicate** the avoidable figure uses (`_coverage_class`), so the coverage a card states cannot disagree with the coverage the code applied; a test asserts the three buckets re-sum to the total. A `coverage_note` states the split in prose and ends with the load-bearing sentence: *the difference between the two figures is what this analyzer did not analyse, not what was unavoidable.*

### 3. `downsize`: ask about the model's ROLE, not its price

`downsize` contributed **$4.80 of a $1,191 rollup**. The reflex diagnosis was that its gate — `input < 5,000 AND output < 500 AND tool_calls <= 5`, summed per session — was unreachable for a coding-agent session.

Measured, that was wrong. **231 sessions (10.23%) clear the gate; they carry $6.74 (0.082%) of spend.** The gate fires fine and selects sessions worth nothing. Loosening it would have produced more near-worthless cards and moved the dollars by nothing, so the thresholds are **unchanged**.

The question worth asking was different: a premium model acting as the **driver** of a long session, doing reads and edits inline instead of routing them to cheap workers. Every inline turn re-reads the accumulated context at the premium rate, so the waste *grows with session size* — and a gate that sums per session becomes **less** likely to fire the worse the problem gets.

The new primary case flags a premium-tier driver, no subagent ever dispatched, context past 50K, ≥10 main-thread tool calls, and prices the inline-vs-routed counterfactual in two halves (offload tail, tier delta). **706 of 2,291 sessions. $512.23 of the $516.14 total.**

### 4. `summarize`: the second-largest finding, previously on no surface

`summarize` was excluded from `COST_ANALYZERS` because it owns its own curate/diff/apply flow. That is a real distinction — and a reason about *where the button goes*, not a reason to hide the finding.

**The Review inbox is the complete index of everything actionable, not the list of things whose apply flow happens to live there.** `summarize` now produces a normal card carrying its figure into the headline, with the action routing to `tj summarize` rather than offering a misleading inline apply.

Left as-is, this omitted the second-biggest number the product computes.

### 5. `relearn`: give it tokenjam's horizon, and stop gates zeroing observed cost

Two defects:

**It read the wrong source.** `relearn` is the analyzer premised on long-horizon recurrence, and it scanned Claude Code's on-disk transcripts — which Claude itself rotates at ~30 days. Its "unbounded history" was **37.4 days**, while tokenjam's own store held **120**. It now reads the DB, so its horizon is what tokenjam retained rather than what Claude has not yet deleted.

**Three gates turned real cost into $0.** Clusters with no fix template (29 of 55), net-negative write budget (17), and budget-deferred (4) all reported zero. But *"we have no action for this"* is not *"this was unavoidable."* All 55 clusters now carry an observed cost; the fix-availability question is answered separately from the did-this-cost-money question.

`relearn` is also adapted to a `CostProposal`, so it reaches the rollup at all — counted once, through its own card.

### 6. Ingestion was not continuous, and had been failing for a week

There was no watcher, no polling loop, no scheduled backfill, and no catch-up on daemon start. Two paths existed: live OTLP (only if the shell had the env vars *and* the daemon was reachable at that exact endpoint — no retry, no buffer), and `ingest_claude_code`, callable **only** by a human running a CLI command.

Ingested sessions had collapsed from ~200/day mid-month to **single digits**. Every figure the product showed was a floor.

Backfill was already idempotent — span ids are content-derived hashes and each run anti-joins — so it is now scheduled as a recurring daemon job with a catch-up on startup, and the remaining gap is surfaced to the user instead of being silent.

### 7. `reuse`: gated off for `claude-code`

`reuse` claims two pieces of planning were interchangeable. Unlike every other analyzer's claim, that is not observable. Measured:

- clustering ran on **tool-name sequence only** — `prompt_prefix_hash` was `None` for all 9 clusters
- **81% of its dollars** sat in one cluster whose only criterion was *"no tool call followed the plan"*: 1,167 members, 15 distinct projects, token counts spanning 0 → 341,199
- no time window (10 minutes and 21 days treated identically), no check for whether the prior attempt failed

$30.84 of a $5,330 rollup, and the one card a user could personally refute — which would cost them their trust in the $3,968 and $6,972 figures that are sound. The code stays; only the persona gating changes. Full record and the bar for re-enabling: `.claude/product-state/REUSE-ANALYZER-INVESTIGATION.md`.

### 8. One canonical dollar field

Three near-identical per-analyzer quantities coexisted: `past_overspend_usd`, `estimated_recoverable_usd`, and `estimated_monthly_usd` (the same number × a pacing ratio, measured 1.0714).

This ambiguity **produced a wrong analysis while this PR was being built** — a comparison of `past_overspend_usd` against `estimated_recoverable_usd` concluded the two were identical for 6 of 7 analyzers, while the Review inbox was rendering the third value.

`estimated_recoverable_*` and `estimated_monthly_*` are removed from `CostProposal` and every cost-analyzer finding. The contract is written into `CLAUDE.md` as an architectural rule with its rationale, and a test asserts no proposal carries a second per-analyzer dollar field. One documented exception remains on `RelearnCluster` for the Applied-tab snapshot, unreadable by any cost surface.

### 9. A measurement that came back negative, reported as such

Does codifying a lesson make repeated work cheaper? `repeat_task.py` measures it: **13 before/after pairs, ratios 0.16×–3.35×, confidence intervals excluding 1.0 in both directions**, and the known-positive project **null at all 4 codification events** once model choice was controlled for. Priced **$0.00**.

The effect is not disproven — it is *confounded*, and the telemetry cannot currently separate it. The machinery ships, prices nothing, and says so. Closing that gap needs a masked task-statement hash and the routed model persisted at ingest.

---

## Verification

- Numbers measured on a read-only copy of a real 359 MB telemetry store, never against the live DB or its lock.
- The recoverable rollup was byte-identical between branch and merge-base on an identical snapshot before the intentional changes, isolating code effects from data drift.
- Tests: **3,662 passing.** 95 failures are pre-existing and reproduce identically on the unmerged base (rich/ANSI console assertions; tracked separately).
- `coverage_note` is now **required** whenever `observed_cost_usd` is set, enforced by a test rather than a docstring — the same defect had appeared twice.

## Notes for review

- `resend`'s `offloadable_share` (0.4021) is still measured across the 175 of 2,291 sessions (7.6%) that delegate at all, then applied to the 92.4% that never do. A structural per-tool-call measure would be better; it is computed nowhere in this tree, so the honest move was to disclose the sample and spread in the basis string. Tracked separately.
- `downsize`'s card does not yet disclose that $1,787.15 of `resend`'s cost figure is the same sessions under a different heading. Rule 27 keeps the *avoidable* figures disjoint; the *cost* figures still overlap silently. Tracked separately.
- Figures here are floors: the ingestion gap in §6 means the corpus was under-counted while these were measured.


anshss:
### Added: a measurement of whether codifying a lesson makes repeated work cheaper

`bf16119e` adds a research instrument plus its findings. Short version: **the signal is not measurable on the current corpus, so `relearn` keeps pricing failure episodes and nothing about its unit of analysis changes here.**

The hypothesis under test was that `relearn`'s $46.26 across 55 clusters badly understates the waste, because the cost is the inflated *session* around a failure rather than the failing call itself. Testing that requires deciding two sessions were "the same work". Four things had to hold:

| gating question | answer | evidence |
|---|---|---|
| Are repeat-task clusters identifiable? | **yes** | 346 clusters over 2,711 sessions; 31 clusters at n≥5 cover 89% |
| Does the method reach the uninformed era? | **no** | transcript availability 99.9% (<30d) / 3.6% (30–60d) / 0.0% (older) |
| Can span-only shape substitute? | **no** | precision 0.54 @ recall 0.04 vs a 0.147 baseline, over 200k sampled pairs |
| Is the ~50% effect above the noise? | **no** | within-cluster cost CV 0.7–1.2; 95% CI on the median at n≈45 is ±46–49% |

Across the whole corpus only **13** (cluster × codification-event) pairs are testable at all, and their ratios span **0.16× to 3.35× with intervals excluding 1.0 in both directions** — a confound, not an effect. The confound is model routing: 5 of the 7 apparently-significant results had the model mix move underneath them. Holding it constant leaves the one project that motivated the idea at null or dearer at *every* codification event, priced at **$0.00**.

**What this does to the write-budget verdict:** neither validates nor overturns it. The measurement that would adjudicate it is not obtainable here. Keep the current suppression rule, and stop describing it as empirically settled in either direction — it is a model, and it remains one.

**Boundary settled as a side effect.** `workflow_restructure.py` (the `script` analyzer) clusters on ordered tool-call signature, which is the same primitive measured at 0.54 precision — so that table is a direct audit of it. Not a bug: a script replaces a *shape*, and shape is the right unit for "could code do this". It is the wrong unit for "was this the same work". The two stay separate, and `relearn` must not adopt `script`'s signature as a same-work test.

**What lands**
- `tokenjam/core/optimize/repeat_task.py` — the instrument. Similarity methods carry their *measured* precision as data; `measure_codification_delta` refuses to price below the bar rather than guessing, returns `null` whenever the bootstrap interval spans 1.0, and prices off the conservative end when it does fire. Not registered as an analyzer; emits no figure into any report.
- `docs/internal/repeat-task-codification-measurement.md` — every table, the confound, and the four instrumentation changes that would make the question answerable (persist a masked task-statement hash at ingest is the cheap, highest-value one).
- `tests/unit/test_repeat_task.py` — 26 tests. Full unit suite: 3165 passed, 1 skipped.

Deliberately untouched: `relearn.py`, `cost_proposals.py`, `write_budget.py`, and every dollar field name.

anshss:
## Collapsed the three per-analyzer dollar fields to one canonical `past_overspend_usd`

The product carried three near-identical per-analyzer dollar quantities. Measured on a real 2,291-session corpus with the pre-change build, `past_overspend_usd` and `estimated_recoverable_usd` were **byte-identical for 23 of 23 priced proposals**, and the aggregate's paced total was a third number. Three overlapping names for one quantity is a standing trap for anyone changing how an analyzer computes.

`estimated_recoverable_usd`/`_tokens` and `estimated_monthly_usd`/`_tokens` are gone. One field per analyzer: `past_overspend_usd`/`_tokens`, observed over the analyzed window. `observed_cost_usd` stays — it is a genuinely different quantity (total cost, of which the avoidable figure is a subset) and only `resend` and `relearn` carry it.

### Consumer audit — every reader of the removed fields, before removal

**Analyzer findings (renamed → `past_overspend_*`)**
`batch_placement`, `cache_efficacy` (4 finding types), `cache_recommend` (2), `context_resend`, `deadweight`, `model_downgrade`, `output_verbosity`, `plan_reuse`, `prompt_bloat`, `subagent_rightsizing`, `summarize`, `workflow_restructure`, plus `types.DowngradeFinding` and `types.ReuseFinding`.

**Core pipeline**
`cost_proposals.CostProposal` (field), all 15 adapters, `_with_past_overspend`, `_apply_write_budget`, `runner.py` (13 dict→finding rehydrations on the daemon path), `cost_apply.py`, `relearn_store.py`, `relearn_proposals.py`.

**Deleted outright**
`compute_projection_ratio`, `_effective_ratio`, `_with_rollup_projection`, `backfill_legacy_monthly_fields`, `estimated_recoverable_rollup`, the `CostProposal.scaling` field, the `MIN_*_FOR_PROJECTION`/`MAX_PROJECTION_RATIO`/`OBSERVED_LABEL`/`PROJECTED_LABEL` constants, and the cached `cost_active_days`/`cost_n_sessions` pace inputs (a cached pace input is an invitation to re-derive a ratio).

**API**
`routes/cost.py` (recoverable list + largest-recoverable payload), `routes/relearn.py`. `GET /relearn/cost-proposals` **drops its `rollup` key** — the `past_overspend` block is now the only aggregate and carries the `excluded` passthrough. relearn's clusters are no longer folded in on their own monthly basis; relearn reaches the aggregate through its own `cost:relearn` card like any other analyzer, counted once.

**UI** (`ui/index.html`)
Analyzer-findings surfaces (Optimize detail cards, the waste band, the lens regression path), the applied-fix estimate mapping, the excluded-waste note, and the `rollup.excluded` fetch → `past_overspend.excluded`.

**CLI**
`cmd_optimize.py`, `cmd_status.py`, `cost_proposal_verbs.py` (now reads `past_overspend_rollup`; headline reworded from "estimated recoverable" to "already overspent, observed").

**Other**
`core/export/reuse_report.py`, `core/recommendations.py` (the downsize read only — the outcomes-ledger key of the same name is a different domain and is untouched).

**Deliberately NOT touched — different domains that share a field name**
`core/db.py`, `core/models.py` (`PolicyLedgerEntry`), `otel/semconv.py`, `proxy/audit.py`, `cli/cmd_policy.py`, `docs/policy/overview.md` — these are the **policy would-action ledger**, unrelated to cost analyzers. A blind rename would have broken them.

### Fate of `compute_projection_ratio()` / `projection.py` — decided and recorded

`compute_projection_ratio()` is **deleted**; it had no consumer left once the monthly fields went.

`projection.py` **stays**, because `write_budget` genuinely needs a per-month horizon: pricing a permanent CLAUDE.md rule requires "how many sessions will re-send this in a month". It is now reached only through `cost_proposals._write_budget_basis`, it never touches a past-tense figure, and nothing it produces lands on a payload — only the netted `past_overspend_*` and the `standing_cost_*` fields do. Both modules carry a comment saying so.

`model_downgrade.monthly_savings_usd` and `AgentPriceRow.projected_30d_delta_usd` survive but reach only the CLI's own `tj optimize` line and an explicitly-labelled evidence sentence — never a card, payload, or aggregate.

### Two documented exceptions, both on relearn

`RelearnCluster` (the per-cluster inbox rows, **not** a `CostProposal`) keeps its own `estimated_recoverable_*` — a fix-gated forward claim over an unbounded corpus, a genuinely different quantity, excluded from every aggregate by design — and its `estimated_monthly_*`, now the only 30-day-basis figure in the product. Renaming the first onto `past_overspend_*` would have been actively wrong: a cluster with no fix template has a zero forward claim but real past overspend. Both are named in the CLAUDE.md contract with the rule that no cost surface may read them.

### Backward compatibility

A proposal cache written before the collapse migrates its `estimated_recoverable_*` keys onto the canonical field and **drops** the paced twin rather than promoting it — reviving a paced figure as the past-tense headline is the exact defect this removes. An entry carrying only the paced key renders no dollar figure until the next recompute (≤6h), which is the honest degradation.

### The durable half

The field contract is written into the repo `CLAUDE.md`: the canonical names, what each means, what time basis each is on, the four rules (one field per quantity; never summed with the cost total; no user-facing surface may render a paced figure; one rollup), the `None` ≠ `0.0` rule, and the two relearn exceptions. Critical Rule 30's paragraph is updated to say this was fixed structurally rather than by convention.

### Validation

Local daemon against an **isolated copy** of the real 343 MB DuckDB (the real user DB was never written to), driving the real API and the real UI headlessly.

| # | Check | Result |
|---|---|---|
| 1 | The three fields really were near-identical | **PASS** — `past_overspend_usd == estimated_recoverable_usd` for 23/23 priced proposals; `rollup.estimated_recoverable_usd` == `past_overspend.past_overspend_usd` == `5981.720134` exactly |
| 2 | The paced aggregate was a third number | **PASS (defect reproduced)** — `projected_usd_30d` = `6002.26871` vs observed `5981.720134` |
| 3 | Collapse changes no number | **PASS** — same-vintage runs: `deadweight` and `downsize` (DB-only analyzers) identical to the cent, 0.000%; `resend` +0.015%, `reuse` −0.001%, `subagent` −0.001%, `summarize` +0.119%, total +0.092% — sliding-window and filesystem drift (`summarize` scans the live filesystem, so no two runs are bit-identical regardless of this change) |
| 4 | Retired fields gone from every payload | **PASS** — zero retired keys across 24 proposals; zero retired or paced keys on the aggregate |
| 5 | Exactly one aggregate | **PASS** — `rollup` key gone, `past_overspend` only |
| 6 | relearn's null not silently zeroed | **PASS** — `past_overspend_usd = None` (not `0.0`), `observed_cost_usd = 46.30` |
| 7 | `summarize` follows the single-field contract | **PASS** — one card, one dollar field |
| 8 | No surface renders a paced figure | **PASS** — every cost card reads "$X avoidable over the last 30 days"; no "est. / mo" anywhere in the Cost-advisories tab. The Recurring-mistakes (relearn cluster) tab still shows it — the documented exception |
| 9 | Both leading surfaces read one server figure | **PASS** — Dashboard hero and Review inbox headline render the same number, no client-side derivation |
| 10 | No test regression | **PASS** — identical failure set before and after (98 pre-existing, environment-related: unicode / onboarding / CLI rendering); 3,707 passing |
| 11 | Lint + compile | **PASS** |

Tests were updated, not dropped: the assertions that pinned the removed distinction now pin the collapse itself — no retired field on any proposal, payload or rollup; exactly one aggregate; legacy-cache migration; relearn's null never coerced to zero.

One flag for review: `docs/optimize/{reuse,subagent,verbosity}.md` got a mechanical field-name rename (no prose or positioning change) so the analyzer docs don't name fields that no longer exist. `docs/policy/overview.md` is the separate policy domain and is untouched.

PR #599, opened 2026-07-26
chore: remove verified dead code

anshss:
Removes eight symbols that nothing calls, found by a full reachability sweep of the repo rather than a single tool's output.

## How the sweep worked

`vulture` on its own reported 277 candidates, but almost all were false positives: it does not understand that a FastAPI route decorator, a `click` command decorator or an MCP tool decorator *is* the call site. So the scan was redone with an AST pass that extracts every top-level `def`/`class`/`CONSTANT` and every class method in `tokenjam/`, then cross-references each name against every `.py`, `.ts`, `.js`, `.html`, `.md`, `.toml`, `.json`, `.yml`, `.sh` and `.txt` file in the repo — package, tests, examples, incidents, `sdk-ts`, `npm-wrapper` and docs — counting references inside the defining file separately from references outside it.

That left 9 unreferenced non-registered symbols and 12 referenced only from tests. Each was then read and verified by hand; the ones below are the subset where removal is provably safe. Complementary passes: a statement-level unreachability check (code after `return`/`raise`, constant-false branches) found **0** hits, `ruff` with `F401`/`F811`/`F841` found **0** unused imports or locals, and `ts-prune` on `sdk-ts` flagged only `index.ts` re-exports, which are the published package surface.

## Removed — zero references anywhere, including tests

| Symbol | Why it is dead |
|---|---|
| `_restart_tj_server_for_secret_rotation` (`cli/cmd_onboard.py`) | A self-described "backward-compatible alias" for `_restart_tj_server(..., reason="secret")`. Nothing has ever called it. |
| `_exposed_for_tests` (`cli/cmd_policy.py`) | A test-visibility shim that no test imports. The three names it re-exported are reachable directly. |
| `ValidationError` (`core/optimize/validate.py`) | Never raised, never caught, never imported. |
| `KNOWN_PROVIDERS` (`proxy/gate.py`) | Never read. The gate resolves the provider through `provider_pricing_mode`, which does its own matching. |

## Removed — dead in the product, tests were the only consumer

**`_codex_mcp_toml_block`** (`cli/cmd_onboard.py`) built the legacy `[mcp_servers.tj]` block. tj no longer writes one — Codex is wired out-of-band via OTel — and the retirement path `_codex_strip_tj_mcp_from_content` matches by regex rather than against this string, so the docstring's stated reason for keeping it ("retained only to *recognize* a previously tj-written block") did not actually hold. Its four tests were really exercising the generic section-writer `_codex_apply_block`, so they now build the same block from a literal fixture and keep their coverage intact.

**`session_shares`** (`cli/cmd_statusline.py`) returned two of the three values from `_session_figures`. Its documented caller, the quickstart preview, no longer exists. The tests now call `_session_figures` directly, which covers the same code path plus the window-occupancy value the wrapper was discarding.

## Removed — assigned but never read (dashboard)

`costBins`, `monthlyUsd` and `hasStack` in `ui/index.html`.

## Deliberately kept

Three categories came up as unreferenced and were left alone on purpose:

- **Unused `semconv` constants** (`KIND_LLM`, `KIND_TOOL`, `APP_VERSION`, `SLUG`, `SERVICE_NAME`). This module mirrors the OTel GenAI / OpenInference specs; it is a vocabulary table where completeness is the point, not a call graph.
- **80 unused `noqa` directives** (`RUF100`). They are inert only under the current rule selection, and several carry rationale in their trailing comment. Stripping them would be churn that deletes intent.
- **`core/optimize/repeat_task.py`**, imported by nothing but its own tests. It landed as explicitly not-yet-wired measurement infrastructure, not as a leftover, so it is reported here rather than deleted.

A handful of other test-only helpers (`_package_uninstall_hint`, `session_token_cost_rollup`, `four_type_token_total`, `infer_schema_from_outputs`) look more like wiring gaps than dead code — each is a function that plausibly *should* have a caller. They are left in place; deciding whether to wire or drop them is a separate call.

## Verification

- `ruff check tokenjam/` — clean
- `mypy tokenjam/` — clean, 222 source files
- Full suite (`tests/unit tests/synthetic tests/agents tests/integration`) — **3813 passed**, 2 skipped

One test fails, `test_transcript_sync.py::test_catch_up_ingests_the_missing_session_and_is_idempotent`. It reproduces identically on pristine `main` with these changes stashed, so it is pre-existing and unrelated to this PR.

Net: **-67 / +24** lines across 8 files.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -115,7 +115,7 @@ Key flags: `--session`, `--last` (default when `--session` is omitted), `--json`
 
 ### `tj quota-audit`
 
-Retroactive audit of your premium (Opus/Fable) quota: which past premium-tier sessions were structurally Sonnet-shaped (small input/output, few tool calls)? Covers Fable (the tier above Opus) as well as Opus. Reports the percent of premium quota that went to Sonnet-shaped sessions (a retrospective behaviour mirror; those tokens are already spent, so it is misallocated, not "reclaimable"), example sessions to spot-check, and an optional tuned routing-config export. Subscription users see a habit nudge; API-billed users additionally see the already-billed dollar counterfactual. Quota-share framing, never a dollar "saving" claim. The JSON field is `percent_quota_misallocated` (the pre-0.6 `percent_quota_reclaimable` key is still emitted as a deprecated alias for one release). Needs a direct DB connection (can't run against a live `tj serve`).
+Retroactive audit of your premium (Opus/Fable) quota: which past premium-tier sessions were structurally Sonnet-shaped (small input/output, few tool calls)? Covers Fable (the tier above Opus) as well as Opus. Reports the percent of premium quota that went to Sonnet-shaped sessions (a retrospective behaviour mirror; those tokens are already spent, so it is misallocated, not "reclaimable"), example sessions to spot-check, and an optional tuned routing-config export. Subscription users see a habit nudge; API-billed users additionally see the already-billed dollar counterfactual. Quota-share framing, never a dollar "saving" claim. The JSON field is `percent_quota_misallocated`. Needs a direct DB connection (can't run against a live `tj serve`).
 
 ```bash
 tj quota-audit
--- a/tests/integration/test_data_access_parity.py
+++ b/tests/integration/test_data_access_parity.py
@@ -162,8 +162,6 @@ def test_quota_audit_parity(tmp_path, monkeypatch):
     assert d_audit.opus_sessions == 3
     assert d_framing.pricing_mode == "subscription"
     # The renamed headline key is present on both paths, and the deprecated
-    # 0.5.4 alias mirrors it byte-for-byte (so parity holds while it lives).
     assert "percent_quota_misallocated" in d_payload
-    assert d_payload["percent_quota_reclaimable"] == \
-        d_payload["percent_quota_misallocated"]
+    assert "percent_quota_reclaimable" not in d_payload
     db.close()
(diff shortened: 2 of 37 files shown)


anshss:
CI note: the single failing test, `test_transcript_sync.py::test_catch_up_ingests_the_missing_session_and_is_idempotent`, is not from this PR — `main` is currently red from it on its own. The tip-of-main run (30199363497, the 0.6.1 bump merge) fails identically, the run before it passed, and the test also fails locally on a pristine `main` checkout with these changes stashed. It is being tracked separately. Every other check on this PR passes, and lint/typecheck are clean.

anshss:
Widened the scan after realising the first pass was narrower than "complete": it only walked the `tokenjam/` package, only at module-top-level and class-method depth, and treated anything with a registering decorator as automatically live.

The second commit covers what was missing — every `.py` file in the repo (examples, incidents, scripts, tests) at every nesting depth including nested functions, class attributes and dataclass fields; `sdk-ts`/`npm-wrapper` internals rather than just exported surface; the dashboard's CSS as well as its JS; and actual caller-reachability for all 69 routes and the CLI groups.

Yield: two more dead Python symbols (one of them only discoverable transitively, after removing the first) and two dead CSS rules. `sdk-ts` and `npm-wrapper` are clean, and all 69 routes have a real caller.

The more useful half was the false positives it rejected — 6 CSS custom properties, 5 `badge-*` classes, 3 `ap-*` classes, 2 depth classes and 3 API routes all looked dead to a text scan but are reached through dynamically composed names (`cssVar('--chart-' + i)`, `badge-\${status}`, `ap-\${kind}`, `'d' + Math.min(a.depth, 3)`, and path building by string concatenation). Details in the commit message.

PR #601, opened 2026-07-26
Add tj upgrade: upgrade the package and restart the daemon

anshss:
`tokenjam/__init__.py` resolves `__version__` once at import time, and `tj serve` is a long-lived daemon (often supervised by launchd). Upgrading the package outside `tj` (`pipx upgrade tokenjam`, etc.) previously left the running daemon serving the OLD version out of process memory indefinitely — verified live: the package on disk was 0.6.1 while the daemon's own `/api/v1/version` endpoint still reported 0.6.0. There was no `tj upgrade` subcommand at all; the only guidance was a manual `tj stop && tj serve &` in the CLI epilog and docs/installation.md.

## Summary
- New `tj upgrade` subcommand: detects how the running `tj` is installed (pipx / `uv tool` / pip+venv), runs the matching upgrade command, and only on a successful install restarts the daemon and verifies it actually picked up the new version.
- Ephemeral/unmanaged installs (uvx/npx/pipx-run cache, or an unwritable pip target) are refused up front with actionable guidance — never half-upgraded.
- Daemon restart prefers `launchctl kickstart -k` when launchd supervises the daemon (verified live to work); otherwise falls back to the existing `stop_tj_serve()` + relaunching `tj serve` in the background, reusing `cmd_stop.py`/`cmd_onboard.py` rather than duplicating their logic.
- After restart, polls `/api/v1/version` on the daemon's own port with a bounded timeout and reports old → new version; fails loudly (non-zero exit) if the daemon still reports the old version once the timeout elapses — it never claims success without observing the new version.
- Updated the CLI epilog and `docs/installation.md` to point at `tj upgrade` (manual pipx/pip + `tj stop && tj serve &` steps kept as a documented fallback).

## Tests / Verification
- `tests/unit/test_cmd_upgrade.py` (25 tests): install-manager detection per layout (pipx / uv-tool / pip / ephemeral-refusal / unwritable-refusal), daemon NOT restarted when the install fails, launchd-kickstart vs stop/serve fallback branch, no-daemon-running no-op, and post-restart version verification (success + failing loudly on timeout). All subprocess/launchctl/HTTP calls are mocked — nothing here upgrades a real package, shells out to launchctl for real, or touches a real daemon.
- Full suite: `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` — 3841 passed, 1 skipped.
- `ruff check tokenjam/` and `mypy tokenjam/` (223 files) both clean.

## What's NOT in this PR
- systemd (Linux) daemon restart uses the existing stop/serve fallback rather than a `systemctl restart`-equivalent fast path — only the launchd fast path was explicitly verified live on this machine; the systemd case still gets a correct (if slightly slower) restart via the shared fallback.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -108,17 +108,24 @@ See [`docs/optimize/trim.md`](optimize/trim.md) for performance numbers, capture
 ## Upgrading
 
 ```bash
-pipx upgrade tokenjam          # if you installed via pipx (recommended)
-pip install --upgrade tokenjam # if you're in a pip + venv setup
+tj upgrade
 ```
 
-After upgrading:
+`tj upgrade` detects how `tj` is installed (pipx / `uv tool` / pip+venv), runs the matching upgrade command, and — only once that succeeds — restarts the `tj serve` daemon (`launchctl kickstart` if launchd supervises it, otherwise `tj stop` + relaunch) so it actually reflects the new version, instead of quietly continuing to serve the old one from process memory. It then polls `/api/v1/version` to confirm the restart landed before printing success.
 
-1. Restart the daemon to pick up the new code: `tj stop && tj serve &`
-2. DB migrations apply automatically on the next `tj` invocation; no manual step required
-3. Verify with `tj --version`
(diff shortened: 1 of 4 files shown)


anshss:

PR #602, opened 2026-07-26
feat(quickstart): close the first run on past overspend, not a session table

anshss:
The zero-install first run (`npx tokenjam`) ended on a Session Story teaser and a per-session timeline table. Neither answers the question a first-time reader actually has once they've seen their quota composition: **what did this already cost me, and what do I change?** The timeline table is also the least useful thing on screen for a user with hundreds of sessions, since a closed session can't be reclaimed.

## Summary

- Removes the Session Story teaser and the "Session timeline (most recent)" table from `cmd_quickstart._render()`.
- In that same slot, adds **one line naming both telemetry ingest paths** so the report doesn't read as a Claude-Code-only tool: Claude Code arrives from the local `~/.claude/projects/*.jsonl` session logs; Anthropic SDK / API traffic arrives from OTel spans or the tokenjam MCP server. (The MCP server is the SDK/API path, deliberately not described as the Claude Code one.)
- Adds a **past-overspend callout**: the single largest `past_overspend_usd` cost proposal over the ingested window, its evidence line, and the `advise_text` fix it ends in.
- Removes the now-uncalled `_render_session_story_teaser` and `_bar` helpers. `tj session-story` itself is untouched and still a real command.

## The callout

It goes through the same `build_report` / `cost_proposals_from_report` path `cmd_status._recoverable_teaser` and the Review inbox already use, so **persona gating stays inside `build_report`** (`PERSONA_DISABLED_ANALYZERS`, a true skip before dispatch) and is not re-implemented here. Quickstart still reads and writes **no config file** and opens **no on-disk DB**: the analyzers get an in-memory `TjConfig(version="1")`, and there's a test that fails loudly if config discovery is ever reached.

`cmd_status`'s stricter api-only pricing gate is deliberately not reused. Quickstart reads no config, so the plan tier is always `unknown` and that gate would suppress every dollar figure; assuming API pricing is the recorded stance, and this same render already prints an "implied API value" total for the same window.

Honesty rules it holds to, all from the per-analyzer dollar-field contract in `CLAUDE.md`:

- the figure is `past_overspend_usd`, observed over the window, never paced, projected, or multiplied by a 30-day ratio, and never summed with `observed_cost_usd`;
- the **single largest** proposal, never a sum: the analyzers price overlapping angles on the same spans, so summing double-counts;
- `None` means "not measured", never `$0`. Below a $1 floor (mirroring `_TEASER_MIN_USD`) it falls back to the token figure and states why; with nothing to report it renders no callout at all.

One judgment call the brief didn't cover: it also **drops any figure larger than the window's own total cost**. This render prints that total ("implied API value $244") a few lines below the callout, so a larger avoidable figure is self-refuting on the same screen. This is real, not hypothetical: on a 300-session corpus `deadweight` returned $357 against a $244 window, because its per-session MCP schema tax is summed over every transcript on disk while the transient DB holds the capped, most-recent subset. The over-ceiling finding is dropped, never rescaled, since rescaling would invent a number nothing measured.

Also strips a trailing money clause from the proposal title (`"Review 17 oversized files, $187.55"`), which would otherwise print two differently-rounded amounts in one sentence now that the callout leads with the figure.

## Runtime

`relearn` is dropped from the analyzer set **for this path only**, and it's the only one dropped. It reports on `cost_of_waste_*` and leaves `past_overspend_*` at `None` (it scans unbounded history and has no fixed window to observe an avoidable figure over), so it can never win this selection — while costing 27.9s of the 35.0s the whole set took on a real 300-session corpus. Per-analyzer timings on that corpus: `relearn` 27.86s, `deadweight` 4.86s, `resend` 0.95s, `summarize` 0.42s, `downsize` 0.28s, everything else ≤0.01s.

End to end on that corpus the command is **~14s** (7s ingest, ~6s analyzers), against ~42s with `relearn` in and ~7s before this change. That keeps the documented sub-30s first-run budget. If `relearn` ever moves onto the canonical field, the helper is a one-line delete.

## Copy

Removes the em dashes from this render's copy per the standing rule, including the caveat this brief called out (`CONTEXT_HONESTY_CAVEAT`: `"...not free) — real quota"` becomes `"...not free): real quota"`). Note the callout renders a winning proposal's `advise_text` and `caveat` verbatim, and several of those strings (authored in `core/optimize/cost_proposals.py`, shared with the CLI, the Review inbox and the web UI) still contain em dashes. They're not rewritten at render time on purpose: the honesty caveats are contractually surfaced verbatim, so that fix belongs at the source in a change that moves every surface at once.

## Tests / Verification

`tests/unit/test_quickstart.py` — the three tests that asserted on the removed blocks are rewritten rather than dropped, plus new coverage:

- selection: largest priced finding wins (not the sum); over-window-cost figure dropped; sub-$1 falls back to tokens with the reason; never-priced falls back to tokens with a different reason; nothing qualifying returns `None`; analyzers raising returns `None` instead of killing the run; money-suffix stripping; unusable advice falls back to a `tj optimize` pointer;
- render: the both-personas line, the dollar form, the token form (asserting no `$0.00` alongside it), and a **real, unstubbed** degrade path — a fixture on the cheapest priced model so no analyzer has anything to propose, asserting no panel, no fabricated figure, and that the rest of the report is unaffected;
- the Session Story teaser is gone while the statusline preview it hung off still renders;
- quickstart's own copy carries no em dash;
- quickstart reads no config and opens no on-disk DB.

```
3829 passed, 1 skipped, 2 warnings in 182.06s (0:03:02)
```
(`env -u FORCE_COLOR python3 -m pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/ -q`)

`ruff check tokenjam/` and `mypy tokenjam/` clean (222 source files).

## What's NOT in this PR

- The em dashes in analyzer-authored `advise_text` / `caveat` strings (see Copy above) — a cross-surface change.
- The `deadweight` population mismatch that produced the $357-vs-$244 figure (its session count spans more transcripts than quickstart ingests). Guarded against here at the display layer only; the analyzer itself is untouched.
- The `--json` payload is unchanged: the callout is computed only on the human path, so `--json` stays byte-clean and fast.


anshss:

PR #603, opened 2026-07-26
Slim the tj onboard close, and cut its palette from eight colours to one accent

anshss:
`tj onboard` ended on a wall of terminal output: a four-command next-steps list, then a full recurring-mistakes report (progress line, cluster list, recoverable-token estimate, a "#1 fix" evidence dump with session IDs and tracebacks), then a blocking `Enable this fix now? [y/N]` ask. That is a lot of screen, and one more decision at the exact moment a new user has the least context to make it.

## Summary
- Claude Code next-steps list is now exactly two entries: the web dashboard (live URL) and `tj tokenmaxx`.
- The recurring-mistakes section is removed entirely, replaced by a two-line pointer at the dashboard's review inbox.
- The relearn scan itself is removed, not just its printing (see below).
- `tj onboard` has one fewer blocking question.

## Next steps: four rows to two

Before (Claude Code persona):

```
  tj context      where your quota goes - re-read vs real work
  tj quota-audit  how much Opus/Fable went to Sonnet-shaped sessions
  Lens (web UI)   already running -> http://127.0.0.1:7391/
  tj tokenmaxx    your shareable efficiency tier
```

After:

```
  web dashboard  already running → http://127.0.0.1:7391/
  tj tokenmaxx   your shareable efficiency tier
```

`tj context` and `tj quota-audit` still exist as commands; they are just not where a fresh user should be sent first, since the dashboard renders the same diagnosis without re-deriving it per invocation. The web UI row is labelled "web dashboard" on this list. This is a rename in the onboard next-steps list only, deliberately not a codebase-wide rename.

## Recurring mistakes: removed, computation included

The whole block is gone: the `Scanning your history for recurring mistakes…` progress line, the cluster list, the `~N estimated recoverable tokens` line, the `Your #1 fix:` evidence dump, and the interactive enable prompt. In its place:

```
Recurring mistakes: tj keeps watching your sessions in the background.
  Review and apply fixes in the web dashboard: http://127.0.0.1:7391/#/review
```

**The computation was removed too, and that is safe.** Traced before deleting: onboard called `compute_relearn_finding` directly with `conn=None` and `transcript_cache_dir=None`, bypassing `core.optimize.relearn_store` entirely. It persisted nothing and warmed no cache, so no other surface read its result: it was compute-for-display only, and expensive (tens of seconds over a real corpus). The daemon's own background job computes and caches the findings the review inbox renders, on its own schedule.

`relearn_apply.apply_relearn_fix` / `enable_enforcement` were reached from the removed prompt but keep their other callers (the API apply route, `tj optimize`), so they stay. The only genuinely orphaned code was the onboard-local `_run_relearn_first_fix` and a now-unused `TYPE_CHECKING` import; both are removed. `_lens_review_url` is kept, renamed `_review_inbox_url`, since the pointer still needs the URL.

No flag or test depended on answering the removed prompt: the ask was already skipped on non-interactive runs, so `tj onboard` piped or in CI behaves the same as before minus the scan.

## Tests / Verification

- `tests/unit/test_onboard_first_run.py`: the Claude Code nudge test now asserts a two-row list in web-dashboard-then-tokenmaxx order, that `tj context` / `tj quota-audit` are absent, and a new test asserts "Lens" never appears on this list in either daemon state.
- `tests/unit/test_onboard_relearn_activation.py`: rewritten. Was seven tests driving the scan/apply/enable orchestration; now pins the replacement — the pointer is short, points at `#/review`, degrades to "run `tj serve`" without a daemon, never calls `click.confirm`, never calls `compute_relearn_finding` / `apply_relearn_fix` / `enable_enforcement`, and a parametrized test asserts every removed string is gone from the module source.
- `tests/unit/test_onboard_add_project.py`: the "must not run the tail" guard repointed at the new function.

Full suite: `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` → **3822 passed, 1 skipped** in 168.58s. `ruff check tokenjam/` and `mypy tokenjam/` both clean.

Live verification: `tj onboard --claude-code --no-daemon --force --project demo` against a throwaway HOME seeded with six real transcripts renders the new two-row list and the new pointer, and completes without prompting.

## What's NOT in this PR
- A codebase-wide "Lens" → "web dashboard" rename. Two other user-facing "Lens (web UI)" strings exist (`cli/home.py`'s home screen and the onboard "Connection details" summary block); they are deliberately untouched here.
- The pre-existing em dash in the next-steps header copy ("already loaded — these work right now"). Not part of this change; flagged rather than swept in.

---

# Follow-up: the onboarding's colour palette

Two further commits, on the same theme of "the close of onboarding is doing too much at once" — this time visually. Walking `npx tokenjam` end to end, a single screen carried white, grey, pink, green, yellow, cyan, blue and red.

## Roughly half of those colours were never chosen by anyone

`console = Console()` in `tokenjam/utils/formatting.py` left **Rich's automatic highlighter on** (its default). Rich then re-coloured *prose* by pattern, independently of the markup we write:

| Pattern | What Rich did | Visible result |
|---|---|---|
| any number | bold cyan | `Max 20x plan` rendered as `Max 2` + cyan `0x`; `tj, version 0.6.1` as cyan `0.6` · `.` · cyan `1` |
| any path | magenta + bright magenta | `~/.config/tj/` then pink `config.toml` — **the pink** |
| any bare `/` in a sentence | pink | "needed for trim `/` cache-recommend `/` reuse", "nudges `/compact`" |
| `(` `)` `[` `]` | bold | `installed **(**SessionStart: resume\|compact**)**` |
| quoted strings | green | `[budget.anthropic] plan = "api"` |
| `...` | yellow | `Ingest secret: 17ccb4a0...` |

Stacked on top of the deliberate cyan banner / green success / yellow panel / blue links / red errors, that is eight colours on one screen with no meaning attached to half of them. It also actively mangled copy — the `20x` plan label is the clearest case.

## The fix

**1. `highlight=False` on both shared consoles.** Nothing is coloured now unless someone asked for it.

**2. One auditable palette**, `tokenjam/utils/theme.py`, as named Rich theme roles rather than raw colour names:

| role | style | means |
|---|---|---|
| `accent` | `#8b7fd4` | a string you can **type or click**: command, path, config key, URL |
| `brand` / `url` | same hue | so no screen ever carries two accents |
| `label` / `heading` | bold | structure is weight, not colour |
| `muted` | dim | secondary detail |
| `ok` | bold | success is a `✓` plus weight, deliberately **not** a colour |
| `warn` / `error` | yellow / red | the only remaining hues: a blocker, and a real failure |

The discipline is modelled on Claude Code's own CLI: prose is plain, one accent means exactly one thing, and colour past that accent is reserved for genuine state.

**3. Swept the flow a new user actually walks through** — onboarding, the banner, the home screen:

- **Success stops being green.** Every `[green]✓[/green]` / `[bold green]` becomes `[ok]`, and the lines that leaned on green *alone* to mean "this worked" now carry the ✓ ("✓ Claude Code observability configured.", "✓ You're set up."). Green was being spent on the least surprising outcome in the flow.
- **Yellow becomes a blocker signal only.** Kept on the restart panel, the PATH-shadow heads-up, verify failures, daemon-install failures. Dropped from three decorative uses: the "left your existing statusLine untouched" row (one coloured row inside an otherwise plain field list reads as a *failure*), the "showing your most-recent N sessions" note, and the "removed a hardcoded OTEL_RESOURCE_ATTRIBUTES" line — that one is something tj did *successfully*, sitting inside an all-dim block. The restart panel's title joins its border on `warn.strong`; a bold title over a yellow border split one signal into two half-emphasised ones.
- **Commands, paths and URLs take the accent.** Next-steps rows, the restart panel's `claude -c` / `claude --resume` / verify command, the written-config path, the PATH-fix paths, `/compact`, `claude --as <name>`, and the home screen's action list. Connection details stay uniformly dim — that block is a deliberately demoted footer, and accenting its paths would undo the demotion.
- **The plan menu accents only the choice key.** The digit is what you type, so it keeps an affordance; the plan names stop getting chopped up.
- **The banner moves from cyan to the accent**, so the wordmark and the body of the screen agree.
- Fixed a `[bold]` nested inside `[dim]`, which renders bold-dim: neither emphasis nor recession, just a smudge.

Also dropped the em dashes from the copy on every touched line — including the next-steps header em dash this PR previously flagged and left alone.

## Before / after

Before, one line of the plan menu (`cat -v`):

```
  ^[[1;36m4^[[0m^[[1m)^[[0m Max 2^[[1;36m0x^[[0m plan ^[[1m(^[[0m$^[[1;36m200^[[0m/mo subscription^[[1m)^[[0m
```

After:

```
  ^[[38;2;139;127;212m4^[[0m) Max 20x plan ($200/mo subscription)
```

Every SGR code the whole `--claude-code` onboarding emits, after:

```
'1'                    bold                'TokenJam'
'2'                    dim                 'v0.6.1'
'38;2;139;127;212'     accent              'tj serve'
'1;38;2;139;127;212'   brand               the wordmark
'4;38;2;139;127;212'   url                 'http://127.0.0.1:7391/'
'33' / '1;33'          warn                the restart panel, and only that
```

Seven variants, each with a stated job. No magenta, no cyan, no green, no bold-dim.

## Tests / Verification

- New `tests/unit/test_cli_palette.py` (19 tests) pins both halves: the highlighter stays off on both shared consoles, prose renders with **zero** SGR codes (the `Max 20x` split is asserted verbatim, alongside the path / bare-slash / bracket / quoted-string / `...` cases), `ok` carries no colour while `warn` and `error` keep theirs, `brand` and `url` share the accent hue, and `cmd_onboard.py` contains no raw `[green]` / `[yellow]` / `[red]` markup at all.
- Verified by rendering `tj onboard --claude-code` against a throwaway HOME and inventorying every SGR code the flow emits (the table above). `tj` and `tj status` re-checked by hand, since the highlighter change is global — cost figures in particular read better plain, having previously been a cyan/pink mix.
- Full suite: `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` → **3841 passed, 2 skipped**. Both skips are pre-existing optional-dependency skips (no `anthropic` / `openai` in this venv), not related to this change.

## What's NOT in this follow-up

- The `Lens (web UI)` → `web dashboard` rename. Still deliberately out of scope; this is a colour change, and the two remaining "Lens" strings keep their wording (they just pick up the new accent).
- Palette work on commands outside the onboarding path. `highlight=False` is global, so they all lose the accidental colour, but only the onboarding, banner and home screen were swept onto the named roles. `tj status` still uses green for "No active alerts" and yellow for an active one — that is state rather than success-of-an-action, so it is defensible as-is, but it has not been reviewed against the new palette.


anshss:

PR #604, opened 2026-07-26
feat(relearn): recovery-arc pricing, 9 more failure families, $5 write floor

anshss:
The relearn analyzer reported **$190 of past overspend** on a real 6,773-session corpus but claimed only **$97**. Investigating the gap turned up four problems — one of which was *understating* the figure by roughly half.

## Headline

| | before | after |
|---|---|---|
| claim (`estimated_recoverable_usd`) | $96.96 | **$273.98** (+183%) |
| observation (`past_overspend_usd`) | $190.26 | **$331.87** |
| clusters | 56 | 45 |
| smallest offered write | cents | **$12.82** |

## 1. The recovery arc (the big one)

relearn charged every failure **exactly one** forced assistant turn. Right shape, wrong size, and never actually measured.

Walking the real step order across **914 failure-bearing sessions**: the median failure costs **2 turns** to recover from, mean 2.47. Per family it ranges from **0.97x** (a git branch that already exists — one clean retry) to **4.62x** (a stale-read race, where the agent re-reads, re-edits and races the linter again). The flat charge was halving every figure.

The arc is measured per occurrence at extraction time, scanning forward to the turn where the **same** tool succeeds again. A different tool succeeding mid-flail does not end the arc.

**The load-bearing detail is de-duplication.** Two failures a step apart have overlapping recovery windows, and billing each its own full detour charges the same turn twice — rule 27's double-count, inside one analyzer instead of between two. It is not small: the naive sum is 2.45x per occurrence against a true union of **2.07x**, so 15.3% would be the same turns counted repeatedly. A turn claimed by k failures is split 1/k between them.

Two deliberately conservative choices: a failure that never recovers inside the scan window is charged that session's **own** measured median (not the scan cap, not a global constant), and the archive/OTel lanes — spans, not ordered steps — fall back to the old 1.0 rather than inheriting a multiplier measured on a different lane. That is why the corpus-wide effect is **1.85x**, not the transcript lane's 2.07x.

## 2. Coverage: the families that were never named

The single most recurrent blocker in the corpus — a context-window overflow, **916 distinct sessions**, more than every other family combined — had no template, so it claimed nothing. New / widened families, each with a rung and a concrete fix:

`context_overflow` · `bash_timeout` · `bash_chained_approval` · `git_branch_exists` · `read_directory` · `read_too_large` · `edit_ambiguous_match`, plus widenings of `command_not_found` (the shell's other `<binary> not found` phrasing), `webfetch_domain_blocked` (the same block surfaces on the model call, not only on `WebFetch`) and `deferred_tool_cold` (`No such tool available`).

Every pattern is pinned in `test_relearn_families_and_boundary.py` against **verbatim corpus wording**, not a plausible-looking paraphrase — a family matching only a paraphrase is silently inert, which the existing `read_offset_malformed` ordering comment already documents as a real failure mode here.

## 3. Correctness: the cascade double-count

Claude Code marks the **siblings** of a parallel tool block as errored when one member fails. Those siblings never ran, taught the agent nothing, and forced no recovery turn of their own — the one recovery turn belongs to the member that actually failed, which was already counted. Clustering them both invented a pothole (`Cancelled: parallel tool call …` was the 4th-largest cluster, 127 sessions) and double-counted the real one.

Excluded at the same extraction-time gate that already drops human declines, along with **bare** permission prompts (the user's own allowlist; the same command succeeds once approved, so no agent-file rule changes the outcome). The chained-command variant is deliberately **kept**, via a negative lookahead, because chaining is what forced that prompt and un-chaining removes it — it reaches its own family instead.

## 4. Value floor

Netting only asked whether a rule breaks even. A rule clearing its own standing cost by $0.40 passes that — and then spends one of five scarce write slots plus a block of a file the user re-sends forever.

`MIN_NET_WRITE_USD` ($5) is applied **after netting and before the budget**, so a small family can never crowd out a large one. It is deliberately a floor on the write **offer** and never on the observation: a below-floor family keeps its past-overspend figure, its net claim and its copyable snippet, exactly like a budget-deferred one. Zeroing an incurred cost because our remedy is too small to write down is the "no action for this" / "this was free" conflation the analyzer's accounting exists to prevent. A family carrying no dollar figure is not held to the floor at all.

On the corpus this blocks 7 families and leaves 5 offered writes, each returning $12+.

## Boundary: relearn vs resend

The two look overlapping because on a coding corpus they name nearly the same physical tokens. Measured: the **median turn in a failure-bearing session is 96,064 prompt tokens, of which 95,919 (99.8%) are cache reads and 2 are fresh input.**

`RELEARN_RESEND_BOUNDARY` lives in `resend_tail.py` — the module that already owns the resend/downsize partition — and both analyzers quote it verbatim in their basis strings rather than paraphrasing, so the two cards cannot drift into two accounts of one line.

**The line is the counterfactual, not the token class:**

- `relearn` owns whether a call **exists**. Its fix deletes the turn, so the turn's whole cost goes with it, cache reads included — you do not pay a call's cache reads for a call you never make.
- `resend` owns how **big** a call is. Its levers shrink calls that still have to happen.

Splitting by token class instead would price relearn's claim at the ~2 fresh tokens a retry turn introduces, valuing the elimination of a rejected 96k-token call at a fraction of a cent. The residual overlap runs the other way — a retry turn is itself one of the later calls `resend` counts in its tails — measured at 5,735 of 275,537 calls (2.08%), and under 0.1% of resend's claim after its own in-scope and offloadable-share gates. Bounded, disclosed, and pinned by a test rather than left to be rediscovered.

## Tests / Verification

- New `tests/unit/test_relearn_families_and_boundary.py` (24 tests): every family against real corpus wording, ordering regressions for the pre-existing families, the cascade/permission filter incl. the chained-command carve-out, and both boundary invariants.
- 6 new value-floor tests in `test_write_budget.py`, including that a below-floor family does not consume a write slot and that an unpriced family is never held to a dollar floor.
- Full suite: **3845 passed, 1 skipped**. `ruff check tokenjam/` and `mypy tokenjam/` both clean.
- Four existing fixtures asserted on the OFFERED write path with cent-scale amounts, so the floor correctly refused them. Each was **scaled past the floor rather than the floor weakened**, keeping the assertion it was written to make, with the $/token rate held inside a real price band.

## What's NOT in this PR

- **The archive lane's unattributed cluster.** 1,058 error spans (340 sessions) carry an empty `status_message`, so `relearn_otel` falls back to the span `name` and produces clusters titled `Bash: gen_ai.tool.call` — the largest is $17.84 across 102 sessions. That is not one recurring pothole, it is a bucket of unrelated failures wearing a cluster's clothes, and it is the biggest remaining defensibility hole. Filed separately; it needs its own named aggregate (like the existing below-threshold residue) rather than a cluster, which changes the finding's field set and the render contract.
- **Re-fragmentation of the below-threshold residue** ($117 across 841 clusters). Coarsening the generic signature to the leading error line was measured and rejected: it recovers $63 but almost all of it lands in one `Bash: exit code <n>` bucket of genuinely unrelated failures, which is the same defect as the item above.

### DIFF
--- a/tests/integration/test_cost_proposals_api.py
+++ b/tests/integration/test_cost_proposals_api.py
@@ -202,13 +202,16 @@ async def test_cost_apply_workspace_writes_note_and_records_marker(app, client, 
     existing relearn apply path, then records the cost marker for delta-verify."""
     from tokenjam.core.optimize import relearn_apply as pa
 
-    # over_powered subagent fan-out on a premium model, in-window.
+    # over_powered subagent fan-out on a premium model, in-window. Sized past
+    # the $5 write floor (`write_budget.MIN_NET_WRITE_USD`): this test asserts
+    # the card is APPLY-CAPABLE, and the budget declines a permanent write for
+    # a couple of dollars, so a cent-scale seed no longer reaches that path.
     now = utcnow()
     for i in range(4):
         db.insert_span(make_llm_span(
             agent_id="claude-code-x", provider="anthropic", model="claude-opus-4-8",
             billing_account="anthropic", input_tokens=60_000, output_tokens=400,
-            cost_usd=0.5, session_id="s1", sub_agent_id=f"sa{i}",
+            cost_usd=15.0, session_id="s1", sub_agent_id=f"sa{i}",
             start_time=now - timedelta(days=2, minutes=i),
         ))
     # Home-anchored target allowlist: point Path.home() at tmp so the CLAUDE.md is "inside".
--- a/tests/unit/test_cost_proposals.py
+++ b/tests/unit/test_cost_proposals.py
@@ -1205,8 +1205,18 @@ def test_resend_persona_flows_through_the_report_dispatch():
     # resend adapter (it was the one adapter that didn't take persona).
     from tokenjam.core.optimize.cost_proposals import cost_proposals_from_report
 
+    from dataclasses import replace
+
     rep = _report()
-    rep.findings["resend"] = _resend_finding()
+    # Scaled past the $5 write floor (`write_budget.MIN_NET_WRITE_USD`): the
+    # report dispatch runs the write budget, which declines a permanent block
+    # for a 50-cent return, and this test is about persona threading rather
+    # than about whether a rule is worth writing. Tokens move with the dollars
+    # so the implied rate stays inside a real price band (CLAUDE.md rule 28).
+    rep.findings["resend"] = replace(
+        _resend_finding(),
+        past_overspend_usd=60.0, past_overspend_tokens=20_000_000,
+    )
     rep.persona = "claude-code"
     prop = {p.analyzer: p for p in cost_proposals_from_report(rep)}["resend"]
     assert prop.suggestion == "", "report persona must reach the resend adapter"
--- a/tests/unit/test_cost_proposals_model.py
+++ b/tests/unit/test_cost_proposals_model.py
@@ -197,7 +197,11 @@ def _subagent_finding(sub_agent_id="explore"):
         sessions_with_subagents=1, total_subagents=1, subagent_cost_usd=1.2,
         subagent_tokens=92_500, window_cost_usd=2.0, percent_of_cost=0.6,
         flagged_cost_usd=1.2, rows=[row], flagged=[row],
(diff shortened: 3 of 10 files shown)


anshss:

PR #606, opened 2026-07-26
chore: retire relearn's forward-savings claim, the last dollar-field exception

anshss:
The per-analyzer dollar-field contract deleted `estimated_recoverable_*` and `estimated_monthly_*` from every cost analyzer, but kept a documented carve-out on `RelearnCluster`/`RelearnFinding` and the Applied-tab render path: a relearn cluster's forward "you could recover $X" claim, gated on whether a fix template existed and netted against what a permanent rule costs to keep. On a real corpus the write-budget gate zeroed that claim for the large majority of clusters, rendering "$0.00 est./mo" for recurrences that had already cost real money.

Founder decision resolving the blockers previously named in the carve-out: a relearn cluster shows its PAST figure only. The forward claim is deleted from the card entirely, not merely de-emphasised.

## Summary
- Removes `estimated_recoverable_tokens/usd`, `estimated_monthly_tokens/usd`, `monthly_rate_basis`, `gross_monthly_tokens/usd`, and `gross_recoverable_tokens` from `RelearnCluster`/`RelearnFinding`. `past_overspend_tokens` now doubles as the write-budget's netting input (it held the same value as the deleted gross field).
- The net-of-standing-cost arithmetic in `write_budget.py` is untouched: it still decides `write_offered`/`advise_only`/`write_blocked_reason`/`payback_ratio`, it just never lands on a rendered forward-savings field.
- `relearn_apply.py`'s Applied-tab snapshot now stamps `past_overspend_usd`/`_tokens` instead of the retired monthly figure (a historical record, so this is a straight field swap, not a live re-measurement).
- `ui/index.html`: the Review inbox's recurring-mistakes row and its collapsed-tail summary drop the second "est./mo if fixed" line and the now-dead `monthlyUsdForDisplay`/`estMonthlyLine`/`combinedEstMonthly` helpers; `sortByEstMonthly` is renamed `sortByPastOverspend` since both the relearn and cost tabs now rank on the same `past_overspend_tokens` basis. The Applied tab's `appliedEstimate()` reads `verify.past_overspend_usd/tokens`.
- `runner.py`'s `report_from_dict` had a latent bug this surfaced: `_relearn()` passed retired `estimate_basis`/`estimate_confidence` kwargs to `RelearnFinding`, which raised and silently dropped the `"relearn"` key from every rehydrated report (round-trip test now covers this).
- `CLAUDE.md`: the per-analyzer dollar-field contract section is rewritten to state one canonical field with no carve-outs; the GREP TRAP callout and the numbered retirement sequence are removed since there is nothing left to retire.

## Tests / Verification
- Guards that used to assert the exception EXISTS are inverted to assert the retired names are absent, rather than deleted outright: `test_relearn.py`, `test_relearn_cli.py`, `test_relearn_archive_and_cost.py`, `test_write_budget.py`, `test_relearn_families_and_boundary.py`, `test_lens_ui_regression.py` (including a static-grep guard over `ui/index.html`), `test_onboard_relearn_activation.py`.
- `env -u FORCE_COLOR python3 -m pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/ -q` — 3853 passed, 1 skipped.
- `python3 -m ruff check tokenjam/` and `python3 -m mypy tokenjam/` — both clean.
- `node --check` on the extracted `<script type="module">` block — clean.

## What's also in this PR
Two small documentation-only fixes the founder asked to fold in here since this PR already touches `CLAUDE.md`:
- Renumbers the two Critical Rules that collided at 31/32 (the backfill `--since` windowed-DELETE rule and the transcript sessionId-dedupe rule become 33/34; nothing in the tree cites them by number, confirmed by grep) and updates the one cross-reference in the `[ingest]` config section accordingly.
- Adds a short note at the top of the Critical Rules section disambiguating this list's numbering from the workspace root `CLAUDE.md`'s separate "anti-pattern N" list (1-19), which the root file now mirrors.

## What's NOT in this PR
- The corpus-span basis every analyzer is due to land on — independent of this work, per the original brief.
- A stale `estimated_recoverable_*` reference in `cmd_optimize.py`'s `_render_resend` docstring (an unrelated analyzer's comment, predates this PR; left alone to keep the diff scoped to relearn).

**Merge order:** branches from `feat/relearn-family-coverage` (unmerged) since that branch heavily rewrites `build_proposals` and both basis strings in `relearn.py` — that branch must merge first.


anshss:

PR #607, opened 2026-07-26
docs: strip stale corpus measurements from CLAUDE.md, keep the durable lessons

anshss:
## Why

`CLAUDE.md` is always-on context: it is auto-loaded into every session, so every sentence in it is read as current fact. That makes it the right home for durable rules, conventions and architecture — and the wrong home for **empirical claims**. A measurement written into it goes stale silently, and the next session plans around it as though it still describes the product.

This is not hypothetical. The rule about `downsize` looking inert cited a specific dollar contribution to the Review-inbox rollup; the *same commit that wrote the rule* reworked the analyzer and moved that figure by two orders of magnitude. The number was obsolete the instant it was committed, and it read to a skimming session as a description of what `downsize` currently finds.

The fix is not to refresh the numbers — an updated number rots exactly the same way. It is to **strip the quantities and keep the lessons**, and to point at live code wherever a list or a threshold is the real source of truth.

## What changed (docs only — one file)

**Measurements stripped from the Critical Rules, mechanisms preserved in full.** Every rule keeps its causal explanation, its code pointers, its file/function names and its "always do X / never do Y" instruction. Where a measurement was load-bearing as *motivation*, it is replaced with a neutral phrase ("measured once on a real corpus") rather than a fresh figure. Rules touched: the disjoint-spans rule's sibling about matching token and dollar bases; the inert-analyzer rule; the avoidable-vs-observed-cost rule; the recorded-CWD measurement trap; the analyzer-horizon rule; and the transcript-dedupe rule.

**Enumerations replaced with pointers to the live source of truth**, because a hand-maintained list in a doc is itself a stale claim:
- the Project Overview analyzer list and the `tj optimize` registered-name list now point at `ANALYZER_REGISTRY` / `ANALYZER_ORDER` in `core/optimize/runner.py` (both were already missing shipped analyzers);
- the persona-gate rule now points at `PERSONA_DISABLED_ANALYZERS['claude-code']` in `runner.py` (the doc list had drifted from the map), and says explicitly not to keep a second copy here;
- the alert-engine notes now point at the `AlertType` enum and the `_build_channel` factory instead of carrying counts.

**Threshold values replaced with the constant names that hold them** — `_RETRY_LOOP_THRESHOLD` / `_RETRY_LOOP_WINDOW` / `_FAILURE_RATE_THRESHOLD` / `_FAILURE_RATE_WINDOW` / `_FAILURE_RATE_CHECK_INTERVAL` / `_SESSION_DURATION_DEFAULT` in `alerts.py`, the `SMALL_*` ceilings in `model_downgrade.py`, the `_TIERS` ladder in `cmd_tokenmaxx.py`, and the default-rate constants in `core/pricing.py`. A constant name is a code reference that stays true; a copied value is a claim that stops being true.

**Duplicate rule numbers resolved.** Two rules were numbered 31 and two were numbered 32, which makes any citation of either number unresolvable. The second pair (the windowed-backfill DELETE rule and the transcript-dedupe rule) is renumbered to 33 and 34; the list now runs 1-34 ascending with no duplicates. The one in-file cross-reference to the old number was updated. A repo-wide grep found no source comment or test citing either number, so nothing outside this file needed touching.

## Deliberately kept

A few numeric literals are code or copy references rather than measurements, and removing them would make the text wrong: the `"$0"` / `"$0.00"` strings quoted as examples of a misleading render, the `>100%` arithmetic bound, the hypothetical worked example in the persona-gate rule (an illustration, not an observation), and the exact assertion literal quoted in the grep-the-tests-first rule — that one names the string a test pins, so paraphrasing it would break the rule's own instruction.

## Test plan

- [x] Docs-only change; no code, no tests, no behaviour touched.
- [x] Verified the Critical Rules list runs 1-34 with no duplicate or skipped numbers.
- [x] Verified no source file or test cites the two renumbered rules.
- [x] Re-swept the file for dollar figures, percentages, thousands-separated counts and "N of M" phrasings; every surviving hit is a quoted literal or an arithmetic bound, listed above.


anshss:

PR #618, opened 2026-07-27
fix: format billion-scale tokens

### LINKED ISSUE #610: format_tokens caps at M, so billion-scale token figures render as '16885.9M'
**Severity:** Low

**Where:** `utils/humanize.py` `format_tokens()`, and the web UI's own `fmtTokens` in `ui/index.html`.

**Observed:** `format_tokens` has a 'k' tier and an 'M' tier and nothing above, so a billion-scale figure prints as '16885.9M'. Already reachable on main: the resend analyzer reports 23,976,965,630 recoverable tokens on a real 90d window. Now more visible since summarize's window-priced token figure is also billion-scale ('~16885.9M tokens across the 7084 session(s)' in `tj optimize` output). Reads as a formatting bug to any user and makes two large figures hard to compare at a glance.

**Fix direction:** Add a 'B' tier (and confirm the UI's `fmtTokens` matches, so the CLI and dashboard never disagree on how the same number reads). Check the existing tests for assertions on the current 'M' rendering before changing it — a green suite may be enforcing the current shape.

**Done when:** A billion-scale count renders as '16.9B' in both the CLI and the web UI, with a unit test pinning each tier boundary.

**Ref:** Found while verifying the summarize token/dollar basis fix in #589.

pollychen-lab:
## Summary

Adds a billion-scale branch to Python `format_tokens()` so CLI text renders billion-token figures as `16.9B` instead of `16885.9M`. The dashboard formatter on current `main` already had the `B` branch, so this brings the shared CLI formatter into parity and pins the boundary behavior in unit tests.

## Related issue

Closes #610

## Checklist

- [ ] Tests pass (`pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/`)
- [ ] Lint clean (`ruff check tokenjam/`)
- [ ] Type check clean (`mypy tokenjam/`)
- [x] Test spans use `tests/factories.py` (not raw `NormalizedSpan`) - N/A, no spans added
- [ ] Requested `@anilmurty` as reviewer (or @-mentioned him above)

## Validation

- `uv sync --extra dev`
- `uv run pytest tests/unit/test_formatting.py -q`
- `uv run pytest tests/unit/test_lens_ui_regression.py -q`
- `uv run pytest tests/unit/test_humanize.py tests/unit/test_formatting.py tests/unit/test_lens_ui_regression.py -q`
- `uv run ruff check tokenjam/utils/humanize.py tests/unit/test_formatting.py tests/unit/test_lens_ui_regression.py`
- `git diff --check`

### DIFF
--- a/tests/unit/test_formatting.py
+++ b/tests/unit/test_formatting.py
@@ -43,6 +43,15 @@ class TestFormatTokens:
     def test_millions_with_fraction(self):
         assert format_tokens(2_500_000) == "2.5M"
 
+    def test_just_below_billion_stays_millions(self):
+        assert format_tokens(999_999_999) == "1000.0M"
+
+    def test_billions(self):
+        assert format_tokens(1_000_000_000) == "1.0B"
+
+    def test_billions_with_fraction(self):
+        assert format_tokens(16_885_900_000) == "16.9B"
+
     def test_zero(self):
         assert format_tokens(0) == "0"
 
--- a/tokenjam/utils/humanize.py
+++ b/tokenjam/utils/humanize.py
@@ -11,8 +11,10 @@ from pathlib import Path
 
 
 def format_tokens(n: int) -> str:
-    """Human-size a token count: ``M`` at/above a million, ``k`` at/above a
-    thousand, otherwise the raw integer."""
+    """Human-size a token count: ``B`` at/above a billion, ``M`` at/above a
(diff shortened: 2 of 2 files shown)


anshss:
Thanks, this is a clean fix! merging it.

pollychen-lab:
> Thanks, this is a clean fix! merging it.

@anshss thanks

PR #619, opened 2026-07-27
fix(optimize): render <0.1× instead of 0.0× for tiny subscription multipliers

### LINKED ISSUE #590: tj optimize prints a 0.0x implied-value multiplier for very small subscription usage
## Problem

For subscription-plan users, `tj optimize` prints a line like:

```
Implied API value: $0.12 — about 0.0x your plan cost.
```

This comes from `cmd_optimize.py`, which computes `multiplier = w.total_cost_usd / fee` and renders it with one decimal place: `f"{multiplier:.1f}x your plan cost."`. Any multiplier below 0.05 rounds down to `0.0`, so genuinely small (but nonzero) usage prints as `0.0x`. That reads as if the tool computed nothing, or that the analyzed usage had zero value, which is misleading.

## Expected

Below a 0.1x multiplier, the line should render `<0.1x` rather than a rounded `0.0x`. At or above 0.1x, keep the existing one-decimal format (e.g. `1.3x`).

## Where to look

- `tokenjam/cli/cmd_optimize.py:739-745` — the subscription-plan branch:

```python
if fee and w.total_cost_usd > 0:
    multiplier = w.total_cost_usd / fee
    console.print(
        f"[dim]Implied API value: "
        f"[bold]{format_cost(w.total_cost_usd)}[/bold] — about "
        f"{multiplier:.1f}x your plan cost.[/dim]\n"
    )
```

## Done when

- Any subscription-plan multiplier below 0.1 renders as `<0.1x` instead of `0.0x`.
- Multipliers at or above 0.1 keep the current one-decimal rendering.
- A test covers the boundary (a multiplier just below 0.1 and one just above) so the threshold behavior doesn't regress.

syf2211:
## Summary

Subscription-plan users with very small implied API value relative to their plan fee saw a misleading `0.0×` multiplier because the header used one-decimal rounding. This adds a small formatter that renders `<0.1×` below the 0.1 threshold.

## Related issue

Closes #590

## Checklist

- [x] Tests pass (`pytest tests/unit/test_optimize_plan_multiplier.py`)
- [x] Lint clean (`ruff check tokenjam/cli/cmd_optimize.py tests/unit/test_optimize_plan_multiplier.py`)
- [x] Type check clean (`mypy tokenjam/cli/cmd_optimize.py`)

cc @anilmurty

### DIFF
--- a/tests/unit/test_optimize_plan_multiplier.py
+++ b/tests/unit/test_optimize_plan_multiplier.py
@@ -1,0 +1,17 @@
+"""Unit tests for subscription-plan multiplier formatting in cmd_optimize."""
+from __future__ import annotations
+
+from tokenjam.cli.cmd_optimize import _format_plan_multiplier
+
+
+def test_format_plan_multiplier_below_threshold_uses_floor_label():
+    assert _format_plan_multiplier(0.0) == "<0.1×"
+    assert _format_plan_multiplier(0.049) == "<0.1×"
+    assert _format_plan_multiplier(0.099) == "<0.1×"
+
+
+def test_format_plan_multiplier_at_or_above_threshold_keeps_one_decimal():
+    assert _format_plan_multiplier(0.1) == "0.1×"
+    assert _format_plan_multiplier(0.12) == "0.1×"
+    assert _format_plan_multiplier(1.34) == "1.3×"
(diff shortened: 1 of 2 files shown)


anshss:
nice catch, and the fix is scoped right! getting this merged.

PR #622, opened 2026-07-27
feat: move analyzer runs off the request path (stored report + scheduled scan + rescan)

anshss:
`GET /optimize` had no cache read at all. It called `build_report(...)` synchronously on the request thread, which dispatches every registered analyzer over the corpus. One of them is `relearn`, whose `run(ctx)` calls the same `compute_relearn_finding` that `core/optimize/relearn_store.py` exists to cache — that module's own docstring calls it "far too slow to compute per HTTP request." `fast=true` did not help: it dropped only `trim`, applied no timeout, and still ran every full-corpus scan. Three more routes did the same thing.

Measured on a real corpus the asymmetry was stark: the Review inbox painted instantly, because every `/relearn/*` GET is a pure store or ledger read, while the Dashboard's recoverable-waste panel and Budget-at-risk card took tens of seconds to minutes.

## Summary

- No HTTP request path runs an analyzer. Four routes now read a stored result plus its computed-at timestamp: `/optimize`, `/reuse/clusters`, `/cost/components`, `/cost/cache`.
- New `core/optimize/report_store.py` holds the report, deliberately shaped like `relearn_store` rather than as a second caching mechanism with its own semantics.
- Three run triggers, one mechanism: daemon boot, a scheduled interval, and a user-pressed rescan.
- Every analyzer-consuming surface gets a rescan control, an auto-rescan on a short interval, and a visible "computed N ago" stamp.
- A cold store is a distinct state from a computed-and-empty one, everywhere.

**Ingestion is out of scope and unchanged.** It keeps running continuously and keeps updating everything derived from it — traces, spans, sessions, maps, approach, timeline. Those surfaces stay live. Only the analyzer layer moved off the request path. The measured component bars on `/cost/components` are a plain spend query and still render on a cold scan; only the recoverable overlay waits for the store.

## Storage and trigger design

`report_store` mirrors `relearn_store`'s guarantees rather than reinventing them:

- atomic temp-file + rename writes, so a concurrent reader never observes a partial file;
- a non-blocking `_LOCK` plus a `_COMPUTING` flag, so an overlapping recompute no-ops immediately instead of blocking (blocking would move the cost, not avoid it);
- `trigger_background_recompute` spawns a daemon thread with a **fresh** `DuckDBBackend`, never the caller's live request connection, so a slow scan never contends with the live writer;
- the path resolves through `relearn_apply._storage_base_dir`, so `--config` / `storage.path` are honored and an in-memory-configured caller never leaks writes into a real `~/.tj`.

Triggers, all extending machinery that already existed rather than adding a parallel one:

| Trigger | Where |
|---|---|
| Daemon boot | `cmd_serve.py`'s lifespan, alongside the existing relearn and cost-proposal startup kicks |
| Scheduled interval | The same APScheduler block those two jobs use |
| User rescan | `POST /api/v1/optimize/rescan` — starts a background pass and returns immediately; it does not compute inline |

## Always-on rails (not staged)

All four live in config or in the store's own guards, never as sequential phases:

- **Overlap guard** — pressing Rescan twice, or pressing it while the scheduled job runs, costs one corpus pass.
- **Rate limit** — `[optimize] scan_min_rescan_seconds` (default 60). A request inside the floor is answered `throttled` with the stored result untouched.
- **Kill switch** — `[optimize] scan_enabled = false` stops the daemon scanning on its own; it never re-enables inline computation, because nothing does. `scan_ui_poll_seconds = 0` separately disables the browser's auto-rescan.
- **Configurable cadence and window** — `scan_interval_hours`, `scan_window_days`. The window travels with the stored result, so a surface labels its figures with the window they were observed over rather than whatever picker the reader's screen is set to.

## Cold vs empty

This is the load-bearing part. `report_status` separates four states, and the render layer keeps them apart:

- **cold** (`never_run` / `error`) — nothing has ever been computed. `/optimize` returns `report_available: false` with **no report body at all**; an empty `findings` dict would read as "every analyzer ran and found nothing." The recoverable totals on `/cost/components` and `/cost/cache` go `None`, not `0.0`.
- **computing** — the last completed result keeps rendering with a scanning indicator. Panels never blank mid-scan.
- **ready** — the only state in which an empty-state string is honest.

In the UI: `recoverableTiles` short-circuits before building any tile on a cold store, so no caller can render a zeroed grid. The Dashboard's `No recoverable candidates.` moved onto the not-cold branch; cold renders "Not computed yet" with a rescan link, and an error-only store adds "this is not a report of zero waste." Budgets-at-risk shows an em dash rather than `0` — "0 budgets at risk" off an un-run scan is the same false reassurance. Under a daemon, `tj optimize`, `tj report --reuse` and `tj route` each say "not computed yet" instead of rendering an empty report.

## Surfaces that got a rescan control

- Dashboard — recoverable-waste band (the band label carries the ScanBar; Budgets-at-risk reads the same scan, so the two panels can never disagree about freshness).
- Optimize view — in the filter row, plus a note stating that findings came from the scan's own window while the picker scopes the spend chart.
- The Review inbox already had this pattern and is unchanged; the new `ScanBar` copies it.

Auto-rescan runs on the server-published cadence, gated on `document.visibilityState` so a hidden tab never asks the daemon to scan.

## Tests / Verification

New `tests/integration/test_analyzers_off_the_request_path.py` (20 tests):

- Every analyzer-consuming route, **warm and cold**, against a registry where every analyzer is wrapped in a tripwire. The cold case is the one that matters — "nothing stored, so just run it this once" is exactly how the inline dispatch would return. Plus a direct guard that `build_report` itself is unreachable from any route.
- A cold store reports `never_run`, withholds the report body, and publishes `None` rather than `0.0` for every recoverable total, while `total_cost_usd` still renders because ingestion is untouched. A scan over an *empty* corpus reports `ready` with `0` — the distinction, asserted directly.
- Overlapping recomputes produce exactly one corpus pass, verified by holding the lock from a thread and confirming the second call returns immediately rather than blocking, plus the endpoint's in-flight and rate-limit branches and the throttle's boundary conditions.

Existing tests that asserted live-dispatch behaviour were **inverted rather than deleted** (Critical Rule 23): `/cost/components must not run relearn` became `must not run ANY analyzer`, and `fast=true skips trim` became `fast=true skips nothing, and the fast and slow payloads are the same report`. Tests wanting findings now warm the store the way `tj serve` does at boot. `conftest.py` gains a per-test store-isolation fixture — the store resolves off `storage.path`'s parent, so tests sharing the session-wide fake HOME shared one file, making every cold assertion order-dependent.

UI changes are guarded by static-grep regressions in `test_lens_ui_regression.py` (no JS runner in CI), and the extracted module block passes `node --check`.

- `pytest tests/unit/ tests/integration/ tests/synthetic/` — **4015 passed, 5 skipped** (plus the 20 new tests).
- **Live verification against a real `tj serve` + real DuckDB**, not just ASGI: cold reports `never_run` with `total_recoverable_usd: null` while the live component bars show `$0.225`; `POST /optimize/rescan` returns in **18ms**; an immediate second press is declined; once ready, `GET /optimize` serves in **20ms** with 10 findings and zero analyzers dispatched on any request thread. A separately-booted `tj serve` confirmed the boot kick writes `optimize_report.json` and the route reports `status: ready` with the full envelope.

## What's NOT in this PR

- **Ingestion**, per the scope: untouched, still continuous.
- **A stored report per (window, agent) combination.** There is one scan over `[optimize] scan_window_days`. `/optimize`'s `since`, `agent_id` and `finding` params are still validated and echoed back, but no longer select the data — the response states the window it was actually computed over. Per-window scans would multiply the corpus passes this PR exists to eliminate.
- **`GET /metrics`**, which still queries the DB per request. It runs no analyzers, so it is outside this guarantee, though it has its own uncached cost.
- **MCP's optimize tool** returns the envelope verbatim under daemon mode, so a machine consumer sees `report_available: false` on a cold store. No bespoke MCP rendering was added.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -77,6 +77,46 @@ include_captured_content = false
   protocol = "http"            # http | grpc
 ```
 
+## Analyzer scans
+
+The optimize analyzers are **never run on an HTTP request**. Several of them scan
+your full corpus and take tens of seconds to minutes, so a page that computed
+them inline would hang for exactly that long. Instead, the `tj serve` daemon
+computes the report in the background and every surface — the Dashboard's
+recoverable-waste panel, Budgets at risk, the Optimize view, `tj optimize` under
+a running daemon — reads the stored result plus the time it was computed.
+
+A scan runs on three triggers: once when the daemon starts, on the interval
+below, and whenever you press **Rescan** in the web UI.
+
+```toml
+[optimize]
+scan_enabled            = true   # false: the daemon never scans on its own.
+                                 #   Rescan still works; nothing ever computes
+                                 #   inline on a request either way.
+scan_interval_hours     = 6.0    # cadence of the background scan
+scan_window_days        = 30     # lookback the scan observes. Stored with the
+                                 #   result, so a surface labels its figures
+                                 #   with the window they came from.
+scan_min_rescan_seconds = 60     # floor between rescans that actually re-run
+                                 #   the analyzers; 0 disables the limit
+scan_ui_poll_seconds    = 300    # how often an open page asks for a rescan;
+                                 #   0 turns the browser's auto-rescan off
+```
+
+Overlapping scans never stack: a rescan pressed while one is already running is
+a no-op, and the panel keeps showing the last completed result with a scanning
+indicator rather than blanking.
+
+**Before the first scan completes, surfaces say "not computed yet" — not `0` and
+not "nothing found."** Those are different claims, and only the first one is true
+on a fresh install.
+
+`[optimize]` also holds each analyzer's sensitivity threshold (`min_cluster_instances`,
+`cache_efficacy_threshold`, and friends). Every default matches the module constant
+it replaces, so an omitted key changes nothing; see `OptimizeConfig` in
+`tokenjam/core/config.py` for the current list and each field's analyzer.
+
 ## Budget limits
 
 Budget limits merge per-field: each agent inherits default limits unless it explicitly overrides them.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -94,3 +94,23 @@ def _tj_guard_real_home_db(monkeypatch):
         original_init(self, config)
 
     monkeypatch.setattr(db_mod.DuckDBBackend, "__init__", _guarded_init)
+
+
+@pytest.fixture(autouse=True)
+def _tj_isolated_analyzer_report_store(tmp_path, monkeypatch):
+    """Give every test its own analyzer-report store.
+
+    `report_store.default_report_path` resolves off `storage.path`'s parent, so
+    two tests that both build a default `TjConfig` share one file under the
+    session-wide fake HOME. That makes cold-vs-warm assertions order-dependent:
+    a test asserting the COLD state ("nothing computed yet") would pass alone
+    and fail after any earlier test warmed the store. Since the cold state is
+    load-bearing here — a cold store must never render as a zero or an absence
+    claim — the isolation has to be per-test, not per-session.
+    """
+    from tokenjam.core.optimize import report_store
+
+    store = tmp_path / "optimize_report.json"
+    monkeypatch.setattr(
+        report_store, "default_report_path", lambda config=None: store,
+    )
--- a/tests/integration/test_analyzers_off_the_request_path.py
+++ b/tests/integration/test_analyzers_off_the_request_path.py
@@ -1,0 +1,369 @@
+"""The architectural guarantee: no HTTP request path runs an analyzer.
+
+Three claims, each of which has been violated in production code before:
+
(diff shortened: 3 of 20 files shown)


anshss:
Superseded: every commit from this branch is now contained in the branch behind PR #620, which carries the display-layer half of the same work (honest not-yet-known states, per-panel reads, the data-span window ladder, and the loading skeletons). The compute-layer change here only makes sense alongside those, so the two are shipping as one reviewable unit rather than as two PRs that each half-work.

Nothing is dropped: the stored-report layer, the scheduled and on-demand rescan, the cold-vs-empty distinction, the MCP fix so a cold store no longer reads as "no waste found", and the lossless report round-trip all land via #620.

PR #625, opened 2026-07-27
fix(config): honor TJ_CONFIG at every find_config_file() call site

anshss:
## Summary

`find_config_file()` was called bare across the CLI, API, and MCP server, silently ignoring the `TJ_CONFIG` env var and falling back to `SEARCH_PATHS` discovery — even at call sites that already held a `TJ_CONFIG`-aware config object from `load_config()`. This caused split-brain behavior: e.g. a budget update could write to a different file than the one config was actually read from, or `tj doctor` could report "No config file found" while other checks used a config that loaded fine.

- Adds `resolve_config_path()` as the single source of truth for "which config file is this process using" (explicit override, then `TJ_CONFIG`, then `SEARCH_PATHS`).
- Migrates every call site that needs write/report/subprocess-handoff behavior onto it: `cmd_budget.py`, `api/routes/budget.py` (both endpoints), `cmd_doctor.py`, `cmd_mcp.py`, `mcp/server.py`'s `setup_project`, and `pricing.py`'s config-section lookup (previously a partial, non-shared `TJ_CONFIG` check).
- `cli/home.py`'s bare `tj` landing screen keeps its two direct `find_config_file()` calls as a deliberate, documented exception — it renders before config is validated, so it needs a non-raising `TJ_CONFIG` check rather than `resolve_config_path()`'s fail-loud contract on a bad override.
- `cmd_onboard.py`'s write paths are intentionally asymmetric: the existing-config check now honors `TJ_CONFIG` (so onboard correctly reports "Config already exists" instead of double-writing `.tj/config.toml` alongside a `TJ_CONFIG`-pointed config), but the actual write target stays a fixed search-path location by design — onboard creates new configs, it doesn't update wherever `TJ_CONFIG` happens to point.
- Adds a regression guard (`test_config_discovery_guard.py`) that greps the package for bare `find_config_file()` calls outside the two documented exceptions, so a new call site can't reintroduce this silently.

## Test plan

- [x] `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` — 4362 passed, 5 skipped
- [x] `ruff check tokenjam/` — all checks passed
- [x] `mypy tokenjam/` — no issues in 229 source files
- [x] Verified `origin/main` was already current with this branch (no merge needed)

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -1118,7 +1118,7 @@ async def test_post_budget_zero_clears_limit(db):
     app = create_app(config=cfg, db=db, ingest_pipeline=pipeline)
     transport = httpx.ASGITransport(app=app)
 
-    with patch("tokenjam.api.routes.budget.find_config_file", return_value="/fake/tj.toml"), \
+    with patch("tokenjam.api.routes.budget.resolve_config_path", return_value="/fake/tj.toml"), \
          patch("tokenjam.api.routes.budget.write_config"):
         async with httpx.AsyncClient(transport=transport, base_url="http://test") as c:
             resp = await c.post("/api/v1/budget", json={"scope": "my-agent", "daily_usd": 0})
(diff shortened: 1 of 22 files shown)


anshss:

PR #626, opened 2026-07-27
Fix model-rate lookup (5-30x mispricing) + scope the filesystem-reading analyzers with --projects-root

anshss:
Two independent correctness bugs, both of which made a surface assert something its data did not support: the cost engine silently mispriced most real-world model names, and `--db` isolated the database but not the analyzers that read the filesystem.

## Summary

- Generalize the pricing lookup so dated, gateway-prefixed and dotted model names resolve to their real rates instead of falling through to the flat default; backfill three genuinely missing models.
- Surface `spans.pricing_source` (already stamped at ingest, read by nothing) in both cost views, so an estimated dollar figure is distinguishable from a quoted one.
- Introduce a first-class filesystem scope (`--projects-root`) that `deadweight`, `relearn` and `summarize` honor, and define what an explicit `--db` means for them.

## Cost engine priced most real models wrong

**Symptom.** A replay of nine real agent runs (21,562 LLM calls, nine distinct model strings) priced seven of the nine models wrong by 5-30x. Gemini 2.0 Flash billed ~5x too high, Claude Opus 4.1 ~30x too low, GPT-4.1 and Claude 3.7 Sonnet ~5-7x too low. Only two model strings matched exactly.

**Root cause.** The arithmetic was fine; `get_rates` never found the row. Two structural gaps in its fallback chain:

1. The date-suffix stripper understood only Anthropic's compact `-YYYYMMDD`, so OpenAI's dashed `-YYYY-MM-DD` (`gpt-4o-2024-08-06`, `gpt-4.1-2025-04-14`) never fell back to its bare table entry — even when that entry already existed.
2. Nothing stripped a gateway routing prefix, so `anthropic/claude-opus-4.1` (the form LiteLLM and OpenRouter emit) never reached the table at all, and its dotted version segment did not match the table's dashed `claude-opus-4-1` key either.

The failure is silent: one log warning per process, then a plausible-looking dollar figure computed at the flat default rate.

**Fix.** `_lookup_candidates` now composes four transforms — routing prefix, release date in either published shape, context tag, dotted version segment — over the names they produce, so combinations resolve without being enumerated by hand. Both `get_rates` and `classify_pricing_source` inherit the fix from the one choke point they already share, so the rate and its provenance cannot drift apart. Three models genuinely absent from the table are backfilled at their published rates (Claude 3.7 Sonnet, GPT-4.1, Gemini 2.0 Flash); the fallback itself stays reachable, so a truly unknown model still classifies as `default_fallback` rather than being resolved to something plausible.

**The silence was the other half of the bug.** `core/pricing_coverage.py` reads the existing `pricing_source` column back and turns it into one sentence naming the models priced at the default rate, shared by `GET /api/v1/cost` and `tj cost` so the two cannot word it differently. It distinguishes "asked, nothing unpriced" from "could not ask", and per Critical Rule 22 it never renders a zero or restates an estimate as a quoted price.

## `--db` did not isolate the filesystem-reading analyzers

**Symptom**, all reproduced against a freshly created, completely empty throwaway database:

- The review inbox showed recurring-mistake entries carrying real file paths from an unrelated project — the machine's global Claude Code history, not the served database.
- The recurring-fix modal defaulted its "where to write it" target to a real path on the operator's machine, so one Approve click in what looks like a sandboxed review view would write a real file outside the database being served.
- `/api/v1/optimize?fast=true` took ~27-33s on a literally empty store. Same root cause, and it is a ~30s blank first paint on every dashboard load including brand-new installs.

**Root cause.** `deadweight`, `relearn` and `summarize` read the filesystem as their evidence source by design — the database holds spans, the filesystem holds the transcripts, MCP config and always-loaded prompt files those three reason about. That split is correct. What was missing was any way to say WHICH disk: no `AnalyzerContext` field, no flag, only the undocumented `TJ_CLAUDE_PROJECTS_ROOT` env var. So `--db` implied an isolation it never provided.

**The contract**, defined in `core/optimize/scope.py` and documented in `docs/configuration.md`. Precedence, first match wins:

1. `--projects-root PATH` / `[optimize] projects_root` — scan exactly that root.
2. `TJ_CLAUDE_PROJECTS_ROOT` — unchanged, now documented.
3. **An explicit `--db` suppresses the filesystem scans.** Naming a database names the corpus you want reasoned about; reading a second, unrelated one into the same report is the bug. This is also what makes an empty or scoped store paint immediately instead of walking a large transcript tree to conclude nothing.
4. Otherwise `~/.claude/projects` — a run with no flag, no env var and no `--db` is unchanged.

The scope is resolved **once** in `build_report`, the way the persona already is, and carried on `AnalyzerContext`; an analyzer that re-derives a root of its own is how the leak existed in the first place. Suppression is never silent: the report carries `filesystem_scan_skipped_reason` (round-tripped through `report_from_dict`, so a report rebuilt from the daemon cache still knows) and `tj optimize` prints it, so every surface can tell "scanned, found nothing" from "did not scan" — the empty-state distinction in root anti-pattern 22.

**Write paths are covered too, not just read scope.** The user-global apply target follows the same Claude home the findings came from, so the two halves of a card can no longer describe different machines. And relearn's distill cache — a second cache beside `relearn_store.default_cache_path`, missed when that one was threaded through `_storage_base_dir`, and firing only after a successful distill LLM call, which is why it went unnoticed — now resolves through the same helper instead of writing real files under the real `~/.tj` from an isolated config. The daemon's background relearn recompute is a second entry point into the same scan, so it consults the scope too; otherwise the machine's global tree would leak back in through the cache the served routes read.

**One incidental fix worth naming.** The default root resolves lazily. A module-level `Path.home()` is baked at import and survives a later `HOME` repoint, which is how a test that isolates itself carefully still ends up scanning the developer's real transcript tree — the same trap `deadweight._global_config_path` already documents.

## Tests / Verification

New, written before the fixes:

- `tests/unit/test_pricing_lookup_forms.py` — one case per lookup form, plus every distinct model string in the benchmark corpus pinned to the rate it must resolve to, and a guard that the default fallback stays reachable. 27 of 38 failed before the fix.
- `tests/unit/test_pricing_coverage.py` — the unpriced-model note: unmeasured vs. clean vs. populated, parameterised SQL, immutability.
- `tests/unit/test_analyzer_scope.py` — the resolver: every precedence step, the no-escape rule for a custom root, lazy home resolution.
- `tests/unit/test_analyzer_scope_isolation.py` — the contract through the analyzers themselves, since a correct resolver is worth nothing if an analyzer still reaches for `Path.home()`. Asserts no filesystem call happens at all under an explicit `--db`, that a scoped-but-empty root reports no skip, that a normal run still scans, and covers the apply target and the distill cache.

Full CI suite: `4422 passed, 5 skipped` (`tests/unit/ tests/synthetic/ tests/agents/ tests/integration/`). `ruff check tokenjam/` and `mypy tokenjam/` clean (231 source files). `sdk-ts`: 27/27.

## What's NOT in this PR

- **No UI rendering of the two new signals.** Both `pricing_coverage` on `GET /api/v1/cost` and `filesystem_scan_skipped_reason` on the report ship as API fields with the CLI rendering them; the Lens panels are left for a follow-up so this PR stays reviewable. The data is what was missing — nothing computed it before.
- **The suppression rule is a deliberate behaviour change** for anyone running `tj serve --db /custom/path` as their normal setup: those runs lose the three filesystem analyzers unless they add `--projects-root`. Flagging it explicitly since it is the one part of the contract a reader might want to argue with. The alternative — scoping to a directory derived from the database's own path — is harder to explain and would silently scan somewhere the user never named.
- **`trim`** also expands the catalog's `~` paths (`prompt_bloat.py` reads the same `global_paths`), but it is not one of the three analyzers named in the report and gating it would widen the diff into a fourth analyzer's persona/prerequisite logic. Its scan is catalog-bounded rather than a corpus walk, so it contributes neither the leak nor the latency.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -155,6 +155,15 @@ why, and the report carries `filesystem_scan_skipped_reason` so every surface ca
 tell **"scanned, found nothing"** from **"did not scan"**. A run with no flag, no
 environment variable and no `--db` is unaffected.
 
+**The scope bounds writes too, not just reads.** The review inbox's Approve step
+refuses any target outside the scope's home directory — with
+`--projects-root /tmp/demo-home/.claude/projects` that root is `/tmp/demo-home`,
+so a scoped daemon can neither be talked into writing to your real `~`, nor into
+escaping the scope with a `..` path or a symlink pointing out of it. The
+suggested target and this check resolve from the same scope, so what the card
+proposes is always something Approve will accept. Without a scope the allowed
+root is your real home, exactly as before.
+
 ## Budget limits
 
 Budget limits merge per-field: each agent inherits default limits unless it explicitly overrides them.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -117,6 +117,44 @@ on a fresh install.
 it replaces, so an omitted key changes nothing; see `OptimizeConfig` in
 `tokenjam/core/config.py` for the current list and each field's analyzer.
 
+### Which filesystem the analyzers read
+
+Three analyzers — `deadweight`, `relearn` and `summarize` — take their evidence
+from the **filesystem**, not from the telemetry database. That is by design: the
+database holds spans, while the transcripts, the MCP configuration and the
+always-loaded prompt files those three reason about live on disk. By default
+they read `~/.claude/projects` and your agent homes, which is what you want for
+a normal run.
+
+`--projects-root PATH` says which filesystem instead:
+
+```bash
+tj optimize --projects-root /path/to/fixtures/.claude/projects
+tj serve --db /tmp/demo.duckdb --projects-root /tmp/demo-home/.claude/projects
+```
+
+The equivalent config key is `[optimize] projects_root`. Resolution order, first
+match wins:
+
+1. `--projects-root` / `[optimize] projects_root`
+2. the `TJ_CLAUDE_PROJECTS_ROOT` environment variable
+3. **an explicit `--db`, which scopes the filesystem scans OFF** — see below
+4. `~/.claude/projects`
+
+**Passing `--db` suppresses the filesystem scans unless you also pass
+`--projects-root`.** Naming a database names the corpus you want reasoned about,
+so reading a second, unrelated corpus off the machine into the same report is a
+bug, not a feature: served against a throwaway database, those analyzers would
+otherwise surface findings carrying real file paths from unrelated projects, and
+the review inbox's "where to write it" target would point at a real location
+outside the database you opened. Suppression is also why an empty or scoped
+database paints immediately instead of walking a large transcript tree first.
+
+Suppression is never silent. `tj optimize` prints which analyzers did not run and
+why, and the report carries `filesystem_scan_skipped_reason` so every surface can
+tell **"scanned, found nothing"** from **"did not scan"**. A run with no flag, no
+environment variable and no `--db` is unaffected.
+
 ## Budget limits
 
 Budget limits merge per-field: each agent inherits default limits unless it explicitly overrides them.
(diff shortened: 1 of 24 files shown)


anshss:

PR #628, opened 2026-07-27
[feature] Detect streamed calls that closed without a usage payload (stream-usage analyzer)

anshss:
A streamed response reports its token usage only in a **final** payload. Two ordinary things stop that payload arriving, and when it does not arrive the call is recorded with no token counts at all — which is indistinguishable from a call that cost nothing. The spend total silently reads low, and no surface says so.

## Summary

- Every path that watches a stream now stamps the signature on the span: the OpenAI and Anthropic provider patches, and a new SSE tap in the proxy.
- A new `stream-usage` analyzer reports the gap, names the affected call sites and provider, estimates the unrecorded spend with its derivation stated, and emits the **provider-correct** remediation snippet.
- The estimate is a **data-quality** figure and is kept structurally out of the recoverable-waste total.
- The analyzer is gated off for the `claude-code` persona; this is an SDK-persona failure mode.

## Why the payload goes missing

| Provider | Where usage lives | What stops it |
|---|---|---|
| OpenAI-compatible | a trailing chunk with an empty `choices` list | emitted **only** when the request carried `stream_options={"include_usage": true}`, and only if the iterator is drained to that last chunk |
| Anthropic | the trailing `message_delta` (`get_final_message()`) | a caller that breaks out of the loop or is disconnected never reaches it — `get_final_message()` then raises rather than returning counts |

The two remediations are therefore different, and offering the wrong one is advice that cannot work. The card emits the right one per detected provider.

## Detection

Three new `TjAttributes` — `tokenjam.llm.streaming`, `stream_usage_reported`, `stream_content_chunks` — stamped by:

- **`sdk/integrations/openai.py`** — `_StreamWrapper` already tracked the usage chunk; it now also counts content-bearing chunks and stamps the three attributes in its `finally`, so an abandoned iterator records the gap rather than vanishing.
- **`sdk/integrations/anthropic.py`** — same, plus `get_final_message()` raising because the stream was not drained is now read as *the answer* (usage missing) instead of being allowed to escape `__exit__`. `__iter__` / `__next__` count content events.
- **`proxy/stream_usage.py`** (new) — an SSE tap that watches the relayed bytes. It is line-buffered because a provider chunk boundary can fall mid-JSON, and a per-chunk parse would miss a split usage payload and report a healthy stream as broken. Its `finally` runs on client disconnect too, which is the case that under-counts. Pass-through stays sacred: the tap alters no byte and swallows its own errors.

The tap is armed only for a request that actually asked to stream — a non-streamed response reports usage in the body and cannot exhibit the gap.

One trap worth naming: Anthropic's `message_start` also carries a `usage` block, but its `output_tokens` is a placeholder for a response that has not been generated yet. Accepting any usage block would mark **every** interrupted stream complete, i.e. silently disable the whole analyzer. There is a test for exactly that.

## The estimate and its derivation

Each missing call is sized at the **median per-call usage of the completed streams for the same model in the same window** — all four token types, including both `cache_tokens` and `cache_write_tokens` — and priced through `core.cost.calculate_cost`. The derivation is printed on the card, not merely implied.

Two degrade rules:

- A call site with **no completed peer** to size against gets `None` on **both** the token and the dollar field, never zero on one and a number on the other (Critical Rule 28's symmetric degrade). The unpriced part is then disclosed in words in the basis string rather than folded in as zero.
- A stream that produced **no content** is not counted: it under-counted nothing, and flagging it would inflate the gap with noise.

## Accounting: why this is not a saving

This is the unusual part of the PR and the part most worth reviewing.

What the analyzer measures is spend that **already happened and was already billed**. Capturing the usage payload makes the spend figure *correct*; it does not make it *smaller*. A figure derived that way must never reach the recoverable-waste total, whose whole meaning is "money that could have been avoided" — mixing populations there is a first-order bug per Critical Rules 27 and 30.

It is kept out **structurally**, via the only two mechanisms that could pull it in:

1. The finding carries **no `past_overspend_*` field**. That name is the canonical avoidable-spend field, and `past_overspend_rollup` is generic over analyzers — it picks a figure up by field presence. The quantity here is named `undercounted_usd` / `undercounted_tokens` instead.
2. `stream-usage` is **absent from `COST_ANALYZERS`**, the independent second selection surface feeding the Review inbox, so no cost proposal is ever minted from it.

A test builds a report end to end (`build_report` → `cost_proposals_from_report` → `past_overspend_rollup`), asserts the gap figure is real and non-zero, and asserts the rollup total cannot see it. A second test asserts the *serialized* payload has no `past_overspend_usd` key, because the Overview's recoverable-waste band keys off presence in the payload rather than on the dataclass — the dataclass-level assertion alone would not have covered the surface that mints a tile.

A `DATA_QUALITY_NOTE` rides on the finding as a dataclass default (the `MODEL_DOWNGRADE_CAVEAT` device) so a renderer cannot drop it, and the CLI prints it verbatim rather than summarising it.

## Persona gating

`stream-usage` is added to `PERSONA_DISABLED_ANALYZERS["claude-code"]` with an inline reason, and runs for `sdk` / `mixed` / `unknown`. An interactive coding agent does not proxy end-user streaming chat and its harness drains its own streams, so the failure mode does not arise; and the fix is a request-side change to code that constructs the provider call, which is on the harness's side of the actionable ceiling. No mirror in `COST_ANALYZERS` is needed precisely because the name was never in it.

The gate pin in `tests/unit/test_persona_analyzer_gate.py` is updated deliberately, as that file requires.

## Tests / Verification

`tests/unit/test_stream_usage_capture.py` (14) — the three observation paths:

- OpenAI stream with no usage chunk → gap recorded; with a trailing usage chunk → complete; **abandoned mid-iteration** → gap recorded.
- Anthropic stream interrupted before `message_delta` → gap recorded; drained to completion → complete; a raising `get_final_message()` never escapes the wrapper.
- SSE scanner: OpenAI without/with the usage chunk, Anthropic's `message_start` placeholder rejected, `message_delta` accepted, a usage payload split across chunk boundaries still found, malformed bytes never raise.
- The tap records on client disconnect, and relays bytes byte-for-byte unmodified.

`tests/unit/test_stream_usage_analyzer.py` (16) — detection, per-provider remediation, the estimate and its derivation, both cache token types in the baseline, the implied-rate check, symmetric degrade, and the accounting separation above.

The implied-rate test is worth a look: it asserts `usd / tokens` equals the **exact blend** the basis advertises, not merely that it lands in a price band. A band spanning cache-read to output rates happily admits a 2x basis mismatch (dollars multiplied per call, tokens left one-time) — verified by mutating the code and watching the band-only version stay green.

Each new guard was mutation-checked: the Anthropic placeholder trap, the symmetric degrade, and the basis mismatch each fail their test when the implementation is broken.

Full suite (`pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/`) green. `ruff check tokenjam/` and `mypy tokenjam/` clean.

## What's NOT in this PR

- **`_StreamWrapper.__next__` on the OpenAI patch bypasses the generator entirely**, so a consumer that calls `next()` directly never records usage and never ends the span. Pre-existing, orthogonal to this change, and fixing it means restructuring that wrapper — left alone rather than folded into a data-quality PR.
- **No backfill of historical spans.** The signature is only present on streams observed after this lands; the analyzer says so in its empty-state hint rather than implying a clean window.
- **No web-UI card.** The finding is deliberately kept off the recoverable-waste band, and giving a non-recoverable figure its own Overview surface is a product decision, not a mechanical follow-on. The CLI renders it and `--json` carries it in full.
- **No minimum-calls floor for `sdk` / `mixed` windows.** Every affected call site is reported; whether a floor is wanted is a judgement best made against a real SDK corpus, which this repo's corpus is not.

### DIFF
--- a/tests/unit/test_persona_analyzer_gate.py
+++ b/tests/unit/test_persona_analyzer_gate.py
@@ -47,6 +47,7 @@ NO_LEVER = set(PERSONA_DISABLED_ANALYZERS["claude-code"])
 # is a product decision and must be made here deliberately.
 assert NO_LEVER == {
     "cache", "cache-recommend", "placement", "trim", "verbosity", "script", "reuse",
+    "stream-usage",
 }
 
 # `placement` is the odd one out: it is not an ANALYZER_REGISTRY name at all
--- a/tests/unit/test_stream_usage_analyzer.py
+++ b/tests/unit/test_stream_usage_analyzer.py
@@ -1,0 +1,379 @@
+"""The `stream-usage` analyzer: report the blind spot, never bank it as savings.
+
+The load-bearing constraint here is accounting, not detection. What this
+analyzer measures is spend that ALREADY HAPPENED and was never recorded —
+capturing it makes the spend figure correct, it does not make it smaller. A
+figure derived that way must never reach the recoverable-waste total, whose
+whole meaning is "money that could have been avoided". The tests below pin
+that separation structurally (no `past_overspend_*` field, absent from
+`COST_ANALYZERS`, invisible to the rollup) rather than by inspecting copy,
+because copy is what drifted the last three times this class of bug shipped.
+"""
+from __future__ import annotations
+
+from datetime import timedelta
+
+import pytest
+
+from tokenjam.core.config import StorageConfig, TjConfig
+from tokenjam.core.db import InMemoryBackend
+from tokenjam.core.optimize import build_report
+from tokenjam.core.optimize.analyzers.stream_usage import StreamUsageFinding
+from tokenjam.core.optimize.cost_proposals import (
+    COST_ANALYZERS,
+    cost_proposals_from_report,
+    past_overspend_rollup,
+)
+from tokenjam.core.pricing import get_rates
+from tokenjam.otel.semconv import TjAttributes
+from tokenjam.utils.time_parse import utcnow
+from tests.factories import make_llm_span, make_session
+
+WINDOW_DAYS = 30
+
+
+@pytest.fixture
+def db():
+    backend = InMemoryBackend()
+    yield backend
+    backend.close()
+
+
+@pytest.fixture
+def cfg(tmp_path):
+    return TjConfig(version="1", storage=StorageConfig(path=str(tmp_path / "t.duckdb")))
+
+
+def _stream_attrs(*, usage_reported: bool, content_chunks: int = 4) -> dict:
+    return {
+        TjAttributes.STREAMING: True,
+        TjAttributes.STREAM_USAGE_REPORTED: usage_reported,
+        TjAttributes.STREAM_CONTENT_CHUNKS: content_chunks,
+    }
+
+
+def _seed_stream(db, *, agent_id, provider, model, usage_reported,
+                 input_tokens=0, output_tokens=0, cache_tokens=0,
+                 cache_write_tokens=0, content_chunks=4, session_suffix="0",
+                 days_ago=1):
+    """One streamed call, complete or truncated, plus its session."""
+    started = utcnow() - timedelta(days=days_ago)
+    session_id = f"{agent_id}-{session_suffix}"
+    db.upsert_session(make_session(
+        agent_id=agent_id, session_id=session_id,
+        plan_tier="api", started_at=started,
+    ))
+    db.insert_span(make_llm_span(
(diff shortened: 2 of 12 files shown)


anshss:

Issue #615, opened 2026-07-26
test_ping fails only in the full-suite run: ordering pollution hides a real isolation bug


anshss:
**Severity:** Medium

**Where:** `tests/unit/test_ping.py` (4 tests: `test_ping_http_mode_confirmed_delivery_exits_zero`, `test_ping_direct_mode_reports_local_db`, `test_ping_json_output_is_machine_readable`, `test_ping_honors_custom_agent_id`); the pollution source is elsewhere in `tests/`.

**Observed:** `pytest tests/unit/test_ping.py` alone → 10 passed. `pytest tests/` (full suite) → the same 4 tests FAIL. Confirmed pre-existing by stashing all working-tree changes and re-running the full suite at HEAD: identical 4 failures, 3801 passed. The failure mode is that `tj ping` reports delivery to a real daemon instead of the expected mode: `assert 'my-real-agent' in output` fails against `'⚠ Could not confirm local interception (the SDK failed to initialise a tracer).\n✓ Delivered to tj serve at http://127.0.0.1:7391 (confirmed received).'`, plus `WARNING tokenjam.otel:provider.py:56 Span export failed: Connection Error: Connection already closed!`.

So an earlier test in the full-suite ordering leaves global OTel/tracer or endpoint state behind (a configured exporter, a patched provider, or a live `TJ_*` endpoint env var) that `test_ping` then picks up. The tests are not wrong; they are the victim.

**Why it matters beyond the 4 tests:** the suite is currently red on every full run, which trains readers to ignore a red suite. It also means the isolation guarantee is already broken — whatever leaks here can leak into any other test that touches OTel, and the next such test to be written will fail mysteriously.

**Fix direction:** Bisect the ordering with `pytest -p no:randomly tests/ -x` plus `--deselect` halving, or run `pytest tests/ --lf` after a full run to find the first polluter. Likely candidates: a test that calls `tokenjam.otel.provider` setup without tearing the global tracer provider down, or one that sets a `TJ_*`/OTLP endpoint env var without `monkeypatch`. Fix the POLLUTER (autouse fixture resetting the global provider, or monkeypatch instead of raw `os.environ`), not the victim — do not add skips or xfails to `test_ping.py`.

**Done when:** `pytest tests/` is green, `pytest tests/unit/test_ping.py` stays green, and the polluting test resets whatever global state it sets (ideally via an autouse fixture so the class of bug cannot recur).

[...]

PR #629, opened 2026-07-27
fix(tests): restore SDK tracer after full_stack fixture teardown

### LINKED ISSUE #581: test_ping.py depends on suite run order: the full_stack fixture leaks a stale SDK tracer
## Problem

Four tests in `tests/unit/test_ping.py` fail if `tests/integration/` runs before `tests/unit/`,
even though they pass when the unit suite runs alone or in the order the project's own
documented test commands declare (`pytest tests/unit/ tests/synthetic/ tests/agents/
tests/integration/`). That means CI is currently green only because of a specific ordering
assumption, and anything that changes it (a parallel/sharded runner, `pytest tests/integration`
run on its own first during local debugging, a future reordering of the test commands) would turn
these tests red for reasons unrelated to whatever change triggered the run.

## Root cause

`tokenjam/sdk/agent.py` holds a module-level global:

```python
_tracer = trace.get_tracer("tokenjam.sdk")
```

used by `AgentSession`, `record_llm_call`, `record_tool_call`, and `record_outcome_span` to start
spans (`tokenjam/sdk/agent.py`, e.g. line 26 and the `_tracer.start_span(...)` call sites).

`tests/integration/test_full_pipeline.py`'s `full_stack` fixture reassigns this global directly to
bind it to a fixture-local `TracerProvider` for the duration of the test:

```python
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(ocw_exporter))
agent_mod._tracer = provider.get_tracer("tokenjam.sdk")
...
yield stack
provider.shutdown()
db.close()
```

The teardown shuts the local provider and closes its backing `InMemoryBackend`, but it never
restores `agent_mod._tracer` to what it was before the test ran. Once any test using this fixture
executes, `agent_mod._tracer` stays pointed at that now-shut-down, disconnected provider for the
rest of the process — every later call to `record_llm_call` anywhere in the same pytest session
routes through a tracer whose processor/exporter/backend no longer exist.

This is a real problem for `tj ping` (`tokenjam/cli/cmd_ping.py`): it attaches a local
`_ProofExporter` to `trace_api.get_tracer_provider()` (the OTel API's *global* provider) and then
emits a span via `record_llm_call`. If `agent_mod._tracer` has been permanently redirected by an
earlier test, the emitted span never reaches the API's global provider or the `_ProofExporter` at
all — `proof.captured` stays empty, so `intercepted` is `False`.

That flips these four assertions in `tests/unit/test_ping.py`:
- `test_ping_http_mode_confirmed_delivery_exits_zero` — checks `"intercepted a test span"` in
  output.
- `test_ping_direct_mode_reports_local_db` — checks `"intercepted a test span"` in output.
- `test_ping_json_output_is_machine_readable` — checks `payload["intercepted"] is True`.
- `test_ping_honors_custom_agent_id` — checks `"my-real-agent"` in output, which only appears in
  the `intercepted`-branch message (`cmd_ping.py`'s non-intercepted fallback message doesn't
  include the agent id at all).

Contrast with `tests/agents/test_record_outcome.py`, which does the same kind of tracer
reassignment but correctly saves and restores the original:

```python
original = _agent_mod._tracer
_agent_mod._tracer = _provider.get_tracer("tokenjam.sdk")
try:
    ...
finally:
    _agent_mod._tracer = original
```

`test_full_pipeline.py`'s `full_stack` fixture is missing the equivalent save/restore.

## Where to look

- `tokenjam/sdk/agent.py` — the module-level `_tracer` global (around line 26) and its use sites.
- `tests/integration/test_full_pipeline.py` — the `full_stack` fixture (its teardown, after
  `yield stack`).
- `tests/agents/test_record_outcome.py` — the save/restore pattern to mirror (`original =
  _agent_mod._tracer` / `finally: _agent_mod._tracer = original`).
- `tests/unit/test_ping.py` — the four affected tests listed above.

## Suggested fix

Have `full_stack` save `agent_mod._tracer` before reassigning it and restore the original value in
teardown (after `provider.shutdown()`), the same way `tests/agents/test_record_outcome.py` already
does. That isolates the fixture's tracer rebinding to just the tests that use it, regardless of
what runs afterward.

## Reproduction

Confirmed by execution. Running:

```
pytest tests/integration/test_full_pipeline.py tests/unit/test_ping.py
```

yields `4 failed, 17 passed`, and the four failures are exactly the four tests named above:
`test_ping_http_mode_confirmed_delivery_exits_zero`, `test_ping_direct_mode_reports_local_db`,
`test_ping_json_output_is_machine_readable`, `test_ping_honors_custom_agent_id`.

## Done when

- `pytest tests/integration/ tests/unit/` (reversed from the declared CI order) is green.
- `agent_mod._tracer` is restored to its prior value in `full_stack`'s teardown, so the fixture
  can't leak state into later tests regardless of suite order.
- A check exists (or the reproduction command above is documented) that would catch a future
  reintroduction of the same class of leak.

syf2211:
## Summary

The `full_stack` integration fixture rebinds `agent_mod._tracer` to a fixture-local `TracerProvider` but never restored the original tracer after shutdown. Later tests in the same pytest process (notably `tests/unit/test_ping.py`) then routed spans through a shut-down provider, causing four ping tests to fail when `tests/integration/` ran before `tests/unit/`.

Save and restore the module tracer in teardown, matching the pattern already used in `tests/agents/test_record_outcome.py`.

## Related issue

Closes #581

## Checklist

- [x] Tests pass (`pytest tests/integration/test_full_pipeline.py tests/unit/test_ping.py` — 21 passed; issue reproduction command now green)
- [x] Lint clean on changed file (`ruff check tests/integration/test_full_pipeline.py` — only pre-existing unused-import warnings in file header)
- [ ] CLAUDE.md updated (if architecture changed) — N/A
- [ ] Test spans use `tests/factories.py` (not raw `NormalizedSpan`) — N/A

## Tests

- `pytest tests/integration/test_full_pipeline.py tests/unit/test_ping.py` — **21 passed** (was 4 failed before fix)
- `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` — 4340 passed, 5 failed (pre-existing unrelated failures in `test_backfill_progress.py`, `test_quickstart.py`, `test_sdk_config_discovery.py`)

## Notes

Reviewed with composer-2.5: **APPROVE**. Minimal three-line fix; teardown order (`provider.shutdown()` then restore) is correct.

### DIFF
--- a/tests/integration/test_full_pipeline.py
+++ b/tests/integration/test_full_pipeline.py
@@ -133,7 +133,9 @@ def full_stack():
 
     ocw_exporter = TjSpanExporter(pipeline)
 
-    # Create a local TracerProvider (not global) and bind the SDK tracer to it
+    # Create a local TracerProvider (not global) and bind the SDK tracer to it.
+    # Save/restore so later tests in the same pytest process keep the default tracer.
+    original_tracer = agent_mod._tracer
     provider = TracerProvider()
     provider.add_span_processor(SimpleSpanProcessor(ocw_exporter))
     agent_mod._tracer = provider.get_tracer("tokenjam.sdk")
@@ -156,6 +158,7 @@ def full_stack():
     yield stack
 
     provider.shutdown()
+    agent_mod._tracer = original_tracer
     db.close()
 
 


anilmurty:
Thanks @syf2211 — this is a clean, correct fix (matches #581 exactly and is Rule #11-compliant: it swaps the module-level `_tracer` reference rather than re-calling `set_tracer_provider`). It turns out to be the same fix as #630 (@Axiya3749), filed against the duplicate issue #615, which just merged with a before/after reproduction (4 failures → 0). Closing this as a duplicate so we don't double-apply — genuinely appreciated, and your #633/#634 are in review. The shared bug is fixed on main.

PR #633, opened 2026-07-28
feat: add mocked-HTTP coverage for contributor_label_guard I/O

### LINKED ISSUE #608: contributor_label_guard's network paths have no mocked-HTTP test coverage
**Severity:** Low

**Where:** `.github/scripts/contributor_label_guard.py` (`_api_request`, `fetch_issue`, `remove_label`, `add_labels`, `main`); tests at `tests/unit/test_contributor_label_guard.py`.

**Observed:** The contributor label guard shipped in #574 with 30 unit tests covering only the pure parsing and decision functions — those tests deliberately scoped to logic with no network calls. The I/O layer is therefore untested: `_api_request` is written to never raise on an HTTP error status (fail-safe by design), which means an auth failure, a rate-limit response, or a GitHub API schema change would all be swallowed silently and the guard would no-op without signal. That is the correct runtime behavior for a label-hygiene workflow, but it also means a regression in the I/O layer is undetectable by CI.

**Fix direction:** Add mocked-HTTP tests (monkeypatch `urllib.request.urlopen` or inject a transport seam) covering: 200 with the expected payload; 401/403; 429 with a rate-limit header; 404 on a missing issue; a payload missing keys the code reads. Assert the no-raise contract holds AND that each failure path is distinguishable in the workflow log, so a silent no-op can be told apart from a genuine no-op. Do not change the fail-safe behavior itself.

**Done when:** Each listed response class has a test; the no-raise contract is pinned; failure paths emit a distinguishable log line.

**Ref:** Gap named as deliberately deferred when the contributor label guard shipped in #574.

syf2211:
## Summary

Add mocked-HTTP unit tests for the GitHub API I/O layer in `.github/scripts/contributor_label_guard.py`, covering the response classes that were deliberately left untested when the guard shipped in #574.

## Motivation

The contributor label guard is designed to fail-safe: HTTP errors must never crash the workflow. That behavior was correct at runtime but untested in CI, so regressions in the I/O layer would be silent. Issue #608 asks us to pin that contract with mocked HTTP tests.

## Changes

- Extend `tests/unit/test_contributor_label_guard.py` with mocked `urllib.request.urlopen` coverage for `_api_request`, `fetch_issue`, and `fetch_issue_timeline`
- Add handler-level tests for `_handle_claim` / `_handle_release` that assert distinguishable stdout/stderr log lines on success, no-op, and warn paths
- Cover 200, 401, 403, 404, 429, non-JSON error bodies, timeline pagination, and issue payloads missing expected keys
- No production code changes — fail-safe behavior is unchanged

## Tests

```bash
python3 -m pytest tests/unit/test_contributor_label_guard.py -v
python3 -m ruff check tests/unit/test_contributor_label_guard.py
```

Result: **45 passed**, ruff clean.

## Notes

- Handler tests stub `fetch_issue` / `remove_label` / `add_labels` for log-contract coverage; the lower-level `urlopen` mocks cover the actual HTTP transport.
- Reviewed with composer-2.5: **APPROVE**.

Fixes #608

cc @anilmurty for review


anilmurty (changes requested):
Thanks @syf2211 — the transport-level mocks are faithful and the timeline-pagination test is nicely load-bearing. One real gap left (matches Greptile's P2):

`test_handle_claim_auth_failure...` injects `fetch_issue → (403, None)`. But GitHub returns a JSON body on a 403, and `_api_request` parses it to a dict — so the real handler gets `(403, {"message": ...})`, skips the `404 / not-a-dict` branch, and falls through to log **"no-op: issue #N is already closed"** rather than the "not found (status 403)" your test asserts. So the test doesn't exercise the auth-failure path it names, and it hides that a 403 is logged indistinguishably from a genuine closed-issue no-op — which is exactly the log-distinguishability #608 asked to pin.

Ask: inject the realistic `(403, {"message": ...})` shape. If that surfaces the misleading "already closed" log, that's a real diagnostic gap — either tighten the handler to branch on `status >= 400` before the state check, or scope it to a follow-up (the guard shipped in #574, so a handler change may be beyond this test-only PR — your call). Ping me when updated.

anilmurty:
@syf2211 - are you still planning to work this?

syf2211:
Thanks for the review — pushed 5ec8d5b addressing the auth-failure gap:

1. **Realistic 403 payload** — `test_handle_claim_auth_failure` now injects `(403, {"message": "Forbidden"})` instead of `(403, None)`.
2. **Handler fix** — `_handle_claim` and `_handle_release` now branch on `status != 200` before checking `issue.state`, so a 403 logs `no-op: issue #N API error (status 403)` instead of the misleading "already closed" message.

`python3 -m pytest tests/unit/test_contributor_label_guard.py -v --noconftest` — 45/45 pass.

syf2211:
Yes — this PR is still active and ready for re-review. The latest commit addresses the structured 403 payload concern from the previous review thread. Happy to make further changes if anything else comes up.

syf2211:
Yes — still on it. The 403 structured-payload regression from 5ec8d5b is covered and unit tests pass (45/45). Let me know if you'd like anything else before merge.

anilmurty:
thanks @syf2211 - appreciate the contribution!

PR #634, opened 2026-07-28
fix(cli): fail fast when --export-templates omits reuse finding

### LINKED ISSUE #578: tj optimize --export-templates gives a misleading reason when the reuse analyzer wasn't run at all
## Problem

`tj optimize <findings...> --export-templates` is supposed to export Reuse analyzer skeletons. The handler calls the export helper unconditionally whenever `--export-templates` is passed, without checking whether `reuse` was one of the analyzers actually requested:

```python
# --export-templates branch: write the Reuse Markdown skeletons and exit,
# without rendering the HTML report. Needs direct DB access (to fetch the
# planning completion text), so it's local-mode only.
if export_templates:
    _export_reuse_templates(report, conn=conn, config=config, agent=agent)
    return
```

`report` only contains results for the analyzers named on the command line (this is how `build_report()` works generally: `findings=None` runs every registered analyzer, a list runs only the named ones). So running `tj optimize downsize --export-templates` never runs the `reuse` analyzer, and `report.findings.get("reuse")` is `None`.

`_export_reuse_templates` then prints:

```python
finding = report.findings.get("reuse")
if finding is None or not finding.clusters:
    console.print(
        "[dim]No repeated planning detected — nothing to export. Try a "
        "longer [bold]--since[/bold].[/dim]"
    )
    return
```

This message asserts the DATA contains no repeated planning, but the analyzer that would detect that never ran. It also tells the user to try a longer `--since` window, which cannot help since the actual problem is that `reuse` wasn't in the requested finding list at all.

The `--validate` flag on the same command avoids exactly this class of bug by restricting itself at the Click layer:

```python
@click.option("--validate", "validate_finding", default=None,
              type=click.Choice(["downsize"]),
              ...)
```

`--export-templates` is currently a bare boolean flag (`is_flag=True`) with no such restriction, so it can be combined with any set of `findings` positional arguments, including ones that don't include `reuse`.

## Expected

Fail fast with a clear error when `--export-templates` is passed and `reuse` is not among the selected findings, instead of silently running the export path against a report that never computed the `reuse` finding. (Note: this can't literally mirror the `--validate` pattern shown above — `--validate` is a valued option using `type=click.Choice([...])`, while `--export-templates` is a bare `is_flag=True` boolean, so the same `Choice`-based restriction doesn't apply here. The check has to compare `export_templates` against the resolved `findings` list instead.)

## Where to look

- `tokenjam/cli/cmd_optimize.py` — the `--export-templates` flag declaration (`click.option("--export-templates", ...)`, around line 106) and the branch that calls `_export_reuse_templates` unconditionally (around line 337).
- `tokenjam/cli/cmd_optimize.py` — `_export_reuse_templates` (around line 1184), whose "No repeated planning detected" message is only accurate when `reuse` actually ran.
- `--validate`'s restriction (`type=click.Choice(["downsize"])`, around line 109) is referenced above only to illustrate the class of bug it avoids — its exact mechanism (a `Choice`-typed value) isn't reusable for a boolean flag.

## Done when

`tj optimize downsize --export-templates` (or any combination of `findings` that excludes `reuse`) fails fast with a clear, actionable error telling the user to include `reuse` in the finding list, instead of printing "No repeated planning detected." `tj optimize reuse --export-templates` (and `tj optimize --export-templates` with no findings named, which runs every analyzer including `reuse`) continue to work exactly as before.

syf2211:
## Summary

When `tj optimize <findings> --export-templates` is run without `reuse` in the finding list, fail fast with a clear error instead of printing a misleading "No repeated planning detected" message.

## Motivation

`--export-templates` exports Reuse analyzer Markdown skeletons, but the export path was reachable even when the `reuse` analyzer never ran (e.g. `tj optimize downsize --export-templates`). The downstream helper treated a missing reuse finding as "no repeated planning," which is incorrect and suggests trying a longer `--since` window — advice that cannot help.

Fixes #578

## Changes

- Add an early guard in `cmd_optimize` after resolving analyzer names: if `--export-templates` is set and `reuse` is not among the selected analyzers, raise `click.ClickException` with an actionable message.
- Add `test_optimize_export_templates_requires_reuse_finding` to assert the failure mode and ensure the misleading message is not shown.

## Tests

```bash
python3 -m pytest tests/integration/test_cli.py::test_optimize_export_templates_requires_reuse_finding tests/integration/test_cli.py::test_optimize_export_templates_writes_markdown -v
```

Both tests pass locally.

## Notes

- `tj optimize --export-templates` (no findings named) continues to work — `findings=None` runs all registered analyzers including `reuse`.
- `tj optimize reuse --export-templates` and `tj optimize downsize reuse --export-templates` are unaffected.


anilmurty (changes requested):
Thanks @syf2211 — the fail-fast itself is clean and the error copy is good and actionable. One thing left before this is mergeable, and it's the most common real case.

The guard runs **before** the persona gate. `reuse` is in `PERSONA_DISABLED_ANALYZERS["claude-code"]`, and `build_report` strips it in the persona pass (`runner.py:306`) even when a user explicitly names `reuse`. So on a Claude-Code-dominant window, `tj optimize reuse --export-templates` passes your guard, `build_report` drops `reuse`, and `_export_reuse_templates` prints the exact "No repeated planning detected — nothing to export" message this PR set out to kill — for Claude Code users specifically. (Ironically, the command your new error suggests, `tj optimize reuse --export-templates`, is itself that broken path there.)

Two asks:
1. Account for the persona gate — e.g. consult `disabled_analyzers_for_persona(persona)` before the misleading message, so a persona-disabled `reuse` returns the same actionable error instead of "nothing to export."
2. Add a regression test on a **claude-code-persona window** (plan set so `reuse` is disabled) that names `reuse` explicitly and asserts the misleading message is NOT shown. The current test uses a plan-less config, so persona resolves to `unknown` and it never hits this path.

Everything else — CI green, honesty, scope — is good. Ping me when it's updated.

anilmurty:
@syf2211 - are you still planning to work this?

syf2211:
Thanks for the review — pushed b4f8cf8 addressing both asks:

1. **Persona gate** — after `dominant_persona` resolves, `--export-templates` now fails fast when `reuse` is in `disabled_analyzers_for_persona(persona)`, with a persona-specific error (instead of suggesting `tj optimize reuse --export-templates`, which is itself broken for claude-code windows).
2. **Regression test** — `test_optimize_export_templates_reuse_disabled_for_claude_code_persona` seeds a claude-code-dominant window and asserts both `optimize reuse --export-templates` and bare `optimize --export-templates` fail without the misleading "nothing to export" message.

`python3 -m pytest tests/integration/test_cli.py -k export_templates -v --noconftest` — 3/3 pass locally. Ready for re-review.

syf2211:
Hi @anilmurty — friendly ping for re-review. Both asks from your last review are addressed in b4f8cf8 (persona gate + claude-code regression test). CI is green. Thanks!

syf2211:
Yes — still on it. Both review items from b4f8cf8 are addressed and CI is green on my side. Happy to make any further tweaks if needed.

anilmurty:
thanks for  working on this @syf2211

PR #645, opened 2026-07-29
ui enhancements

anshss:
## Summary

A pass over the Lens dashboard's structure, spacing and copy. No new analyzers and no
changes to how any number is derived; this is presentation only.

**Persona, pinned above every page.** The detected persona moves into a dropdown in the
header, and it gates the nav rather than merely labelling it: a persona with no data on
this machine now says so instead of rendering an empty screen, and views that cannot
apply to the selected persona are hidden rather than shown broken. `/api/v1/persona`
backs it with the detected persona plus per-persona session counts. One spelling of the
coding-agent persona is used throughout, replacing the `claude_code` / `claude-code`
split that had accumulated across `framing.py`, `cmd_doctor.py`, `cmd_onboard.py` and
`onboard_verify.py`.

**Navigation.** Optimize moves into the Improve lens, above Sessions, and loses its
in-title FAQ link.

**One spacing scale.** Every surface here is a card body, a row inside one, a head strip
or a table cell, so they now share one horizontal inset token and vary only in vertical
density. The shared `.panel-body` / `.panel-row` / `.opt-tile` / `.empty` classes replace
the ad-hoc style attributes that had drifted apart, which is what let a panel and its own
loading skeleton stop matching. Two table-column traps are fixed and written down: a cell
with `overflow-wrap:anywhere` collapses to a one-character column without an explicit
`min-width` floor, and a `td` left at `nowrap` holding long text dictates the whole
table's width.

**Copy.** Em dashes are gone from the dashboard copy, the framing leads with past
overspend, group descriptions are derived rather than hand-written, and the Source label
no longer overlaps Trigger.

**Testing.** `tests/unit/test_lens_ui_regression.py` is deleted: 344 tests asserting on
literal substrings of `index.html`. None of them executed any code, they pinned wording
and markup structure rather than behaviour, and ordinary visual iteration broke
assertions that guarded nothing (one used a `useEffect` dependency array as a slice
anchor). What it protected that mattered is kept and strengthened in `test_ui_offline.py`:
the NUL-byte integrity check, plus a new `node --check` over the extracted module script.
That parse is worth more than the 344 it replaces, since a syntax error in this file
renders a blank page with no server-side error while still matching every substring test
one could write. The rule is now recorded in `CLAUDE.md`: if a UI behaviour is worth
guarding, extract it into a pure function and execute it.

## Test plan

- [x] `test_ui_offline.py`, `test_lens_dashboard_states.py`, `test_lens_select_all_behaviour.py`, `test_framing.py`
- [x] `node --check` over the extracted module script
- [ ] Visual pass with `tj serve` across both personas, including the no-data state


anshss:
Folded into #646 — this branch is merged into that one, so the whole batch lands as a single merge for review.

Leaving this PR **open**: its commits are preserved by the merge commit on the other branch, so it will auto-flip to Merged when that lands. Nothing to close by hand.

One ask: please don't push anything else to `context-audit-page` in the meantime. A new commit here changes the head SHA, breaks the ancestry, and would strand this PR open with a stale diff.

PR #648, opened 2026-07-29
Fix #642: tj backfill status counts ingested sessions under the daemon

### LINKED ISSUE #642: tj backfill status reports "0 already ingested" after a successful backfill
After onboarding a fresh install (which reported `Backfilled: 26 new · 26 total sessions`), running `tj backfill status` shows:

```
Compared 1285 transcript files at ~/.claude/projects → 26 distinct sessions (deduped on internal sessionId).
  0 already ingested · 0 missing · 0 skipped (no assistant turn).
  9 file(s) had no readable session id.
✓ Every on-disk session with usage is ingested.
```

**Problem:** the three buckets (`0 already ingested · 0 missing · 0 skipped`) sum to **0**, which both fails to account for the **26 distinct sessions** and contradicts the earlier backfill (`26 new`) and the final `✓ Every on-disk session with usage is ingested`. A user reads "0 already ingested" as "it didn't work."

**Candidate root causes to check:**
1. **DB-visibility mismatch** — `backfill status` may be querying a different / empty DB than the one backfill wrote to (e.g. the daemon holds the write-lock, or a path mismatch), so it sees 0. If so, the number is genuinely wrong.
2. **Counter/label bug** — "already ingested" only counts a narrow sub-case, and the real ingested set lands in an uncounted bucket.

**Expected:** the displayed buckets sum transparently to the distinct-session count (26), the 9 no-session-id files are accounted for separately, and the "ingested" figure reflects reality (~26).

anilmurty:
`tj backfill status` reported `0 already ingested · 0 missing · 0 skipped` on a fully-ingested install — the buckets summed to 0 and contradicted the preceding backfill (`26 new`) and the same command's own `✓ Every on-disk session with usage is ingested`. A user reads that as "it didn't work."

Closes #642

## Summary
- Root cause was a **DB-visibility mismatch** (candidate #1), not a counter bug.
- Route the disk-vs-DB anti-join through the daemon when the CLI is on the read-only HTTP shim, so the buckets reconcile in serve mode.

## Root cause
Onboarding starts `tj serve`, which holds the DuckDB write-lock. DuckDB allows one writer OR many readers across processes with no read-only escape hatch, so the CLI's `open_db()` fails on the lock and `main.py` falls back to the read-only HTTP shim `ApiBackend`.

`reconcile_claude_code` computed the ingested set with `getattr(db, "conn", None)`. `ApiBackend` has no `.conn`, so it saw `None`, returned early with `ingested_sessions == 0` and an **empty `missing` list** — which then tripped the misleading "everything is ingested" line. The `disk_sessions` count comes straight off disk (correct); only the DB-derived buckets were blanked.

## Fix
- Add `POST /api/v1/sessions/ingested-ids` — returns the subset of a candidate id list present in the `sessions` table (parameterised SQL, chunked under DuckDB's bind-parameter ceiling, api-key gated).
- Add `ApiBackend.fetch_ingested_session_ids` (+ a small `_post` helper).
- `reconcile_claude_code` uses `conn` when present and otherwise dispatches to `fetch_ingested_session_ids`. `_has_billable_turns` (the missing/skipped split) is filesystem-based and already worked regardless of backend, so those buckets now populate correctly in serve mode too.

Honest framing preserved: this reports what is genuinely ingested; no claim was strengthened.

## Tests / Verification
- New load-bearing regression `test_reconcile_counts_ingested_sessions_when_daemon_holds_the_lock` in `tests/unit/test_transcript_sync.py`: ingests 3 sessions into a real `DuckDBBackend`, stands up a **real in-process uvicorn server** over that DB, then runs `reconcile_claude_code` against the `ApiBackend` shim (no `.conn`). Asserts `ingested_sessions == 3` and that `ingested + missing + skipped == disk_sessions`. **Confirmed it fails before the fix** with `assert 0 == 3` — exactly the reported symptom.
- `ruff check` and `mypy` clean for all touched files.
- `pytest tests/unit/test_transcript_sync.py tests/unit/test_backfill.py tests/integration/test_storage_backend_parity.py tests/integration/test_data_access_parity.py` — 93 passed; session/backfill integration sweep 68 passed.

## What's NOT in this PR
- No change to the direct-DB (daemon-down) path — it already counted correctly; the bug was serve-mode only.
- `fetch_ingested_session_ids` is an `ApiBackend`-only `fetch_*` outside the `StorageBackend` protocol (like the other `fetch_*` shim methods), so it needs no `test_storage_backend_parity.py` classification.


anilmurty:
Both Greptile findings addressed in 201160d.

**P1 (transcript_sync.py — false "0 ingested" on lookup failure).** The serve-mode ingested-id lookup used to swallow auth/connectivity/server/decode failures and return a partially-initialized report, so `tj backfill status` reprinted the exact `0 ingested · 0 missing · 0 skipped` + `✓ Every session ingested` that #642 set out to kill — this time on the failure path. `SyncReport` now carries a `verified` flag (default `True`) that `reconcile_claude_code` clears to `False` whenever the anti-join couldn't actually run (no `.conn` **and** no working HTTP shim). The status command now surfaces an honest `! Couldn't verify ingestion status — ... the tj daemon was unreachable` with a `tj stop` / `tj serve` remedy hint instead of a confident all-good report. This distinguishes "verified 0" from "couldn't verify".

**P2/security (api/routes/sessions.py — unbounded input).** `POST /api/v1/sessions/ingested-ids` accepted an unbounded id array (auth off by default on the local daemon). Added `MAX_INGESTED_ID_CANDIDATES = 50_000` (comfortably above any real on-disk session count, which is in the thousands) and reject oversized requests with **413** rather than silently clamping (clamping would under-report the ingested set and reintroduce a wrong count). SQL stays parameterised and chunked.

**Tests:** reconcile marks `verified=False` when the daemon fetch raises and when the backend has neither `.conn` nor a fetch method; the status command prints "Couldn't verify" and never "already ingested" / "Every session ingested" on an unverified report; the endpoint rejects an over-limit array (413) and accepts one exactly at the cap. The existing `test_reconcile_counts_ingested_sessions_when_daemon_holds_the_lock` still passes. `ruff` + `mypy` clean on touched files; `test_storage_backend_parity.py` + `test_data_access_parity.py` still green.

PR #659, opened 2026-07-29
Fix #653: cap + strip attributes from trace-detail payload (large traces never loaded)

### LINKED ISSUE #653: Lens: trace detail never loads for large traces (/traces/{id} returns full span attributes uncapped — ~1GB payload)
**Symptom:** In the Lens web UI, Observe → Traces → click a large trace → the detail/waterfall pane spins on skeleton loaders forever. Small traces open instantly.

**Root cause (measured against a live serve):** `GET /api/v1/traces/{trace_id}` (`get_trace` in `api/routes/traces.py`) returns **every span** with its **full `attributes`** dict (`_span_to_dict`, line ~190 — captured prompt/completion/tool content), with **no cap**.

| Trace | Spans | Payload | Server time |
|---|---|---|---|
| small | 1 | 1.4 KB | 0.01s |
| large | 45,384 | **951 MB** | 13.4s |

A ~1 GB JSON response can't be fetched/parsed/rendered by the browser (and a 45k-row waterfall is unrenderable anyway), so the skeleton never clears. These are exactly the big fan-out/agent traces.

**Fix (suggested):**
1. **Drop the heavy `attributes` from the waterfall payload.** The waterfall needs structure/timing/tokens/cost/name — not the full captured content of every span. That alone removes most of the ~1 GB. If the span-detail panel shows captured content when a user expands a single span, fetch it lazily (a per-span endpoint or an `?include=attributes` on a single span) rather than shipping all of it upfront.
2. **Cap the span count** returned to a sane limit (e.g. ~2,000) and surface an honest truncation notice — return `span_count` (true total) + a `truncated` flag so the UI can say "showing first N of M spans" (no silent cap).
3. **SPA hardening (`ui/index.html`):** add a fetch timeout + error state so the skeleton ALWAYS clears (never hangs), and render large traces gracefully (virtualize or cap rendered rows).

**Verification:** the 45,384-span trace should return a small payload quickly and the detail view should render (with the truncation notice) instead of hanging.

anilmurty:
Clicking a large trace in Lens → Traces spun on the skeleton loaders forever. `GET /api/v1/traces/{trace_id}` returned every span with its full captured-content `attributes` dict, uncapped — a measured 951 MB / 13.4s response for a 45,384-span trace, which the browser could neither fetch nor render.

## Summary
- Waterfall payload is now **attribute-free** (structure/timing/tokens/cost/ids only) and **capped at 2000 spans** — removes ~all of the ~1 GB.
- New lazy endpoint `GET /traces/{trace_id}/spans/{span_id}` returns one span **with** its attributes, so the span-detail-on-expand feature is preserved.
- SPA hardened: fetch timeout + error/retry state (skeleton always clears), truncation notice, rendered-row cap, lazy attributes fetch on span select.

## API (`tokenjam/api/routes/traces.py`)
- `get_trace`: waterfall payload drops `attributes`; caps at `TRACE_SPAN_CAP = 2000`; returns true `span_count`, `returned_count`, `truncated`, and `span_cap`. The 5 costliest spans are always included even when truncated so the "jump to costliest" badges resolve.
- `_span_to_dict(..., include_attributes=False)` — attributes only when explicitly requested; adds a cheap `has_attributes` boolean so the UI knows whether a lazy fetch is worth making.
- New `get_trace_span` route returns a single span with full attributes.

## SPA (`tokenjam/ui/index.html`, `TraceDetailView`)
- `Promise.race` fetch timeout (20s) + `loadState` error branch with a **Retry** button — the skeleton can never hang.
- Renders `truncated` as a visible "Showing N of M spans" notice (honest disclosure, Rule 14 — no silent cap).
- `RENDER_ROW_CAP = 1000` on rendered DOM rows so a big tree can't freeze the tab; the remainder is disclosed in the same notice.
- Selected span's captured content fetched lazily via the new per-span endpoint (was rendered straight from the waterfall payload). No new external HTTP (Rule 18).

## Tests / Verification
- `tests/integration/test_trace_detail_payload.py` — a 3000-span trace returns a capped, attribute-free, <5 MB payload with `truncated: true` and correct total `span_count`; small traces aren't truncated; the per-span endpoint returns attributes for one span and 404s for an unknown span.
- `tests/unit/test_lens_ui_regression.py` — static-grep guards: load-error/retry state present, fetch timeout present, truncation notice present, lazy per-span attributes fetch present (and the old `JSON.stringify(sel.attributes` is gone), render-row cap present.
- `test_ui_offline.py` green; `node --check` on the extracted module script passes; `ruff` + `mypy` clean on touched files; storage/data-access parity tests pass.
- **Before/after (measured):** a 5000-span trace with 8KB/span captured content went from **43.3 MB** (old) to **1.06 MB in 185ms** — 41x smaller. The real 45k-span trace shrinks far more.

## What's NOT in this PR
- No true windowed virtualization (react-window etc.) — deferred; the server cap + render cap already keep the DOM bounded, and adding a vendored virtual-list lib is out of scope for a bug fix.
- No `ApiBackend`/serve-shim change — this is a route-level fix over the existing `get_trace_spans` protocol method; the lazy per-span endpoint reuses it.

Closes #653

### DIFF
--- a/tests/integration/test_storage_backend_parity.py
+++ b/tests/integration/test_storage_backend_parity.py
@@ -31,6 +31,7 @@ See ``tokenjam/CLAUDE.md`` → "StorageBackend parity" for the one-line rule.
 from __future__ import annotations
 
 import dataclasses
+import json
 import threading
 import time
 from contextlib import contextmanager
@@ -75,6 +76,7 @@ SESSION = "parity-session"
 SHIM_PARITY_METHODS = {
     "get_traces",
     "get_trace_spans",
+    "get_span",
     "get_cost_summary",
     "get_alerts",
     "get_tool_calls",
@@ -158,7 +160,10 @@ def _proj_traces(traces) -> list:
 
 def _proj_spans(spans) -> list:
     # Includes cache_tokens / cache_write_tokens — the exact fields the shim
-    # used to drop on daemon-fetched spans (#_dict_to_span regression).
+    # used to drop on daemon-fetched spans (#_dict_to_span regression). Also
+    # includes `attributes` (#653): get_trace_spans now defaults to the FULL
+    # payload, so the shim must carry captured content — the attribute-free
+    # waterfall payload is opt-in (?attributes=false) and never taken here.
     return sorted(
         (
             s.span_id,
@@ -169,8 +174,28 @@ def _proj_spans(spans) -> list:
             round(s.cost_usd or 0.0, 6),
             s.model,
             s.provider,
+            json.dumps(s.attributes or {}, sort_keys=True),
         )
         for s in spans
+    )
+
+
+def _proj_span(span) -> tuple | None:
+    # Single-span parity (get_span) — same projection as one row of _proj_spans,
+    # including attributes so the targeted single-span fetch is proven to carry
+    # captured content across the shim.
+    if span is None:
+        return None
+    return (
+        span.span_id,
+        span.input_tokens,
+        span.output_tokens,
+        span.cache_tokens,
+        span.cache_write_tokens,
+        round(span.cost_usd or 0.0, 6),
+        span.model,
+        span.provider,
+        json.dumps(span.attributes or {}, sort_keys=True),
     )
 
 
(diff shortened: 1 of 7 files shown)


anilmurty:
Addressed all three Greptile P1s (commit 4de8a2a).

**P1-1 — Waterfall response erases exported attributes** (`traces.py:186`)
The attribute-free payload is now **opt-in**, not the default. `GET /traces/{id}` defaults to `attributes=true` (FULL spans), so `ApiBackend.get_trace_spans` and every complete-span consumer (exports, backfill) are unchanged. The Lens waterfall fetch now passes `?attributes=false` for the light payload. Response gained an `attributes_included` flag. The storage-backend parity test was tightened: seeded spans now carry captured content and `_proj_spans` includes `attributes`, so `get_trace_spans` full-attribute behavior across the shim is parity-tested (it was silently uncovered before).

**P1-2 — Lazy lookup loads entire trace** (`traces.py:210`)
Added `DuckDBBackend.get_span(trace_id, span_id)` — a parameterised `WHERE trace_id=? AND span_id=?` single-row lookup (DuckDB-only) — plus `ApiBackend.get_span` (hits the single-span endpoint). The route uses `get_span` instead of scanning + deserializing the whole trace on every expand. 404-on-unknown preserved and scoped to the trace. `get_span` is a new `StorageBackend` protocol method, classified in `test_storage_backend_parity.py` with a spec + attributes assertion.

**P1-3 — Costliest badges target hidden rows** (`index.html:5859`)
Beyond-cap costliest spans (badge targets) are now **pinned** into the rendered set, and each badge additionally gates on a `renderedRowIds` set, so a jump badge can never point at a row hidden by `RENDER_ROW_CAP`. No dead badges.

**Tests / checks**
- New: default full-attrs vs opt-in light payload; single-span targeted-fetch (spy asserts `get_trace_spans` is not called) + 404 + trace-scoping; `get_span` parity; Lens static-grep guards for the `?attributes=false` fetch and rendered-only badges.
- Existing `test_trace_detail_payload.py` adjusted for the opt-in param.
- `ruff` + `mypy` clean on touched files; `node --check` clean on the extracted module; integration (656) + both parity tests + lens-unit all pass locally.

Not merging — leaving for review.

PR #701, opened 2026-08-06
Fix #616: Refresh Review inbox relearn cache from tj optimize

### LINKED ISSUE #616: Review inbox's relearn finding is refreshed only by the daemon recompute, so tj optimize leaves it stale
**Severity:** Medium

**Where:** `core/optimize/relearn_store.py` (`write_cache` is called ONLY from `recompute_now`, which is reached from `trigger_background_recompute` / the serve API route); `cli/cmd_optimize.py` (`tj optimize` computes the relearn finding in-process and renders it, but never writes the store).

**Observed:** the Review inbox reads `relearn_proposals.list_proposals()` → `relearn_store.read_cache()`. That cache held a finding with `generated_at: 2026-07-25T22:45` and **47 clusters, every one with `past_overspend_usd: null`** — fields that had shipped since. Running `tj optimize relearn` did NOT refresh it. Running a FULL `tj optimize` did NOT refresh it either (the file's mtime changed, because the cost-proposals block lives in the same file and IS rewritten, but the `finding` key stayed a day old — a genuinely misleading signal for anyone checking freshness by mtime). Only calling `relearn_store.recompute_now()` directly refreshed it: 55 clusters, all 55 priced.

So a user who runs `tj optimize` — the obvious way to "re-analyze" — sees an inbox that does not change, with no indication the two surfaces disagree. The CLI printed 55 clusters while the inbox served 47 from the previous day.

**Why this is more than a staleness nit:** it makes the inbox structurally unable to show any newly-added cluster field until a daemon recompute happens to fire, which silently defeats field-level work landing on relearn (this was found because a new display field appeared absent from the inbox payload while being present in the analyzer's own output). It also means the CLI and the web surface can state different numbers for the same question at the same moment.

**Fix direction:** Decide the intended contract and make it legible, rather than assuming. Either (a) have `tj optimize` write the relearn store when it computes a full finding (careful: the CLI may hold a read-only DB handle while the daemon owns the write lock, which is plausibly WHY it does not today — check before assuming it is an oversight), or (b) keep the daemon as the sole writer but have the inbox display the finding's own `generated_at` and offer an explicit "rescan" that triggers `recompute_now`. The inbox already has a `⟳ Rescan now` control — verify whether it actually reaches `recompute_now`, since that would make this mostly a discoverability + staleness-disclosure fix.

**Done when:** running the documented "re-analyze" path visibly updates the Review inbox, or the inbox states how old its finding is and how to refresh it; and the CLI and inbox can no longer report different cluster counts with nothing on screen saying why.

**Ref:** Found while verifying relearn render work on #597; the stale cache initially read as a bug in the new code.

Axiya3749:
`tj optimize` computed and displayed a fresh relearn finding locally, but never persisted that finding to the Review inbox cache. This allowed the CLI and inbox to report different cluster counts until the daemon independently recomputed the cache.

## Summary

- Persist the exact relearn finding computed by direct `tj optimize` runs.
- Cover both bare `tj optimize` and `tj optimize relearn`.
- Leave the daemon-backed path read-only because it consumes an already-stored report rather than computing a new finding.
- Preserve the previous relearn result when an optimize subset does not run the relearn analyzer.

## Related issue

Closes #616

## Tests / Verification

- `env -u FORCE_COLOR .venv/bin/python -m pytest tests/integration/test_cli.py -k 'optimize and not validate' -q`
  - 19 passed
- `.venv/bin/ruff check tokenjam/cli/cmd_optimize.py tests/integration/test_cli.py`
- `.venv/bin/mypy tokenjam/cli/cmd_optimize.py`
- `git diff --check`

## What's NOT in this PR

- This does not make daemon-backed `tj optimize` perform an inline recompute. The daemon continues to own its scheduled and explicit Rescan workflow.
- No architecture or configuration changes were required.

## Checklist

- [ ] Full test suite passes
- [x] Changed files pass Ruff
- [x] Changed production file passes Mypy
- [x] CLAUDE.md update not required; architecture is unchanged
- [x] Test spans use `tests/factories.py`

@anilmurty


anshss:
@Axiya3749 thanks for pr! lgtm!

PR #705, opened 2026-08-06
Fix #614: verify daemon lifecycle and stale server state

### LINKED ISSUE #614: tj serve daemon lifecycle: launchd unit written but never loaded, stale OTLP endpoints lose telemetry silently
**Severity:** Medium

**Where:** daemon lifecycle (`tj onboard`'s launchd/systemd unit installation, `core/server_state.py`, the OTLP endpoint baked into shells by `_zshrc_otel_block`).

**Observed:** On a machine audit, `~/Library/LaunchAgents/com.tokenjam.serve.plist` EXISTS with `RunAtLoad`/`KeepAlive` true, but `launchctl list | grep -i tokenjam` returns NOTHING — it was written and never loaded. The daemon actually holding the DB lock was a manually-started foreground process; if it dies nothing restarts it. Separately, a second unrelated `tj serve --port 9341` was running from a different venv, and `~/.local/share/tj/server.state` pointed at a THIRD, dead pid/port. Any terminal whose baked-in `OTEL_EXPORTER_OTLP_ENDPOINT` targets a dead port exports into the void with no error surfaced anywhere.

**Why it still matters:** the scheduled transcript catch-up (shipped in #598) closes the data gap for anything Claude Code wrote to disk. But it only runs when the daemon runs — so an unloaded launchd unit means the self-healing itself never fires. It also does nothing for non-Claude-Code OTLP sources, which have no on-disk transcript to reconcile from.

**Fix direction:** (1) make `tj onboard` VERIFY the unit is loaded (`launchctl list` / `systemctl is-enabled`) rather than merely written, and load it if not; report the verified state in `tj doctor`. (2) Reconcile to one daemon: detect multiple live `tj serve` instances and a `server.state` pointing at a dead pid/port, and say so. (3) Probe the configured OTLP endpoint for reachability and warn when a shell's baked-in endpoint cannot be reached.

**Done when:** onboard fails loudly (or self-corrects) when the unit is not loaded; `tj doctor` reports daemon-loaded state, duplicate instances, and endpoint reachability; a test covers a written-but-unloaded unit being detected.

**Ref:** Found while working on the ingest scheduling fix in #598.

Axiya3749:
## Summary

Verifies that `tj onboard` actually registers the generated launchd/systemd unit instead of treating a written file or successful load command as proof. Extends `tj doctor` to report unloaded services, multiple live `tj serve` processes, stale or mismatched `server.state`, and the existing OTLP endpoint reachability result in one diagnostic surface.

The endpoint probe and shell-host correction had already landed on `main` in commit `c76e1b7c`; this PR keeps that implementation and adds the missing service-manager/process/state lifecycle checks from #614.

## Related issue

Closes #614

## Validation

- `pytest tests/unit/test_server_state.py tests/unit/test_onboard_daemon.py tests/unit/test_doctor_daemon_lifecycle.py tests/unit/test_doctor_otlp_endpoint.py -q` — 49 passed
- `ruff check tokenjam/` — passed
- `mypy tokenjam/` — passed
- Full local CI test scope — 5,404 passed; 11 failures and 8 setup errors were limited to sandbox-denied socket/`ps` access and Python 3.14 Rich-color behavior outside the CI Python 3.10-3.12 matrix
- GitHub CI — lint/type checking, Python 3.10, Python 3.11, Python 3.12, `test-ts`, `version-lockstep`, and contributor guard all passed

## Checklist

- [x] Tests pass (`pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/`) in GitHub CI
- [x] CLAUDE.md updated (not required; no architecture contract changed)
- [x] Test spans use `tests/factories.py` (not applicable; no spans added)
- [x] Requested `@anilmurty` as reviewer (mentioned here: @anilmurty)

Note: this PR changes tj doctor's exit code behavior: the new daemon-lifecycle checks (Daemon service, Daemon instances, Server state) can emit warnings that make doctor exit 1 in situations that previously exited 0. Anything scripting on tj doctor's exit code should account for this.

@anilmurty

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -83,6 +83,12 @@ def _isolate_real_world_side_effects(tmp_path_factory, monkeypatch):
     monkeypatch.setattr(
         "tokenjam.cli.cmd_onboard._stop_serve_for_db_write", lambda: False
     )
+    # Doctor's duplicate-instance check is intentionally machine-wide. Keep
+    # this integration file hermetic instead of letting a developer's real
+    # foreground/managed serve processes change its exit-code assertions.
+    monkeypatch.setattr(
+        "tokenjam.core.server_state.list_serve_processes", lambda: []
+    )
     # Storage paths resolve via os.path.expanduser($HOME); redirect them at the
     # default ``~/.tj/telemetry.duckdb`` so any stray open_db lands in tmp.
     monkeypatch.setenv("HOME", str(iso))
--- a/tests/unit/test_doctor_daemon_lifecycle.py
+++ b/tests/unit/test_doctor_daemon_lifecycle.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import json
+from pathlib import Path
 
 from tokenjam.cli.cmd_doctor import _check_daemon_lifecycle
 from tokenjam.core.config import ApiConfig, TjConfig
@@ -99,3 +100,36 @@ def test_doctor_reports_live_state_on_the_wrong_port(tmp_path, monkeypatch):
     assert state["level"] == "warning"
     assert "port 9341" in state["message"]
     assert "expects 7391" in state["message"]
+
+
+def test_doctor_reports_live_state_under_a_different_config(tmp_path, monkeypatch):
+    """A same-port daemon running under a DIFFERENT config file is not
+    healthy — it validates the wrong database and ingest secret, and port
+    matching alone can't see this (two configs can legitimately share a port
+    across separate installs/worktrees)."""
+    state_path = tmp_path / "server.state"
+    state_path.write_text(json.dumps({
+        "pid": 555, "port": 7391, "config_path": "/other/project/.tj/config.toml",
+    }))
+    monkeypatch.setattr(
+        "tokenjam.core.server_state.inspect_daemon_unit",
+        lambda: DaemonUnitState(None, None, False, None, None),
+    )
+    monkeypatch.setattr(
+        "tokenjam.core.server_state.list_serve_processes",
+        lambda: [ServeProcess(555, "/opt/venv/bin/tj serve")],
+    )
+    monkeypatch.setattr("tokenjam.core.server_state.is_pid_alive", lambda pid: True)
+    monkeypatch.setattr("tokenjam.core.server_state.is_serve_process", lambda pid: True)
+    _patch_state_path(monkeypatch, state_path)
+
+    config = TjConfig(
+        version="1", api=ApiConfig(port=7391),
+        config_path=Path("/this/project/.tj/config.toml"),
+    )
+    checks = _check_daemon_lifecycle(config)
+    state = next(check for check in checks if check["name"] == "Server state")
+
+    assert state["level"] == "warning"
+    assert "/other/project/.tj/config.toml" in state["message"]
+    assert "/this/project/.tj/config.toml" in state["message"]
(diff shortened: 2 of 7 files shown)


anshss:
Thanks for this. The doctor checks and the process-listing helper are genuinely useful and I want them in.

The problem is main moved under you. The launchd install path now uses `bootstrap`/`bootout`/`enable` and verifies by comparing the running unit's program args, so it already covers the "load reported success but nothing registered" case, and it also catches a resident-but-stale registration that a `launchctl list <label>` check can't see (list returns 0 for a resident job regardless of whether its args are current). The branch is conflicting against main now too, so the green checks here predate that. Can you rebase and rework the launchd piece around what's there? I'd expect most of that hunk to fall away; the systemd, doctor and `server_state` additions should survive fine.

Two things to fix while you're in there:

1. The Server state check only compares the port. `ServerState` already carries `config_path`, so a daemon running under a different config on the same port reads as healthy and doctor ends up validating the wrong database and secret. Worth one more comparison.
2. `_SYSTEMD_ENABLED_STATES` includes `linked`, `linked-runtime` and `alias`. Those don't autostart at login, so reporting them as enabled tells the user something untrue. We only ever install via `enable --now` so it can't bite today, but it's a one-line tightening.

One thing I want to flag rather than change: a new warning here flips `tj doctor`'s exit code from 0 to 1. I think that's correct, it's just a real behavior change for anything scripting on it, so worth a line in the PR body.

Also noticed the `launchctl`/`systemctl`/`ps` calls have no timeout, so a wedged service manager hangs doctor with no output. That's pre-existing and not your job here, I'm tracking it separately.

Axiya3749:
> Thanks for this. The doctor checks and the process-listing helper are genuinely useful and I want them in.
> 
> The problem is main moved under you. The launchd install path now uses `bootstrap`/`bootout`/`enable` and verifies by comparing the running unit's program args, so it already covers the "load reported success but nothing registered" case, and it also catches a resident-but-stale registration that a `launchctl list <label>` check can't see (list returns 0 for a resident job regardless of whether its args are current). The branch is conflicting against main now too, so the green checks here predate that. Can you rebase and rework the launchd piece around what's there? I'd expect most of that hunk to fall away; the systemd, doctor and `server_state` additions should survive fine.
> 
> Two things to fix while you're in there:
> 
> 1. The Server state check only compares the port. `ServerState` already carries `config_path`, so a daemon running under a different config on the same port reads as healthy and doctor ends up validating the wrong database and secret. Worth one more comparison.
> 2. `_SYSTEMD_ENABLED_STATES` includes `linked`, `linked-runtime` and `alias`. Those don't autostart at login, so reporting them as enabled tells the user something untrue. We only ever install via `enable --now` so it can't bite today, but it's a one-line tightening.
> 
> One thing I want to flag rather than change: a new warning here flips `tj doctor`'s exit code from 0 to 1. I think that's correct, it's just a real behavior change for anything scripting on it, so worth a line in the PR body.
> 
> Also noticed the `launchctl`/`systemctl`/`ps` calls have no timeout, so a wedged service manager hangs doctor with no output. That's pre-existing and not your job here, I'm tracking it separately.

Axiya3749:
Rebased onto main and dropped the _install_launchd verification hunk entirely. Your bootstrap/bootout/enable rework already covers what it checked, and catches the resident-but-stale case mine couldn't. Systemd, doctor, and server_state additions carried over clean.

Fixed both: Server state now compares config_path too (a same-port daemon under a different config warns instead of reading healthy), and _SYSTEMD_ENABLED_STATES drops linked/linked-runtime/alias. Added a regression test for each.

Added a note to the PR body about the exit-code change per your flag.

PR #711, opened 2026-08-08
fix(storage): cap the per-span system prefix and bound DuckDB's buffer pool

anshss:
Found while diagnosing a Mac sitting at 4–6 GB of RAM with 11.4 GB of swap in use. The `tj` daemon was a large part of it and `~/.tj/telemetry.duckdb` had reached 3.7 GB. Two independent causes.

## 1. The whole `CLAUDE.md` was stored on every span

`_read_project_claude_md` captured the project's entire `CLAUDE.md` onto each assistant LLM span. The value is written **per span** but is **identical across every span of a project**, so its real cost is size × span count:

| | |
|---|---|
| Spans carrying it | 92,514 |
| Avg bytes each | 43,832 |
| Total on disk | **4.06 GB** |
| Distinct values | 61 |
| Unique text | **1.84 MB** |
| Amplification | **2207×** |

All of it had to page through DuckDB's buffer pool on every scan of `spans`.

### The fix: store a fingerprint, not the file

`cache_recommend` is the only consumer, and it never read the prefix as text. It uses it exactly three ways:

| Use | Needs |
|---|---|
| `_prefix_hash(text)` | an identity |
| display | `text[:120]` |
| skip gate | `len(text)` |

All three are derivable at capture time. The span now carries those three (~230 bytes) instead of the file (~43,000) — the same real database would hold **~18 MB** where it once held **4.06 GB**.

An intermediate commit here truncated the text to 2000 chars. That cut the worst of it but left the shape of the defect intact: measured on the real database, the truncated form is still **155 MB across 76,962 spans to carry 61 KB of distinct text** — 32 distinct values, a 2505× amplification. The problem was never the size of the value, it was storing a per-project constant per span. Both commits are kept so the reasoning stays on the record.

`core.system_prefix` owns the derivation because the producer and the consumer previously agreed only by coincidence — the parser stored whatever it read off disk while the analyzer hashed `text[:PREFIX_HASH_BYTES]`. Both now route through one function. Divergence there has no symptom: it would silently split one project's spans across two candidates and understate both occurrence counts.

### Existing history keeps working

The legacy `SYSTEM_PREFIX_CONTENT` branch stays as a **read** path. Every span backfilled before this change carries the text, and dropping the fallback would make that history invisible to the analyzer the day the release shipped. `test_legacy_and_compact_spans_land_in_one_candidate` pins that old and new spans describing the same prefix group together — verified by deleting the fallback, which fails it plus one pre-existing persona test.

Capture-off strips all four keys, not just the text one: `SAMPLE` is literal prompt content, and leaving `HASH`/`LENGTH` behind would keep a fingerprint of a file the user just said not to capture.

## 2. DuckDB was given no memory budget

`duckdb.connect()` was called with no config. Left unset, DuckDB sizes `memory_limit` at **80% of physical RAM** — 19.1 GiB on a 24 GB Mac — and its buffer manager has no reason to evict below that, so the daemon grows until the machine swaps.

Both connect sites now route through `_connect_bounded`. That detail is load-bearing: applying the bound only in `__init__` leaves it **silently absent** after `_reopen` recovers from a fatal error, and a recovered daemon running unbounded has no symptom until the machine is already swapping.

The ceiling is **2GB**, measured rather than picked. At 1GB the daemon returned 500s:

```
_duckdb.OutOfMemoryException: failed to allocate data of size 16.0 KiB (1.0 GiB/953.6 MiB used)
  File "tokenjam/core/db.py", line 3828, in get_daily_cost
```

That aggregate runs fine under 1GB in isolation (0.1s on the 768k-span database). What broke it is that `start_catch_up` opens its **own** backend beside the request path, so the pool was already full. Verified separately that same-path connections share one database instance, so the ceiling is 2GB for the process rather than 2GB per backend.

`temp_directory` makes the ceiling a spill point rather than a failure, and an override DuckDB rejects degrades to the default instead of leaving the database unopenable.

## Result

Daemon RSS: **880 MB idle (spiking multi-GB) → 111 MB, settling to 25 MB** once startup catch-up finishes. Endpoints verified serving 200s throughout.

## Tests

`tests/unit` + `tests/synthetic`: **4813 passed**. `ruff check tokenjam/` and `mypy tokenjam/` clean.

Notable additions, each pinning something with no other symptom:

- `test_legacy_and_compact_spans_land_in_one_candidate` — the upgrade must not split a project's history in two.
- `test_hash_window_is_single_sourced` — producer and consumer share one window constant.
- `test_stored_hash_matches_what_the_analyzer_computes_from_text` — identity survives the change of storage.
- `test_two_prefixes_differing_after_the_window_are_one_candidate` — states what the window *means*, so changing it is deliberate.
- `test_capture_off_strips_every_system_prefix_key` — no fingerprint left behind.
- `test_backfill_stores_a_fingerprint_not_the_file` — the defect as a property: a big `CLAUDE.md` must not make a big span.

The two pre-existing backfill tests that asserted the old attribute are **updated, not deleted** — what they guard (a stable per-project signal that is not the per-turn prompt; the lazy-load retry after a record with no `cwd`) is unchanged.

## Two notes

**Existing databases.** This stops the growth; it does not shrink a database that already has the blobs. DuckDB never returns free blocks to the OS, so reclaiming needs the stored values reduced in place and then a `COPY FROM DATABASE` rewrite into a fresh file. Doing that took a real 3.94 GB database to 0.75 GB with all 767,128 spans and 7,167 sessions preserved.

**Pre-existing lint.** The `lint` job also runs `mypy tokenjam/`, which reports an unrelated pre-existing error in the `ingest_secret` patcher (`config.py`, `m.group(1) + f'"{secret}"'`) on some mypy versions. Left alone rather than folded into a storage PR.

### DIFF
--- a/tests/unit/test_backfill.py
+++ b/tests/unit/test_backfill.py
@@ -922,6 +922,12 @@ def test_capture_off_extracts_no_content_only_provenance(tmp_path):
         GenAIAttributes.PROMPT_CONTENT, GenAIAttributes.COMPLETION_CONTENT,
         GenAIAttributes.TOOL_INPUT, GenAIAttributes.TOOL_OUTPUT,
         TjAttributes.SYSTEM_PREFIX_CONTENT,
+        # The compact prefix keys ride the same capture toggle: SAMPLE is
+        # literal prompt text, and HASH/LENGTH are a fingerprint of a file the
+        # user just said not to capture.
+        TjAttributes.SYSTEM_PREFIX_HASH,
+        TjAttributes.SYSTEM_PREFIX_SAMPLE,
+        TjAttributes.SYSTEM_PREFIX_LENGTH,
     }
     for span in (llm, tool):
         assert span.attributes["source"] == "backfill.claude_code"
@@ -999,9 +1005,15 @@ def test_capture_prompts_reads_project_claude_md_as_system_prefix(tmp_path):
     """#272: the human's per-turn message never repeats verbatim, so
     cache-recommend's prefix-hash needs a different, genuinely stable
     signal. The project's CLAUDE.md is read straight off disk (it's not in
-    the transcript) and stamped as `TjAttributes.SYSTEM_PREFIX_CONTENT` --
+    the transcript) and stamped as `TjAttributes.SYSTEM_PREFIX_HASH` --
     identical on every assistant span for the same project, unlike
-    PROMPT_CONTENT."""
+    PROMPT_CONTENT.
+
+    The prefix is stored as a fingerprint rather than as the text: it was the
+    same value on every span, so keeping it whole cost (size x span count).
+    What this asserts is unchanged -- that a stable per-project signal exists
+    and that it is NOT the per-turn prompt."""
+    from tokenjam.core.system_prefix import prefix_hash
     from tokenjam.otel.semconv import TjAttributes
 
     project_dir = tmp_path / "proj"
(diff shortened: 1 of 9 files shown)


anshss:
Pushed a third commit typing the `config=` dict — duckdb types that parameter as an invariant dict over a value union, so the inferred `dict[str, str]` was rejected even though every value is a `str`.

**One pre-existing lint failure is not from this PR.** The `lint` job also runs `mypy tokenjam/`, which is already red on `main`:

```
$ git checkout origin/main -- . && mypy tokenjam/core/config.py
tokenjam/core/config.py:623: error: Unsupported operand types for + ("bytes" and "str")  [operator]
tokenjam/core/config.py:623: note: Left operand is of type "bytes | Any"
Found 1 error in 1 file
```

That's the `ingest_secret` patcher (`m.group(1) + f'"{secret}"'` — mypy can't narrow the match to `str`), unrelated to storage. It appears at line 640 on this branch only because these changes shift it down 17 lines. I've left it alone rather than fold an unrelated fix into a storage PR, but flagging it since it will keep the lint job red until someone takes it.

PR #714, opened 2026-08-09
Fix UTC day bucketing and a lying cast in the API read path

anshss:
## Summary

Two timestamp-correctness fixes, both outside the ingest path:

1. **UTC day bucketing.** Three sites used a bare `CAST(start_time AS DATE)` on the `TIMESTAMPTZ` column: `get_cost_summary`'s `day` grouping (and its default fallback for an unrecognised `group_by`), and the optimize report's `active_days` count. A bare cast resolves through the session's local timezone before truncating, so on a machine running ahead of UTC, spans logged late in the UTC day get stamped with tomorrow's date. That means `tj cost --group-by day` silently disagrees with any UTC-based figure elsewhere, and `active_days` (which every per-day extrapolation in the optimize report divides by) is off by one near local midnight. Fixed by switching to `CAST(start_time AT TIME ZONE 'UTC' AS DATE)`, matching the convention already used elsewhere in the codebase (`core/data_span.py`, `api/routes/analytics.py`, `api/routes/cost.py`).

2. **A `cast()` in the HTTP-API fallback backend that lied to the type checker.** `ApiBackend._dict_to_span` (used when `tj serve` holds the DuckDB write lock) built a `NormalizedSpan` with `start_time=cast(datetime, ... or None)`. `NormalizedSpan.start_time` (and `TraceRecord.start_time`, which had the identical pattern in `get_traces`) is declared non-Optional by design — the header comment on the field says ingest already rejects any span with no observed time rather than substituting a default. The `cast()` didn't convert anything; it just told the type checker to stop looking while a `None` could flow through, leaving every downstream consumer of an HTTP-fallback span unguarded.

   **Decision:** a `start_time` is required here, not optional. A span the daemon serializes always has a real `start_time` (ingest enforces it), so a missing/unparseable value in the wire payload means the response itself is malformed, not that the field is legitimately absent. I raise a `ValueError` naming the offending record instead of casting a `None` past the checker, and factored the shared logic into `_require_start_time()` since `get_traces` had the exact same bug for `TraceRecord`.

## Tests

- `tests/integration/test_db.py::test_get_cost_summary_by_day_buckets_on_the_utc_date` — parametrized over UTC/Asia-Kolkata/America-Los_Angeles session timezones.
- `tests/unit/test_optimize.py::test_summarize_window_active_days_buckets_on_the_utc_date` — same parametrization, two spans straddling UTC midnight.
- `tests/unit/test_api_backend_span_deserialize.py` — new file covering the present/missing/absent-key `start_time` contract for `_dict_to_span` and `_require_start_time`.

I confirmed both new day-bucketing tests fail against the pre-fix bare `CAST(... AS DATE)` (reverted each site locally, ran red, restored, ran green) before finalizing.

## CI

`ruff check` and `mypy` pass on all touched files. Ran the touched test files plus the broader db/optimize/mcp/proxy/storage-parity suites locally (183 passed); did not run the full suite (that's what remote CI is for).

## Deliberately not done

- Did not widen the sweep beyond the three named sites plus the one `get_traces` cast — a full-repo grep for `CAST(... AS DATE)` on timestamp columns turned up no other bare occurrences; everything else already used the `AT TIME ZONE 'UTC'` form.

### DIFF
--- a/tests/integration/test_db.py
+++ b/tests/integration/test_db.py
@@ -1,7 +1,7 @@
 """Integration tests for the database layer."""
 from __future__ import annotations
 
-from datetime import timedelta
+from datetime import datetime, timedelta, timezone
 
 import pytest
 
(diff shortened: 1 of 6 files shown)


anshss:
Bundled — this lands via https://github.com/Metabuilder-Labs/tokenjam/pull/715, which now carries every commit from this branch. Leaving this PR open on purpose: once the anchor merges (as a **merge commit**, not a squash), this one flips to Merged automatically because its head SHA stays an ancestor of `main`.

Please don't push to this branch in the meantime. A new commit here changes the head SHA, breaks that ancestry, and strands this PR open with a stale diff.

Issue #249, opened 2026-06-23
Subscription '% of cycle' framing is misapplied at per-item granularity (traces, status cards)


anilmurty:
For subscription users, dollar figures are suppressed and rendered as **"% of cycle"**. That's sensible for **window-level aggregate** spend (the Cost/Overview totals), but it's being applied at **per-item granularity** where it's confusing or alarming:

- **Traces list COST column:** a single LLM call shows "0.2% of cycle" — implying one call eats 0.2% of your whole month. Weird denominator at the per-row level.
- **Status cards "Cost today":** showed **"466.7% of cycle"** for one agent — a >100% per-item figure reads as broken.

## Fix
At **per-item / per-row** granularity, render subscription cost as **tokens** — exactly what the Cost table already does with its "COST (TOKENS)" column — not "% of cycle". Optionally show implied-API-value (clearly framed as "would-be API cost", not "spent") on hover. Reserve **"% of cycle"** for **window-level aggregate** spend only.

This is one consistent rule: % of cycle = aggregate spend vs the cycle budget; per-row = tokens. Apply it to the Traces COST column, the Status "Cost today" cards, and audit any other per-item dollar surface for the same misuse.

## Why it matters
Honesty discipline + first-run clarity: a subscription user (the common case) currently sees confusing/alarming per-item "% of cycle" numbers across multiple screens. Most of the dashboard was clearly built/tested against API-plan ($-visible) data, so the suppression path has these granularity bugs.

## Constraints
Consume the framing block; never re-derive suppression in JS. UI-only. Static-grep regression + keep `test_ui_offline.py` green.

PR #717, opened 2026-08-12
docs: fix comments citing unrelated issue #17 for cache-write-token work

### LINKED ISSUE #594: Several source comments cite issue #17 for unrelated cache-write-token work
## Problem

Several source comments in this repo reference issue #17 with a bare `#17`, but issue/PR #17 in this repository is "Fix pricing file not included in pip wheel" — unrelated to what these comments are actually describing (cache-write token visibility). A reader who clicks through from one of these comments lands on a closed pip-packaging PR that doesn't explain the code they're looking at:

- `tokenjam/core/db.py:1898` — "...cache-write is often the dominant cost driver yet was invisible above the DB — issue #17)."
- `tokenjam/cli/cmd_cost.py:108` — "...cache-write is often the dominant driver and was previously invisible (#17)."
- `tokenjam/core/models.py:345` — `cache_write_tokens: int = 0   # cache-CREATE tokens (the hidden cost driver, #17)`
- `tokenjam/api/routes/traces.py:125` — `"cache_write_tokens": span.cache_write_tokens,  # cache-CREATE tokens (#17)`

This repo's own `CLAUDE.md` already documents the general hazard class in Critical Rule 23: a bare `#N` reference is ambiguous, and the fix depends on whether the number actually documents the thing it sits next to — qualify it with a full URL when it's a genuine reference to a different issue/PR, or drop the number and describe the thing plainly when it doesn't match. These four sites are the "doesn't match" case.

## Expected

Each of the four comments above describes the cache-write-tokens visibility work in plain language, with no `#N` reference, since #17 doesn't document that work. A repo-wide grep for `#[0-9]+` inside source comments should also be run once to check for any other reference whose target issue/PR doesn't match what the comment describes. Most `#N` comments in this codebase (there are many) do point at the right issue or PR — for example the cache-write field's other neighboring comments correctly cite #209 and #249, and the prompt/completion-capture comments in `tokenjam/sdk/integrations/openai.py` correctly cite #320 — so this is not a blanket "remove every reference," only a fix for the ones that are actually wrong.

## Where to look

- `tokenjam/core/db.py:1898`
- `tokenjam/cli/cmd_cost.py:108`
- `tokenjam/core/models.py:345`
- `tokenjam/api/routes/traces.py:125`
- `CLAUDE.md` Critical Rule 23 — documents this hazard class and the qualify-vs-describe decision rule to apply when auditing other references

## Done when

- The four sites above no longer cite #17.
- A repo-wide grep for bare `#[0-9]+` in source comments turns up no other site whose reference doesn't match the actual GitHub issue or PR it names.

laishettikarthik-tech:
Fixes source comments that cite issue #17 for work #17 doesn't describe.

## Summary

Four source comments referenced `#17` ("Fix pricing file not included in
pip wheel") to explain cache-write-token visibility work, which #17 is
unrelated to. Per Critical Rule 23 in `CLAUDE.md`, a bare `#N` reference
that doesn't match what it sits beside should describe the thing plainly
rather than cite the wrong number.

## Fixed sites

- `tokenjam/core/db.py` — `get_cost_summary` comment
- `tokenjam/cli/cmd_cost.py` — CACHE R / CACHE W columns comment
- `tokenjam/core/models.py` — `CostRow.cache_write_tokens` field comment
- `tokenjam/api/routes/traces.py` — `_span_to_dict`'s `cache_write_tokens` key comment

Each now describes the cache-write-tokens visibility fact in plain
language, with the `#17` reference removed. No other reference near
these sites was touched (e.g. `#209`, `#249` remain — confirmed correct
against their actual issue content).

## Tests / Verification

- `python -m py_compile` on all four modified files — clean.
- Repo-wide `grep -rn '#17[^0-9]' tokenjam/` after the fix — zero remaining
  `#17` references anywhere in the codebase.

## What's NOT in this PR

The issue's second "Done when" criterion asks for a repo-wide grep to
check for *any* other reference whose target doesn't match its comment —
not just `#17`. I ran the grep (`grep -rn '#[0-9]\+' --include='*.py'
tokenjam/`) and it surfaced several hundred `#N` references across the
codebase. Verifying each of those against its actual issue content is a
much larger audit than this issue's four named sites, and requires
issue-tracker access I don't have as an external contributor. I've
confirmed no other `#17` citation exists; a broader audit is better
suited to a maintainer with tracker access, or a dedicated follow-up
issue if warranted.

Closes #594

### DIFF
--- a/tokenjam/api/routes/traces.py
+++ b/tokenjam/api/routes/traces.py
@@ -294,7 +294,7 @@ def _span_to_dict(span: object, include_attributes: bool = False) -> dict:
         "input_tokens": span.input_tokens,
         "output_tokens": span.output_tokens,
         "cache_tokens": span.cache_tokens,            # cache-READ tokens
-        "cache_write_tokens": span.cache_write_tokens,  # cache-CREATE tokens (#17)
+        "cache_write_tokens": span.cache_write_tokens,  # cache-CREATE tokens
         "cost_usd": span.cost_usd,
         "request_type": span.request_type,
         "conversation_id": span.conversation_id,
--- a/tokenjam/cli/cmd_cost.py
+++ b/tokenjam/cli/cmd_cost.py
@@ -92,7 +92,8 @@ def cmd_cost(ctx: click.Context, agent: str | None, since: str,
         return _cost_cell(value)
 
     # CACHE R / CACHE W columns make cost reconcilable from the shown tokens —
-    # cache-write is often the dominant driver and was previously invisible (#17).
+    # cache-write is often the dominant driver and was previously invisible
+    # before this column existed.
     cache_r = sum(r.cache_tokens for r in rows)
     cache_w = sum(r.cache_write_tokens for r in rows)
     if group_by == "day":
--- a/tokenjam/core/db.py
+++ b/tokenjam/core/db.py
@@ -3671,11 +3671,11 @@ class DuckDBBackend:
 
         # Cache-read + cache-write are summed alongside in/out so callers can
         # show the full token picture (cache-write is often the dominant cost
-        # driver yet was invisible above the DB — issue #17). call_count is
-        # the only genuinely honest metric for the "tool" grouping: tool-call
-        # spans carry no cost/tokens of their own — cost is attributed to the
-        # LLM completion span the tool call accompanied, not the tool
-        # invocation itself.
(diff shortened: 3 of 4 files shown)


anilmurty (approved):

anilmurty:
Thanks for the contribution @laishettikarthik-tech and welcome!

PR #718, opened 2026-08-12
chore: split CLAUDE.md into path-scoped rules (52% less always-resident context)

anilmurty:
`CLAUDE.md` had grown to ~180 KB / ~45k tokens, every byte of it always-resident — re-sent at the head of every session in every worktree, whether or not the work touches the subsystem being described. Two thirds of it was reference material for one package at a time.

This is the same cost our own `resend` and `downsize` analyzers price on users' files, and it runs against the guidance the file itself records: length and redundancy reduce adherence, and specific must not mean longer.

## Summary

- Move per-subsystem detail into 8 path-scoped rules under `.claude/rules/`, each carrying a `paths:` glob so it loads only when Claude reads a file it covers
- Always-resident context: **~45k → ~21.4k tokens (52% reduction)**
- **Nothing is reworded.** Every content line in every rule file is verbatim from the `CLAUDE.md` this PR replaces
- Narrow `.gitignore` exception so the rules are actually tracked, plus a CI guard for the invariant that exception inverts

## What moved where

| Rule file | Loads on | On-demand |
|---|---|---|
| `optimize.md` | `core/optimize/`, `fixes/`, `rulewrite/`, `summarize/` | 39.6k |
| `cli.md` | `tokenjam/cli/` | 15.5k |
| `core-domain.md` | `core/*.py`, `ingest_adapters/`, `export/` | 13.1k |
| `mcp-daemon.md` | `mcp/`, onboard/stop/uninstall commands | 7.3k |
| `web-ui.md` | `tokenjam/ui/` | 7.2k |
| `api-otel.md` | `api/`, `otel/` | 6.9k |
| `pricing.md` | `pricing/`, `core/pricing.py`, `core/cost.py` | 6.8k |
| `sdk.md` | `tokenjam/sdk/`, `examples/`, `incidents/` | 3.1k |

726 characters of frontmatter across all eight is the entire standing cost of the ~99.5k now delivered on demand.

## Critical Rules are untouched

All 41 stay in `CLAUDE.md`, complete and numbered as before. They are cross-cutting, and the list opens by requiring citation as "Critical Rule N" — relocating any would force a stub or a renumber, and compressing them means rewriting prose whose exact wording is load-bearing. That is a separate decision, deliberately not taken here. Critical Rules are still ~31% of the remaining file if we want to revisit it.

## The `.gitignore` exception, and why it needs a guard

`.claude/` was ignored wholesale, so rules placed there would have been local-only and never reached contributors or CI. Only `.claude/rules/*.md` is now trackable; everything else under `.claude/` stays ignored, including non-markdown dropped into `rules/` itself.

That inverts an invariant this repo has relied on for its whole life — "under `.claude/`" meaning "private" — for exactly one subdirectory, while the material it protects is markdown too. A note dropped into `rules/` by muscle memory would publish. So it is pinned in CI rather than left to memory: `tests/unit/test_claude_rules_are_the_only_public_dotclaude.py` asserts nothing but `rules/*.md` is tracked, and that every tracked rule really is path-scoped (a rule that loses its `paths:` line silently becomes always-resident again — the file is written, the guidance works, only the cost is wrong).

## Tests / Verification

- Verified in both directions, mechanically: **no line lost** (every >40-char line of the old file survives somewhere) and **no line added** (every content line in the rule files is verbatim from `CLAUDE.md` at HEAD; the only new text is frontmatter plus a 4-line header per file)
- All 8 rules confirmed `path-scoped` by our own `core/summarize/load_semantics.classify` — not by eyeballing the format
- New guard test passes alongside `test_no_tracked_dev_secrets.py`
- `git add -An .claude/` stages exactly the 8 rule files; `settings.local.json`, the strategy/competitive docs, and a non-`.md` file inside `rules/` all still resolve as ignored

Caveat worth a reviewer's eye: I verified the path-scoping assertion fails correctly on an unscoped rule, but did **not** negative-test the "only rules are tracked" assertion, since that meant staging a decoy under `.claude/`. Please read that assertion rather than trusting it.

## What's NOT in this PR

- **Critical Rules compression** — see above; needs a call on the citation convention
- **`sync.sh` exclusion** — the nightly `.claude/` mirror will now also archive these 8 public files into the private repo. Harmless duplication, but it lives outside this repo so it is not changed here
- **No code, config, or behaviour changes** — markdown, `.gitignore`, and one test


anilmurty:
Closing — this is already done on `main`, and better. `origin/main` carries 18 path-scoped rules under `.claude/rules/` and a 33.5 KB CLAUDE.md (~8.4k tokens); this branch was cut from a stale local `main` at 58ca780 and would have reverted that work, including the Critical Rules compression it deliberately left undone. No action needed.

PR #725, opened 2026-08-15
Fix #582: include closed sessions in drift baselines

### LINKED ISSUE #582: Drift baselines silently exclude 'closed' sessions: get_completed_sessions filters on status='completed' only
## Problem

`DuckDBBackend.get_completed_sessions` and `DuckDBBackend.get_completed_session_count`
(`tokenjam/core/db.py`, around lines 2031-2051) filter sessions with hand-written SQL that only
matches `status = 'completed'`:

```python
def get_completed_sessions(self, agent_id: str, limit: int) -> list[SessionRecord]:
    cur = self.conn.execute(
        "SELECT * FROM sessions WHERE agent_id = $1 AND status = 'completed' "
        "ORDER BY COALESCE(ended_at, started_at) DESC LIMIT $2",
        [agent_id, limit],
    )
    ...

def get_completed_session_count(self, agent_id: str) -> int:
    result = self.conn.execute(
        "SELECT COUNT(*) FROM sessions WHERE agent_id = $1 AND status = 'completed'",
        [agent_id],
    ).fetchone()
    ...
```

But sessions have two distinct terminal statuses: `'completed'` (a session wrapped by a real
session span — how backfilled sessions land) and `'closed'` (a session explicitly ended via the
close endpoint). Both are defined together as terminal in `tokenjam/core/models.py`:

```python
# Terminal session statuses: a session that will never produce another span.
# 'closed' = explicitly ended (POST /sessions/close); 'completed' = wrapped by a
# real session span (how backfilled Claude Code sessions land). status_at()
# passes both through untouched, and both belong in the archive rather than the
# live tiles. Any query that partitions sessions by lifecycle must use this
# constant instead of spelling one literal into hand-written SQL.
TERMINAL_STATUSES = ("closed", "completed")
```

`get_completed_sessions`/`get_completed_session_count` are consumed directly by
`DriftDetector.on_session_end` (`tokenjam/core/drift.py`, lines 239-241) to decide whether enough
history exists to build a drift baseline, and to fetch the sessions that go into it:

```python
count = self.db.get_completed_session_count(agent_id)
if count >= agent_config.drift.baseline_sessions:
    sessions = self.db.get_completed_sessions(
        agent_id, limit=agent_config.drift.baseline_sessions,
    )
    baseline = build_baseline(agent_id, sessions, self.db)
```

Because the query only recognizes `'completed'`, any session that ended via the explicit close
path never enters a drift baseline, and an agent whose sessions are mostly `'closed'` may never
accumulate enough counted sessions to reach `baseline_sessions` in the first place — even though
real terminal history exists.

There's already a working precedent for querying both statuses correctly:
`tokenjam/api/routes/status.py` builds its archive query off the shared constant instead of a bare
literal:

```python
params: list = list(TERMINAL_STATUSES)
terminal_placeholders = ", ".join(f"${i}" for i in range(1, len(params) + 1))
...
clause = f"status IN ({terminal_placeholders}) OR ..."
```

## Where to look

- `tokenjam/core/db.py` — `get_completed_sessions` and `get_completed_session_count` (around lines
  2031-2051).
- `tokenjam/core/models.py:20` — the `TERMINAL_STATUSES` constant to consume instead of the bare
  `'completed'` literal.
- `tokenjam/core/drift.py:239-241` — the consumer (`DriftDetector.on_session_end`) whose baseline
  inputs shift once this is fixed.
- `tokenjam/api/routes/status.py` (around line 139) — an existing example of parameterizing a
  DuckDB `IN` clause off `TERMINAL_STATUSES`.

## Suggested fix

Change both `DuckDBBackend` methods to filter on `status IN (...)` over `TERMINAL_STATUSES`
instead of the bare `'completed'` literal, following the `status.py` pattern (build the
placeholder list from `len(TERMINAL_STATUSES)`, keeping the query parameterized — no f-string SQL
per this repo's own rules).

Because this changes the *inputs* to drift detection (baselines, and any thresholds tuned against
them, will shift once `'closed'` sessions are included), the PR should also report the measured
baseline shift on a real or synthetic corpus, and add a test asserting that both a `'closed'`
session and a `'completed'` session are counted and returned.

## Done when

- `get_completed_sessions` and `get_completed_session_count` in `tokenjam/core/db.py` both use
  `TERMINAL_STATUSES` (or an equivalent `status IN (...)` clause covering both terminal statuses)
  instead of a bare `'completed'` literal.
- A test pins that a session with `status='closed'` and one with `status='completed'` both count
  toward and appear in a drift baseline for the same agent.
- The PR includes a synthetic fixture with a known mix of `closed` and `completed` sessions and
  states the measured before/after baseline shift on that fixture, since drift verdicts may move
  as a result and outside contributors have no access to a real corpus.

sanmaxdev:
Explicitly closed sessions were valid terminal history, but the DuckDB drift queries selected only sessions marked `completed`. Agents using the close endpoint could therefore miss valid history and fail to form a drift baseline.

## Summary
- use the shared `TERMINAL_STATUSES` set in both DuckDB drift-history queries
- keep every status value parameterized with generated placeholders
- add an end-to-end regression covering retrieval, counting, and baseline construction

## Related issue
Closes #582

## Behavior verified
The synthetic fixture contains one `closed` session and one `completed` session for the same agent.

- before: 1 session counted, so the 2-session baseline threshold was not reached
- after: 2 sessions counted and returned, producing a baseline with `sessions_sampled == 2`
- measured eligibility shift: 1 to 2 sessions, a 100% increase for this fixture
- measured baseline input average: 300 tokens for the pre-fix completed-only population versus 200 tokens after including both terminal statuses, a 33.3% decrease

## Tests / Verification
- new regression test fails against the pre-fix implementation with `assert 1 == 2`
- `pytest tests/synthetic/test_drift_detection.py::test_terminal_sessions_feed_drift_baseline -v` passed
- relevant drift and storage checks passed: 59 tests
- `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/ -q` passed: 5,586 passed, 18 skipped, 2 xfailed
- `ruff check tokenjam/` clean
- `mypy tokenjam/` clean
- `git diff --check` clean

## What's NOT in this PR
- no schema, API, CLI, or drift-threshold changes
- no unrelated cleanup of pre-existing test-file lint findings

## Checklist
- [x] Contributor docs updated (not needed, no architecture change)

@anilmurty ready for review.

### DIFF
--- a/tests/synthetic/test_drift_detection.py
+++ b/tests/synthetic/test_drift_detection.py
@@ -87,6 +87,32 @@ def test_baseline_built_after_n_sessions(db):
     assert baseline.avg_input_tokens == 1000.0
 
 
+def test_terminal_sessions_feed_drift_baseline(db):
+    _setup_agent(db)
+    closed = make_session(agent_id="test-agent", status="closed", input_tokens=100)
+    completed = make_session(agent_id="test-agent", status="completed", input_tokens=300)
+    db.upsert_session(closed)
+    db.upsert_session(completed)
+
+    assert db.get_completed_session_count("test-agent") == 2
+    sessions = db.get_completed_sessions("test-agent", limit=2)
+    assert {session.session_id for session in sessions} == {
+        closed.session_id,
+        completed.session_id,
+    }
+
+    detector = DriftDetector(
+        db,
+        MagicMock(),
+        _make_config(baseline_sessions=2),
+    )
+    detector.on_session_end("test-agent", completed)
+
+    baseline = db.get_baseline("test-agent")
+    assert baseline is not None
+    assert baseline.sessions_sampled == 2
+
+
 def test_drift_inactive_before_baseline_complete(db):
     _setup_agent(db)
     config = _make_config(baseline_sessions=10)
--- a/tokenjam/core/db.py
+++ b/tokenjam/core/db.py
@@ -37,6 +37,7 @@ from tokenjam.core.models import (
     SessionRecord,
     SpanKind,
     SpanStatus,
+    TERMINAL_STATUSES,
     TraceCostStats,
     TraceFilters,
     TraceRecord,
@@ -3821,19 +3822,24 @@ class DuckDBBackend:
         # still active afterwards — otherwise the status tile shows a 40s blip
         # instead of the real multi-hour session. Falls back to started_at when
         # ended_at is NULL.
+        terminal_placeholders = ", ".join("?" for _ in TERMINAL_STATUSES)
         cur = self.conn.execute(
-            "SELECT * FROM sessions WHERE agent_id = $1 AND status = 'completed' "
-            "ORDER BY COALESCE(ended_at, started_at) DESC LIMIT $2",
-            [agent_id, limit],
+            "SELECT * FROM sessions WHERE agent_id = ? AND status IN ("
+            + terminal_placeholders
+            + ") ORDER BY COALESCE(ended_at, started_at) DESC LIMIT ?",
+            [agent_id, *TERMINAL_STATUSES, limit],
         )
         rows = cur.fetchall()
         cols = [d[0] for d in cur.description]
         return [_row_to_session(r, cols) for r in rows]
 
     def get_completed_session_count(self, agent_id: str) -> int:
+        terminal_placeholders = ", ".join("?" for _ in TERMINAL_STATUSES)
         result = self.conn.execute(
-            "SELECT COUNT(*) FROM sessions WHERE agent_id = $1 AND status = 'completed'",
-            [agent_id],
+            "SELECT COUNT(*) FROM sessions WHERE agent_id = ? AND status IN ("
+            + terminal_placeholders
+            + ")",
+            [agent_id, *TERMINAL_STATUSES],
         ).fetchone()
         return result[0] if result else 0
 


anilmurty (approved):

anilmurty:
Merged — thanks @sanmaxdev. Three things I want to call out specifically, because they're the parts that made this quick to review:

**You consumed the existing `TERMINAL_STATUSES` instead of minting a second status set.** That constant's docstring exists precisely to prevent a second copy, and this is its third consumer. You also stayed stricter than the precedent the issue pointed you at — plain concatenation for the placeholder list where `status.py` uses an f-string, which keeps Critical Rule 7 unambiguous at a glance.

**The measured shift in the PR body is reproducible to the digit.** I re-ran `build_baseline` against your fixture and got the count shift and the 300→200 average exactly as stated, and reverting only `db.py` reproduces the `assert 1 == 2` you quoted. A regression test whose load-bearing-ness is claimed *and* true is worth more than it sounds — this repo has shipped decorative guards before.

**It also fixes a second bug you didn't claim.** Those two methods have six other callers. `tj status` was falling back to the newest `completed` session, so an agent whose most recent terminal session was `closed` displayed a stale session as its latest — while the Lens archive, already on `TERMINAL_STATUSES`, showed the correct one. Those two surfaces now agree.

One optional follow-up, not for this PR: both methods are still named `get_completed_*` while now returning all terminal sessions. Renaming touches the `StorageBackend` protocol, `ApiBackend`, the MCP shim and the parity classification, so it was right to leave it out of scope — but the stale docstring at `core/drift.py:218` could ride along with it if you ever pick it up.

PR #726, opened 2026-08-17
fix(cli): skip DB probe when nested subcommand requests --help

### LINKED ISSUE #580: `tj <group> <subcommand> --help` fails with a database-lock error while `tj serve` is running
## Problem

Running `--help` on a subcommand of a group command fails with a database-lock error if `tj serve`
is running, instead of printing help. For example:

```
tj optimize downsize --help
tj backfill langfuse --help
tj loop annotate --help
```

all fail with:

```
Error: Database is locked (tj serve is running?)
```

instead of showing the subcommand's help text.

## Root cause

The root Click group's callback (`cli()` in `tokenjam/cli/main.py`) opens/probes the database for
any invoked subcommand that isn't in the `no_db_commands` allowlist, and it does this *before*
Click gets a chance to evaluate the child command's own `--help` flag:

```python
no_db_commands = {
    "stop", "uninstall", "onboard", "mcp", "demo", "policy",
    "proxy", "summarize", "pricing", "otel-resource-attrs", "session-end",
    "statusline",
    "ping",
}
invoked = ctx.invoked_subcommand
...
db = open_db(config.storage)   # this runs even for `tj optimize downsize --help`
```

(`tokenjam/cli/main.py`, `cli()`, roughly lines 63-88.)

`ctx.invoked_subcommand` is set to the top-level group name (e.g. `optimize`), not the leaf
subcommand, so the callback can't distinguish "the user wants to run `downsize`" from "the user
wants help for `downsize`" — it opens the DB unconditionally for every group not in the allowlist.

This affects any multi-level group: `optimize`, `route`, `backfill`, `loop` are examples currently
registered as groups in `tokenjam/cli/main.py` (see the `cli.add_command(...)` calls near the
bottom of the file) and not in `no_db_commands`. Top-level `tj <group> --help` (no subcommand) is
unaffected since Click short-circuits before invoking the callback's DB logic in that case;
`policy`, `proxy`, `pricing`, and `summarize` are unaffected because they're already in the
allowlist.

## Why it matters

Reading `--help` is the most basic thing a CLI user does, and it should never require a database
connection. This fails precisely when the user is most likely to be running the product day-to-day
(the daemon holding the DB lock is the normal, recommended operating mode).

## Suggested fix

Short-circuit the DB probe when the invocation is a help request, rather than growing
`no_db_commands` command-by-command (which just leaves the next new group to rediscover this bug).
Detect `--help` (or `-h` if aliased) in `sys.argv` / `ctx.args` within the parent callback and
skip `open_db(...)` when present.

## Where to look

- `tokenjam/cli/main.py` — the `cli()` group callback, particularly the `no_db_commands` block and
  the `open_db(...)` call that follows it.
- `tests/integration/test_cli.py` — existing pattern for CLI tests using `click.testing.CliRunner`
  with `unittest.mock.patch` on `tokenjam.cli.main.open_db`, useful for asserting the DB is never
  opened for a `--help` invocation.

## Done when

- `tj <group> <subcommand> --help` prints help for every affected group (e.g. `optimize`, `route`,
  `backfill`, `loop`) even while something else holds the database write lock (simulate by holding
  a `tj serve` process, or by patching `open_db` to raise the lock error in a test).
- A regression test covers at least one previously-failing case, asserting `open_db` is not called
  when `--help` is requested on a group subcommand.

syf2211:
## Summary

Skip the DuckDB probe in the root CLI callback when the invocation requests `--help`, so nested commands like `tj optimize downsize --help` work while `tj serve` holds the write lock.

## Motivation

Fixes #580. Running `--help` on a group subcommand failed with "Database is locked" whenever the daemon was running, because `open_db()` ran before Click rendered help.

## Changes

- Detect help tokens in unconsumed `ctx.args` during `TjCommand.invoke` and set `ctx.obj["_skip_db_for_help"]`.
- Skip `open_db()` in the root `cli()` callback when that flag is set (same path as `no_db_commands`).
- Add parametrized integration tests for `optimize downsize`, `backfill langfuse`, and `loop annotate`.

## Tests

```bash
pytest tests/integration/test_cli.py::test_nested_subcommand_help_skips_db_probe -v
ruff check tokenjam/cli/main.py tokenjam/cli/tj_status.py tests/integration/test_cli.py
mypy tokenjam/cli/main.py tokenjam/cli/tj_status.py
```

All passed.

## Notes

- composer-2.5 review: APPROVE
- Normal invocations without `--help` still probe the DB as before.
- Also fixes top-level `tj <cmd> --help` under the same lock condition.

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -2753,3 +2753,18 @@ def test_validate_daemon_lock_fails_cleanly(runner, config):
                          ["optimize", "--validate", "downsize"])
     assert result.exit_code == 1
     assert "tj serve" in result.output
+
+
+@pytest.mark.parametrize("args", [
+    ["optimize", "downsize", "--help"],
+    ["backfill", "langfuse", "--help"],
+    ["loop", "annotate", "--help"],
+])
+def test_nested_subcommand_help_skips_db_probe(runner, args):
+    """Nested group subcommand --help must not open DuckDB (#580)."""
+    with patch("tokenjam.cli.main.load_config", return_value=TjConfig(version="1")), \
+         patch("tokenjam.cli.main.open_db",
+               side_effect=AssertionError("must not open db for --help")):
+        result = runner.invoke(cli, args)
+    assert result.exit_code == 0, result.output
+    assert "Usage:" in result.output
--- a/tokenjam/cli/main.py
+++ b/tokenjam/cli/main.py
@@ -124,7 +124,7 @@ def cli(ctx: click.Context, config_path: str | None, output_json: bool,
     }
     invoked = ctx.invoked_subcommand
 
-    if invoked in no_db_commands:
+    if invoked in no_db_commands or ctx.obj.get("_skip_db_for_help"):
         ctx.obj["config"] = config
         ctx.obj["config_path_override"] = config_path
         ctx.obj["db"] = None
--- a/tokenjam/cli/tj_status.py
+++ b/tokenjam/cli/tj_status.py
@@ -105,7 +105,21 @@ class TjCommand(click.Command):
         self.no_status = no_status
         super().__init__(*args, **kwargs)
 
+    def _tokens_request_help(self, ctx: click.Context) -> bool:
+        names: set[str] = set()
+        node: click.Context | None = ctx
+        while node is not None:
+            names.update(node.help_option_names)
+            node = node.parent
+        tokens = list(ctx.args)
+        return any(tok in names for tok in tokens)
+
     def invoke(self, ctx: click.Context) -> Any:
+        ctx.ensure_object(dict)
+        if self._tokens_request_help(ctx):
+            # Root/nested group callbacks probe DuckDB before Click renders
+            # ``--help`` on a leaf subcommand (#580).
+            ctx.obj["_skip_db_for_help"] = True
         if not self.status_message:
             return super().invoke(ctx)
         with tj_status(self.status_message, ctx):


anilmurty (approved):
Approving. I reproduced the original failure against a genuinely held DuckDB write lock (second process on an r/w connection, `[api] port` pointed at a dead port so the API fallback couldn't mask it) and confirmed the fix across the tree.

**It fixes more than the issue claimed.** #580 says top-level `tj <group> --help` is unaffected — it wasn't. `tj cost --help` and `tj optimize --help` were broken too. So this covers `--help` for every command not already in `no_db_commands`, at any depth. The `no_db_commands` set and the eager `tj --help` / `--version` paths are unchanged, as expected.

The seam is right: `cli` is a `TjGroup`, which resolves `invoke` to `TjCommand.invoke`, so the hook runs before the root callback fires and before the sub-context is built — the only point where the flag can be set in time. There's no separate `TjGroup.invoke` needing the same patch.

**The detail worth calling out:** reading `ctx.help_option_names` off the context chain rather than hardcoding `"--help"`. `-h` isn't wired in this app today, and this picks it up for free the day someone adds `context_settings`. That's the kind of choice that stops a fix rotting.

Tests are load-bearing — reverting only the two source hunks fails all three parametrized cases — and they use the `open_db(side_effect=AssertionError)` pattern CLAUDE.md prescribes.

Two nits, non-blocking:

1. **The scan can trip on a token Click will consume as an option value.** `tj loop annotate SID --note --help` sets the skip flag, runs with `db=None`, and dies with `AttributeError: 'NoneType' object has no attribute 'execute'`. It needs the token to be *exactly* `--help`, and I checked that no command in the tree uses `ignore_unknown_options`/`allow_extra_args`, so there's no pass-through command legitimately forwarding a `--help` to a subprocess — which was my main worry and it's clean. Nonsense input, nothing written on the crash path. Filing the general form as a follow-up rather than holding this.
2. Setting the flag unconditionally (`ctx.obj["_skip_db_for_help"] = self._tokens_request_help(ctx)`) rather than only on `True` would make it self-clearing if a long-lived process ever reused one `obj` dict. Safe, since the leaf's `invoke` runs after the root callback has read it.

Test gaps worth a line each if you're already in here: nothing asserts the **inverse** — that a non-help invocation still opens the DB, which is the direction that would actually break the product — and nothing pins the newly-fixed single-level shape (`tj cost --help`), which is the broader win.

Issue #728, opened 2026-08-17
CLI: --help detection should skip tokens Click consumes as an option value


anilmurty:
Follow-up to #726, which is otherwise correct and merged on its own merits.

#726 skips the DuckDB probe when the invocation asks for `--help`, by testing membership of the help tokens in the root's unconsumed `ctx.args`. That is the right shape, but a plain membership test also matches a `--help` that Click will go on to consume as an **option value**:

```
tj loop annotate SID --note --help
→ AttributeError: 'NoneType' object has no attribute 'execute'
```

The user did not ask for help; the command runs with `db=None` and crashes.

## Why this is narrow

- The token must be **exactly** `--help` — a note *containing* the word is fine
- No command in the tree uses `ignore_unknown_options` or `allow_extra_args`, so there is no pass-through command that legitimately forwards a `--help` to a subprocess
- Most similar invocations die earlier on option validation (`tj cost --group-by --help` hits Choice validation; `tj cost --since --help` hits the window parser)
- Nothing is written on the crash path

So this is nonsense input rather than a plausible mistake, which is why it did not block #726.

## Done when

- Help detection ignores a token positioned as an option value, rather than matching anywhere in `ctx.args`
- A regression test covers `tj loop annotate SID --note --help` running normally (or failing on its own terms) rather than skipping the DB probe
- A test asserts the **inverse** direction too: an ordinary invocation still opens the DB. #726's tests cover only the skip direction, and the inverse is the one that would actually break the product

## Related

#727 makes the failure mode itself less bad — a clean `ClickException` instead of an `AttributeError` — but does not fix the false positive. The two are independent.

PR #729, opened 2026-08-18
fix(cli): fail cleanly when a DB-requiring command has no backend

### LINKED ISSUE #727: CLI: a command that needs a DB should fail cleanly when the backend is None
Surfaced while reviewing #726.

When a command runs with `ctx.obj["db"] = None`, it currently dereferences the `None` and raises a raw traceback rather than telling the user what went wrong.

Reproduction (needs #726 or the `no_db_commands` path):

```
tj loop annotate SID --note --help
→ AttributeError: 'NoneType' object has no attribute 'execute'   (core/loop.py:160)
```

This is pre-existing and not caused by #726 — `no_db_commands` has the same shape. Any command that skips `open_db()` and then reaches DB-touching code produces a traceback instead of a message.

`cmd_optimize` already handles this by hand:

> `optimize requires either a direct DuckDB connection or a running tj serve at the configured api.{host,port}.`

That is the right behaviour; it just isn't shared. `loop annotate` and any other DB-touching command outside `no_db_commands` have nothing equivalent.

## Done when

- A single shared guard (a decorator, or a check in `TjCommand.invoke`) raises a `ClickException` naming the command when a DB-requiring command finds `db is None`
- `cmd_optimize`'s hand-rolled version is replaced by it, so there is one copy of the message rather than two
- A test asserts the clean error rather than the traceback, for at least one command that is not `optimize`

Nothing is written on the current crash path, so this is a diagnostics problem rather than a data-integrity one.

syf2211:
## Summary

Add a shared guard in `TjCommand.invoke` that raises a `ClickException` when a DB-requiring leaf command runs with `ctx.obj["db"] is None`, instead of crashing with an `AttributeError` traceback.

## Motivation

Issue #727: commands that skip `open_db()` (including the `--help` skip path from #726) but still execute can reach DB-touching code with `db=None`. `cmd_optimize` already had a hand-rolled message; other commands like `loop annotate` did not.

## Changes

- Add `db_required_message()` and `_ensure_db_available()` to `TjCommand` (leaf commands only; groups skip the guard)
- Set `ctx.obj["requires_db"]` in `cli/main.py` (`False` for `no_db_commands`, `True` otherwise)
- Remove the redundant `db is None` check from `cmd_optimize`; reuse the shared message for the ApiBackend path
- Add unit + integration regression tests

## Tests

```bash
pytest tests/unit/test_tj_db_guard.py tests/integration/test_cli.py::test_db_requiring_command_fails_cleanly_when_db_is_none tests/integration/test_cli.py::test_nested_subcommand_help_skips_db_probe -v
ruff check tokenjam/cli/tj_status.py tokenjam/cli/main.py tokenjam/cli/cmd_optimize.py
mypy tokenjam/cli/tj_status.py tokenjam/cli/main.py
```

All passed locally.

## Notes

- Builds on the nested `--help` DB skip from #726 (`fix/group-subcommand-help-no-db`). If #726 merges first, this branch can be rebased to a single commit on top of `main`.
- Related follow-up: #728 (ignore `--help` tokens consumed as option values) is separate.

Fixes #727

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -2768,3 +2768,17 @@ def test_nested_subcommand_help_skips_db_probe(runner, args):
         result = runner.invoke(cli, args)
     assert result.exit_code == 0, result.output
     assert "Usage:" in result.output
+
+
+def test_db_requiring_command_fails_cleanly_when_db_is_none(runner, config):
+    """A DB-requiring leaf with db=None raises ClickException, not a traceback (#727)."""
+    with patch("tokenjam.cli.main.load_config", return_value=config), \
+         patch("tokenjam.cli.main.open_db",
+               side_effect=AssertionError("must not open db")):
+        result = runner.invoke(
+            cli,
+            ["loop", "annotate", "session-1", "--note", "--help"],
+        )
+    assert result.exit_code == 1, result.output
+    assert "annotate requires either a direct DuckDB connection" in result.output
+    assert "AttributeError" not in result.output
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -2780,5 +2780,5 @@ def test_db_requiring_command_fails_cleanly_when_db_is_none(runner, config):
             ["loop", "annotate", "session-1", "--note", "--help"],
         )
     assert result.exit_code == 1, result.output
-    assert "annotate requires either a direct DuckDB connection" in result.output
+    assert "`--help` was consumed as an option value" in result.output
     assert "AttributeError" not in result.output
--- a/tests/unit/test_tj_db_guard.py
+++ b/tests/unit/test_tj_db_guard.py
@@ -14,16 +14,18 @@ from tokenjam.cli.tj_status import (
 )
 
 
-def test_db_required_message_names_command() -> None:
-    msg = db_required_message("annotate")
-    assert "annotate requires either a direct DuckDB connection" in msg
+def test_db_required_message_names_command_path() -> None:
+    msg = db_required_message("tj loop annotate")
+    assert "tj loop annotate requires either a direct DuckDB connection" in msg
     assert "api.{host,port}" in msg
 
 
-def test_help_consumed_as_option_value_message_names_command() -> None:
-    msg = help_consumed_as_option_value_message("annotate")
-    assert "annotate cannot run" in msg
+def test_help_consumed_as_option_value_message_names_command_path() -> None:
+    msg = help_consumed_as_option_value_message("tj loop annotate")
+    assert "tj loop annotate cannot run" in msg
     assert "`--help` was consumed as an option value" in msg
+    assert "Give that option a value" in msg
+    assert "Put `--help` before other flags" not in msg
 
 
 def test_tj_command_guard_raises_click_exception_not_traceback() -> None:
@@ -50,6 +52,17 @@ def test_tj_command_guard_help_skip_message() -> None:
         demo_cmd.invoke(ctx)
     assert "`--help` was consumed as an option value" in str(exc.value)
     assert "direct DuckDB connection" not in str(exc.value)
+
+
+def test_tj_command_guard_fail_open_without_requires_db_key() -> None:
+    """``requires_db`` absent from ``ctx.obj`` must not fail-closed (#729)."""
+    @click.command("demo-cmd", cls=TjCommand)
+    def demo_cmd() -> None:
+        return "ok"
+
+    ctx = click.Context(demo_cmd)
+    ctx.obj = {"db": None}
+    assert demo_cmd.invoke(ctx) == "ok"
 
 
 def test_tj_command_guard_allows_live_backend() -> None:
--- a/tests/unit/test_tj_db_guard.py
+++ b/tests/unit/test_tj_db_guard.py
@@ -1,0 +1,33 @@
+"""Regression tests for the shared DB-required guard (#727)."""
+from __future__ import annotations
+
+import click
+import pytest
+
+from tokenjam.cli.tj_status import TjCommand, TjGroup, db_required_message
+
+
+def test_db_required_message_names_command() -> None:
+    msg = db_required_message("annotate")
+    assert "annotate requires either a direct DuckDB connection" in msg
+    assert "api.{host,port}" in msg
+
+
+def test_tj_command_guard_raises_click_exception_not_traceback() -> None:
+    @click.command("demo-cmd", cls=TjCommand)
+    def demo_cmd() -> None:
+        raise AssertionError("must not reach command body")
+
+    ctx = click.Context(demo_cmd)
+    ctx.obj = {"db": None, "requires_db": True}
+    with pytest.raises(click.ClickException) as exc:
+        demo_cmd.invoke(ctx)
+    assert "demo-cmd requires either a direct DuckDB connection" in str(exc.value)
+    assert "AssertionError" not in str(exc.value)
+
+
+def test_tj_group_skips_db_guard() -> None:
+    group = TjGroup("demo-group")
+    ctx = click.Context(group)
+    ctx.obj = {"db": None, "requires_db": True}
+    group._ensure_db_available(ctx)
--- a/tests/unit/test_tj_db_guard.py
+++ b/tests/unit/test_tj_db_guard.py
@@ -1,16 +1,29 @@
 """Regression tests for the shared DB-required guard (#727)."""
 from __future__ import annotations
 
+from unittest.mock import MagicMock
+
 import click
 import pytest
 
-from tokenjam.cli.tj_status import TjCommand, TjGroup, db_required_message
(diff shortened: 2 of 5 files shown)


anilmurty (changes requested):
Thanks @syf2211 — the shape is right and all three of #727's criteria are met. Two things before this can go in, one of which is failing CI.

## Blocking: `_requires_db` defaults fail-closed, and it breaks 143 tests

```python
return bool((ctx.obj or {}).get("requires_db", True))   # tj_status.py:132
```

`ctx.obj["requires_db"]` is only ever set by `main.py`'s root callback. Any invocation that builds its own `obj` — which is how a large share of this repo's CLI tests work (`CliRunner().invoke(cmd_onboard, ..., obj={"config": ...})`) — has no such key, so the guard reads `True`, finds no `"db"`, and hard-fails a command that never needed a database.

`test (3.10)`, `test (3.11)` and `test (3.12)` are all red on this PR: **143 failed, 5298 passed**, every failure carrying the new message verbatim, across 21 files (`test_onboard_*` ×55, `test_relearn_cli` ×18, `test_cost_proposal_verbs` ×15, `test_cli_uninstall` ×10, and others).

Flipping the default to `False` gives **5441 passed, 17 skipped, 2 xfailed**, and both of your new tests still pass — the unit test sets `requires_db=True` explicitly, and the integration test enters through `cli()`, which always sets the key. Nothing is lost by failing open here: the key is present on every real CLI path, so the default only governs contexts the root callback never built.

## The message is wrong wherever it currently fires

This is the more interesting one, and it isn't your bug so much as something the guard surfaced.

`ctx.obj["db"] = None` is assigned in exactly one place (`main.py:130`), reachable only via `no_db_commands` (where the guard is skipped) or #726's `_skip_db_for_help`. On the lock path, `open_db` raises its own specific `ClickException` and never leaves `db` as `None`. So the only state in which this guard can actually fire is *"#726 declined to open the DB because it thought you asked for help."*

Which makes the message untrue exactly where it is raised:

> `annotate requires either a direct DuckDB connection or a running tj serve at the configured api.{host,port}.`

A user who follows that and starts `tj serve` gets the identical error, because the help short-circuit happens before `open_db` either way. The traceback is genuinely gone — I confirmed the raw `AttributeError` at `core/loop.py:160` on the reverted source — but it's been replaced with a confident wrong diagnosis and an action that cannot help. For `optimize` the same sentence *is* accurate.

#727 asked for one copy of the message; it turns out these are two different causes wearing one sentence. Suggest branching: keep `db_required_message()` for the genuine no-backend case, and when `ctx.obj.get("_skip_db_for_help")` is what produced the `None`, say so instead — something like *"`--help` was consumed as a value for `--note`; put `--help` first, or drop the option."* That would also close #728 rather than leaving it to re-open this surface.

## What's good

The census came out clean: all 79 leaf commands are `TjCommand` (`TjGroup.command_class` propagates it), so there is no command that touches the DB and escapes the guard. And testing `is not None` rather than `isinstance` is the right call — I ran `tj serve` holding the write lock and confirmed `tj optimize` and `tj cost` still resolve through `probe_api` to an `ApiBackend` and render normally. An `isinstance(db, DuckDBBackend)` check would have looked equally correct and silently broken every daemon-up invocation.

## Merge ordering

I've just squash-merged #726, so this branch will now show as needing an update — its copy of #726's commits no longer shares ancestry with main's squashed commit. `git rebase --onto main 07350b0` should do it; expect conflicts in `main.py` and `tj_status.py`, since your hunks sit adjacent to #726's.

## Smaller nits

- `cmd_optimize.py:189` — the surviving `if config is None:` branch still says *"optimize requires a database connection."*, which described the `db` half you just removed. It should say the config failed to load.
- `from tokenjam.cli.tj_status import db_required_message` is imported inside the function body at `cmd_optimize.py:265`, while `tj_status` is already a module-level import elsewhere in the CLI. No cycle risk — hoist it.
- `test_tj_group_skips_db_guard` asserts nothing (it's a "does not raise" test). Fine, but worth a comment saying so.
- Neither test pins the *unguarded* direction — that a command with a live `ApiBackend` still passes. That's the closest regression class to this change.
- A DB-needing subcommand added later under an existing `no_db_commands` group (`summarize`, `rules`, `policy`, `pricing`, `proxy`) would inherit `requires_db=False` and go unguarded. Your new per-command `requires_db` kwarg fixes that, but nothing uses it yet — worth a one-line comment on `no_db_commands` pointing at it.

syf2211:
Thanks @anilmurty — addressed in 7d2f295:

- Flipped `_requires_db` default to `False` (fail-open for test invocations without `ctx.obj` from `main.py`)
- Added explicit error when `--help` is consumed as an option value (#728)
- Expanded unit/integration tests for both paths

The remaining CI failures on Python 3.12 (32 tests in `test_quickstart.py` / `TestCombinationBillingHoistedUpFront`) look unrelated to this PR — they fail with `TypeError: _fake_progress() got an unexpected keyword argument 'console'`, which appears to be a test-fixture/main drift issue rather than the DB guard change. Python 3.11 passed cleanly on the same run.

Happy to rebase if needed once those upstream failures are resolved.

anilmurty (commented):
Thanks @syf2211 — the design here is sound and I confirmed it works end to end. Three things to fix before merge, none of them large, plus one note for the PR body.

**First, your read on CI is right.** I merged current `main` into the branch and ran the full suite: 5596 passed, 12 skipped, 2 xfailed. `tests/unit/test_quickstart.py` — the file failing on the 3.10/3.12 jobs with `_fake_progress() got an unexpected keyword argument 'console'` — passes both on `main` alone and on the merged branch, so those failures are the stale base rather than anything in this change. Please merge `main` in so CI reruns clean.

**1. The advice in the new message doesn't actually work.** `help_consumed_as_option_value_message` says "Put `--help` before other flags, or drop the option that took it." The first half never helps, because Click parses the whole argument list before eager options fire:

- `tj cost --help --since` → exit 2, `Option '--since' requires an argument.`
- `tj loop annotate --help sess-1 --note` → exit 2, `Option '--note' requires an argument.`
- `tj cost --since 7d --help` → renders help correctly.

So the two remedies that work are giving the swallowing option a value, or dropping it. Critical Rule 44 is about exactly this: a remedy the product prints has to work when the user follows it. Something like "Give that option a value (`tj cost --since 7d --help`) or drop it" would be accurate.

**2. The fail-open default isn't pinned by a test.** `_requires_db()` returning `False` when `requires_db` is absent is the behavior you changed in the last round, and I mutated it back to `True` with the guard tests plus the whole CLI suite still green. A single test — a context with no `requires_db` key and `db=None`, asserting no raise — locks the decision in so it can't be silently reverted.

**3. Both messages name the leaf rather than the invocation path.** A user who typed `tj loop annotate` reads "annotate requires either a direct DuckDB connection…", and `tj backfill claude-code` reads "claude-code cannot run…". `ctx.command_path` gives the full `tj loop annotate` (I checked), which is both unambiguous and copy-pasteable.

**One note for the PR body rather than the code.** #727 asked for one copy of the message, and there are about fifteen other hand-rolled `db is None` ClickExceptions with their own wording — `cmd_backfill` (six), `cmd_report` (two), `cmd_context`, `cmd_route`, `cmd_tokenmaxx`, `cmd_quota_audit`, `cmd_session_story`, and `data_access.py`. Your shared guard now fires before those on any DB-requiring leaf, so they're shadowed there and stay live only on the `no_db_commands` paths. I'd keep them out of this PR, but please say so under "What's NOT in this PR" so the next reader knows it was a decision rather than an oversight, and I'll file the follow-up.

Everything else checks out, and I verified rather than assumed:

- The guard is genuinely load-bearing: removing the `_requires_db` gate fails 23 tests, removing the group skip fails 3.
- Live behavior is correct — a clean `Error:` at exit 1 with no traceback, while `tj loop annotate --help`, `tj optimize --help` and `tj loop --help` all still render. The help-swallow path also catches `tj cost --since --help` and `tj backfill claude-code --since --help`, which is nice coverage beyond the one command the issue named.
- `ruff check tokenjam/` is clean, and the only two `ctx.invoke` cross-command sites are both no-DB, so the guard can't shadow an internal invocation.

Two small nits, take or leave: the `isinstance(self, click.Group)` check inside `_ensure_db_available` would read better as a `TjGroup` override, and `False if invoked in no_db_commands else True` is just `invoked not in no_db_commands`.

syf2211:
Thanks @anilmurty — addressed all three items in 72085f7:

1. **Help-skip advice** — replaced the non-working "put `--help` first" wording with remedies that actually work: give the swallowing option a value or drop it (with `tj cost --since 7d --help` as the example).

2. **Fail-open default** — added `test_tj_command_guard_fail_open_without_requires_db_key` so `requires_db` absent from `ctx.obj` cannot silently revert to fail-closed.

3. **Invocation path** — guard errors now use `ctx.command_path` (e.g. `tj loop annotate`) instead of the leaf name alone.

All 7 `test_tj_db_guard` tests pass locally. Branch is rebased on current `main`.

PR #730, opened 2026-08-18
fix(optimize): rank findings in core instead of importing the CLI

### LINKED ISSUE #579: API layer imports a CLI-private function, inverting the documented core/cli/api layering
## Problem

`tokenjam/api/routes/optimize.py` imports a private function from the CLI package:

```python
from tokenjam.cli.cmd_optimize import _rank_findings
```

(`tokenjam/api/routes/optimize.py:22`, used at line 131). `_rank_findings` is defined in
`tokenjam/cli/cmd_optimize.py:512` and is clearly CLI-owned: it's prefixed with an underscore
(private) and its module is the `tj optimize` CLI command implementation.

This inverts the layering rule stated in this repo's own `CLAUDE.md`:

> `tokenjam/core/` is pure domain logic. Must never import from `tokenjam.cli` or
> `tokenjam.api`. CLI and API import from core, not the reverse.

While the rule as written targets `core/`, the spirit is the same one level up: `api/` reaching
into `cli/` internals creates a dependency in the wrong direction. The API now depends on a
private helper that the CLI package is free to rename, inline, or delete without any signal that
it will break the API route.

## Expected

`api/routes/optimize.py` should not import anything from `tokenjam.cli`. Both the CLI and the API
should call a shared ranking helper that lives in `tokenjam/core/optimize/`.

## Where to look

- `tokenjam/api/routes/optimize.py:22` — the offending import, and line 131 where it's called.
- `tokenjam/cli/cmd_optimize.py:512` — `_rank_findings` definition (also called internally at
  line 833).
- `tokenjam/core/optimize/` — the natural home for the relocated function (e.g. `runner.py` or a
  new `ranking.py` alongside it).

## Suggested fix

Move `_rank_findings` into `tokenjam/core/optimize/` (dropping the leading underscore since it
becomes a shared public helper), update `tokenjam/cli/cmd_optimize.py` to import it from there,
and update `tokenjam/api/routes/optimize.py` to import it from `tokenjam.core.optimize` instead
of `tokenjam.cli.cmd_optimize`.

## Done when

- `tokenjam/api/routes/optimize.py` has no import from `tokenjam.cli`.
- The ranking logic lives in `tokenjam/core/optimize/` and both the CLI and API call the same
  shared function.
- Existing tests covering `tj optimize` output ordering and the `/api/v1/optimize` route still
  pass.

myukitty:
`GET /optimize` imported `_rank_findings` from `tokenjam.cli.cmd_optimize`, which inverts the core/cli/api layering rule.

Ranking now lives in `tokenjam.core.optimize.rank`. The API imports it from core. The CLI keeps thin wrappers so existing renderer tests stay put. `CARD_FINDING_NAMES` is the card-bearing set, not the full analyzer registry, and a drift test pins it to `_FINDING_RENDERERS`.

Closes #579

### DIFF
--- a/tests/unit/test_optimize_rank.py
+++ b/tests/unit/test_optimize_rank.py
@@ -1,0 +1,109 @@
+"""Core ranking used by both the CLI text view and GET /optimize."""
+from __future__ import annotations
+
+from datetime import datetime, timezone
+from pathlib import Path
+from types import SimpleNamespace
+
+from tokenjam.core.optimize.rank import (
+    ALWAYS_FULL_FINDINGS,
+    CARD_FINDING_NAMES,
+    rank_findings,
+    reclaimable_share,
+)
+from tokenjam.core.optimize.types import OptimizeReport, WindowSummary
+
+UTC = timezone.utc
+
+
+def _report(**kwargs) -> OptimizeReport:
+    summary = WindowSummary(
+        since=datetime(2026, 1, 1, tzinfo=UTC),
+        until=datetime(2026, 1, 8, tzinfo=UTC),
+        days=7,
+        sessions=3,
+        spans=10,
+        total_tokens=1000,
+        total_cost_usd=1.0,
+        thin_data=False,
+    )
+    return OptimizeReport(window=summary, **kwargs)
+
+
+def test_reclaimable_share_none_without_estimate():
+    assert reclaimable_share(SimpleNamespace(), 1000) is None
+    assert reclaimable_share(SimpleNamespace(past_overspend_tokens=None), 1000) is None
+    assert reclaimable_share(SimpleNamespace(past_overspend_tokens=50), 0) is None
+
+
+def test_reclaimable_share_clamps_negative():
+    assert reclaimable_share(SimpleNamespace(past_overspend_tokens=-10), 1000) == 0.0
+
+
+def test_rank_orders_by_share_then_name_order():
+    report = _report(
+        downgrade=SimpleNamespace(past_overspend_tokens=100),
+        findings={
+            "resend": SimpleNamespace(past_overspend_tokens=400),
+            "cache": SimpleNamespace(past_overspend_tokens=400),
+            "trim": SimpleNamespace(past_overspend_tokens=50),
+        },
+    )
+    ranked = rank_findings(report, None)
+    names = [name for name, _ in ranked]
+    # cache is listed before resend in CARD_FINDING_NAMES; equal share keeps that order
+    assert names[:3] == ["cache", "resend", "downsize"]
+    assert names[3] == "trim"
+
+
+def test_rank_skips_downsize_when_not_requested():
+    report = _report(
+        downgrade=SimpleNamespace(past_overspend_tokens=999),
+        findings={"cache": SimpleNamespace(past_overspend_tokens=10)},
+    )
+    ranked = rank_findings(report, ["cache"])
+    assert [name for name, _ in ranked] == ["cache"]
+
+
+def test_rank_includes_empty_downsize_when_requested():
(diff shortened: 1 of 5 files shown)


anilmurty (approved):
Approving — and welcome, @myukitty, this is a strong first contribution. CI is now green across all seven checks (I approved the workflow run; fork PRs from first-time contributors need that manually).

## Verification

The premise checked out on fresh `main` — one offending import, exactly where #579 said:

```
tokenjam/api/routes/optimize.py:34: from tokenjam.cli.cmd_optimize import _rank_findings
```

After this PR, a full-package grep for module-level *and* function-local imports returns zero hits from `core/` or `api/` into `cli/`. All three of #579's "Done when" criteria are met.

Behaviour-preservation was checked rather than assumed: `main`'s `_rank_findings` was run against both new paths over 30,000 randomized reports, plus an exhaustive sweep of insertion-order permutations at identical shares to exercise tie-breaking specifically — zero mismatches. The three-key sort tuple is intact, and `CARD_FINDING_NAMES` is byte-identical to `_FINDING_RENDERERS`'s key order. `tj optimize --since 30d` against a real database still ranks biggest-first, and a persona-gated subset still emits its "Not run" line.

All 233 added lines account for: 109 test file + 100 `rank.py` + 10 re-exports + 11 CLI wrapper + 3 API import swap. Nothing in the diff is neither a move nor a test.

## What you did well — specifically

**You noticed the card-bearing name set is not the analyzer registry.** Reaching for `ANALYZER_REGISTRY` or `ANALYZER_ORDER` is the obvious move when relocating ranking into `core/` — it's in the same package and it looks canonical. It would have been wrong: `budget-projection` is registered but has no ranked card, and `placement` has a card but is deliberately unregistered (it's produced inside `downsize`). That's the single easiest way this refactor could have quietly broken `tj optimize placement`, and you walked past it and said so in the PR body.

**You pinned the duplication you introduced, in the same PR.** Having created a second declaration of the name ordering, you wrote the drift test binding it to `_FINDING_RENDERERS` — and bound `_ALWAYS_FULL_FINDINGS` by *identity* rather than equality, so a re-declared literal fails instead of passing on equal contents.

**You preserved `relearn`'s unranked rule and tested it.** That's a subtle, hard-won behaviour that a refactor could very easily have dropped on the floor.

**Scope discipline.** You kept two CLI re-exports with an honest `noqa` comment rather than rewriting six unrelated test modules. That was the right call.

## Nits — none blocking, all fine as follow-ups

1. `CARD_FINDING_NAMES` being a second declaration of the renderer key order is safe because your drift test catches desync — but the stronger shape is for `core` to own the names and `cmd_optimize` to build its renderer table keyed off them, so drift becomes impossible rather than merely detected.
2. `known_names` / `always_full` are parameters that provably can only hold their defaults (your own drift test proves it). Dropping them would remove the injection seam and about 8 lines. Defensible as-is.
3. `test_api_optimize_route_does_not_import_cli` guards one file by reading its source text. Two upgrades worth considering: resolve the path via `inspect.getsourcefile(...)` so it survives a non-editable install, and generalize it to walk every module under `core/` and `api/`. There's no repo-wide guard on `main` today, so that would close the whole class rather than the one instance #579 found. Good fast-follow if you want another.
4. `_reclaimable_share`'s original docstring explained *why* `None` isn't `0.0`, with the concrete example (cache-recommend recommends a placement, not a token count) and the consequence (it would hide an analyzer's own empty-state message). `rank.py` condenses that to "they are not de-minimis". That sentence is precisely what stops a future reader "simplifying" the `None` into a `0.0` — worth carrying across.

One housekeeping note: I just merged #726, so this will need `Update branch` before it goes in.

anilmurty:
Great first contribution @myukitty - thanks!

PR #732, opened 2026-08-22
fix(alerts): hydrate cooldown and failure-rate dedup state from the alerts table on startup

### LINKED ISSUE #592: Alert dedup/cooldown state is in-memory only, so a daemon restart re-fires duplicate alerts
## Problem

`AlertEngine` keeps its dedup and cooldown state entirely in process memory:

- `CooldownTracker` (`tokenjam/core/alerts.py:109-129`) suppresses repeat alerts of the same type for the same agent within a cooldown window, using an in-memory `_last_fired: dict[tuple[str, str], datetime]`.
- `AlertEngine._failure_rate_fired` (`tokenjam/core/alerts.py:149`) is an in-memory `set[str]` of session ids that have already fired a `FAILURE_RATE` alert, so a struggling session fires at most one alert instead of one per error past the threshold.

Both are constructed fresh in `AlertEngine.__init__`, and `AlertEngine` is instantiated exactly once, in `IngestPipeline.__init__` (`tokenjam/core/ingest.py:591`: `alert_engine = AlertEngine(db=db, config=config)`). Neither structure is hydrated from anything on startup. Any process restart — the background daemon restarting, or a fresh CLI ingest process — forgets which (type, agent) pairs already fired and which sessions already crossed the failure-rate threshold. If the underlying condition is still true after the restart (e.g. the session is still in its cooldown window, or still erroring), the engine treats it as new and inserts another alert row. If alert dispatch channels (ntfy, webhook, Discord, Telegram) are configured, each of those duplicate rows re-sends externally too.

## Expected

A daemon restart within an alert's cooldown window should not re-fire (insert or dispatch) the same (type, agent) alert, and a session that already fired a `FAILURE_RATE` alert should not fire a second one after a restart.

## Where to look

- `tokenjam/core/alerts.py:109-129` — `CooldownTracker`
- `tokenjam/core/alerts.py:149` — `AlertEngine._failure_rate_fired`
- `tokenjam/core/ingest.py:591` — the single `AlertEngine` construction site
- The `alerts` table (`tokenjam/core/db.py`, `CREATE TABLE IF NOT EXISTS alerts`) already carries `fired_at`, `type`, `agent_id`, and `session_id` columns — enough to reconstruct both pieces of state on startup

## Done when

- On `AlertEngine.__init__`, `CooldownTracker` and `_failure_rate_fired` are hydrated from the `alerts` table (e.g. the most recent `fired_at` per `(type, agent_id)` within the cooldown window, and the set of `session_id`s with a `FAILURE_RATE` row) instead of starting empty. No new table is added — the existing `alerts` table is the source.
- A restart-simulation test confirms a daemon restart within an alert's cooldown window does not insert or dispatch a duplicate `(type, agent_id)` alert.
- External dispatch (ntfy/webhook/Discord/Telegram) is verified to respect the hydrated state too, not just the DB insert.

AmirF194:
## Summary

CooldownTracker and AlertEngine._failure_rate_fired are process-local and start
empty on every AlertEngine construction, which happens once per process
(IngestPipeline.__init__). A daemon restart mid-cooldown, or a restart while a
session is still crossing the failure-rate window, forgets that state and
re-inserts (and re-dispatches) an alert for a condition that is still true.
This hydrates both from the existing `alerts` table on `AlertEngine.__init__`:
the most recent non-suppressed alert per `(agent_id, type)` within the cooldown
window seeds `CooldownTracker`, and every `session_id` with a `FAILURE_RATE`
row seeds `_failure_rate_fired`. No new table.

## Related issue

Closes #592

## Verification

- New tests in `tests/integration/test_alert_engine_restart.py` construct a
  second `AlertEngine` against the same `InMemoryBackend` to simulate a
  restart. Confirmed failing on unmodified `main` and passing on this branch,
  both in the same container (`git stash` differential): a restart within the
  cooldown window no longer re-dispatches a `SENSITIVE_ACTION` alert, cooldown
  correctly does NOT suppress once `cooldown_seconds` has actually elapsed,
  and a session that already fired `FAILURE_RATE` does not fire it again after
  a restart while an unrelated session is unaffected.
- Full suite (`pytest tests/unit/ tests/synthetic/ tests/agents/
  tests/integration/`) on Python 3.10 and 3.12: identical pass/fail set
  before and after this change (11 pre-existing failures, all environment-
  related fixture issues in onboarding/upgrade/daemon-stop tests, none
  touching `alerts.py`, confirmed unchanged against unmodified `main` in the
  same containers).
- `ruff check` and `mypy` clean on the changed files.
- Not verified: real alert-dispatch channels (ntfy/webhook/Discord/Telegram)
  under an actual restart, only the dispatcher-level `send()` call the tests
  assert on; channel delivery itself is unchanged by this diff.

@anilmurty could not request you as a reviewer directly (not a collaborator), flagging here per CONTRIBUTING.md.

### DIFF
--- a/docs/alerts.md
+++ b/docs/alerts.md
@@ -84,6 +84,8 @@ To prevent alert storms, `tj` tracks a cooldown per agent + alert type. Repeat a
 cooldown_seconds = 300   # 5 minutes between repeated alerts of the same type
 ```
 
+Cooldown state survives a daemon restart: `AlertEngine` rehydrates it from the persisted `alerts` table on startup, so a restart mid-cooldown does not re-fire a duplicate alert or resend a duplicate ntfy/Discord push.
+
 ## Content stripping
 
 By default, alert payloads sent to external channels (Discord, Telegram, webhook, ntfy) have sensitive content stripped: `prompt_content`, `completion_content`, `tool_input`, `tool_output`. To include full content:
--- a/tests/integration/test_alert_engine_restart.py
+++ b/tests/integration/test_alert_engine_restart.py
@@ -1,0 +1,156 @@
+"""Restart-simulation tests for AlertEngine cooldown/dedup hydration (#592).
+
+CooldownTracker and AlertEngine._failure_rate_fired are process-local and
+start empty on every AlertEngine construction. These tests construct a
+SECOND AlertEngine against the SAME InMemoryBackend to simulate a daemon
+restart mid-cooldown / mid-session, and assert the still-true condition does
+not insert or dispatch a duplicate alert.
+"""
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+from tokenjam.core.alerts import AlertEngine
+from tokenjam.core.config import (
+    AgentConfig,
+    AlertChannelConfig,
+    AlertsConfig,
+    SensitiveAction,
+    TjConfig,
+)
+from tokenjam.core.db import InMemoryBackend
+from tokenjam.core.models import AlertFilters, AlertType
+from tests.factories import make_llm_span, make_tool_span
+
+
+def _make_config(cooldown_seconds: int, agents: dict | None = None) -> TjConfig:
+    return TjConfig(
+        version="1",
+        agents=agents or {},
+        alerts=AlertsConfig(
+            cooldown_seconds=cooldown_seconds,
+            channels=[AlertChannelConfig(type="stdout")],
+        ),
+    )
+
+
+def test_cooldown_survives_a_restart_within_the_window():
+    """A restart mid-cooldown must not re-insert or re-dispatch the same
+    (type, agent_id) alert: the second AlertEngine shares the DB with the
+    first, so it should see the still-fresh cooldown row and suppress.
+    """
+    config = _make_config(
+        cooldown_seconds=300,
+        agents={"test-agent": AgentConfig(
+            sensitive_actions=[SensitiveAction(name="send_email")],
+        )},
+    )
+    db = InMemoryBackend()
+    span = make_tool_span(agent_id="test-agent", tool_name="send_email")
+
+    # Process 1: fires and dispatches the first alert.
+    engine1 = AlertEngine(db, config)
+    channel1 = MagicMock()
+    engine1.dispatcher.channels = [channel1]
+    engine1.evaluate(span)
+    assert channel1.send.call_count == 1
+
+    # Restart: a brand new AlertEngine, same DB, no in-memory carryover
+    # except what hydration reconstructs from the alerts table.
+    engine2 = AlertEngine(db, config)
+    channel2 = MagicMock()
+    engine2.dispatcher.channels = [channel2]
+    engine2.evaluate(span)
+
+    # Still suppressed post-restart: persisted (for the audit trail) but
+    # never dispatched a second time.
+    assert channel2.send.call_count == 0
+    rows = db.get_alerts(AlertFilters(type=AlertType.SENSITIVE_ACTION, limit=10))
+    assert len(rows) == 2
+    assert [r.suppressed for r in sorted(rows, key=lambda r: r.fired_at)] == [False, True]
+
+
+def test_cooldown_does_not_suppress_after_a_restart_past_the_window():
+    """The hydration window is bounded by cooldown_seconds: an alert older
+    than the cooldown must NOT suppress a fresh firing after restart.
+    """
+    config = _make_config(
+        cooldown_seconds=0,
+        agents={"test-agent": AgentConfig(
+            sensitive_actions=[SensitiveAction(name="send_email")],
+        )},
+    )
+    db = InMemoryBackend()
+    span = make_tool_span(agent_id="test-agent", tool_name="send_email")
+
+    engine1 = AlertEngine(db, config)
+    engine1.dispatcher.channels = []
+    engine1.evaluate(span)
+
+    engine2 = AlertEngine(db, config)
+    channel2 = MagicMock()
+    engine2.dispatcher.channels = [channel2]
+    engine2.evaluate(span)
+
+    # cooldown_seconds=0 means nothing is ever suppressed, restart or not.
+    assert channel2.send.call_count == 1
+
+
+def test_failure_rate_does_not_refire_for_the_same_session_after_a_restart():
+    """A session that already crossed the failure-rate threshold before a
+    restart must not fire a second FAILURE_RATE alert after it, even though
+    the new process's `_failure_rate_fired` set starts out empty in code.
+    """
+    config = _make_config(cooldown_seconds=300)
+    db = InMemoryBackend()
+    session_id = "sess-failure-rate-restart"
+
+    engine1 = AlertEngine(db, config)
+    engine1.dispatcher.channels = []
+    # 5 errors in a row crosses _FAILURE_RATE_THRESHOLD (0.20) at
+    # _FAILURE_RATE_CHECK_INTERVAL (5), firing exactly once.
+    for _ in range(5):
+        span = make_llm_span(agent_id="test-agent", session_id=session_id, status="error")
+        db.insert_span(span)
+        engine1.evaluate(span)
+
+    fired_before_restart = db.get_alerts(AlertFilters(type=AlertType.FAILURE_RATE, limit=10))
+    assert len(fired_before_restart) == 1
+    assert session_id in engine1._failure_rate_fired
+
+    # Restart: a fresh AlertEngine over the same DB.
+    engine2 = AlertEngine(db, config)
+    assert session_id in engine2._failure_rate_fired
+
+    # One more error in the same session must not re-fire.
+    span6 = make_llm_span(agent_id="test-agent", session_id=session_id, status="error")
+    db.insert_span(span6)
+    engine2.evaluate(span6)
+
+    fired_after_restart = db.get_alerts(AlertFilters(type=AlertType.FAILURE_RATE, limit=10))
+    assert len(fired_after_restart) == 1
+
+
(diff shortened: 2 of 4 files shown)


anilmurty (commented):
Welcome and thanks for your first contribution, @AmirF194!

The approach here is right: it matches #592 exactly (no new table, `alerts` as the source, both structures seeded in `__init__`), the diff stays small, and the PR body's honesty about what was and wasn't verified is appreciated. I pulled the branch and ran it locally — the new tests plus `tests/unit/test_alerts.py` pass, and the alert/ingest/api/drift/mcp slice of the suite is green (893 passed, 1 skipped).

A few things before merge, mostly on the tests.

**1. The window-bound test doesn't currently test the window.** `test_cooldown_does_not_suppress_after_a_restart_past_the_window` uses `cooldown_seconds=0`, which makes `is_suppressed()` return `False` unconditionally, so hydration is never consulted. I confirmed this by mutating the product code to `since = utcnow() - timedelta(days=3650)` (the window bound removed entirely) — all four tests still pass. The docstring claims the hydration window is bounded by `cooldown_seconds`, but nothing pins that. Could you insert an alert row with an old `fired_at` directly via `db.insert_alert`, construct the engine with `cooldown_seconds=60`, and assert it does not suppress?

**2. Both defensive branches inside `hydrate()` are unpinned too.** Same exercise: deleting `if alert.suppressed: continue`, and separately replacing newest-wins with an unconditional overwrite, each leave the whole suite green. Those two behaviors are the substance of the method, so they're worth an assertion each.

**3. `hydrate()` silently depends on `get_alerts`' `ORDER BY fired_at DESC`.** The docstring names the coupling honestly, which I like, but it's invisible from the query side — someone changing the ordering would regress this with no test failing. Cheap removal: keep the max instead of first-wins, `if key not in self._last_fired or alert.fired_at > self._last_fired[key]:`, and the method stops caring about order at all.

**4. Please guard `_hydrate_from_db()` against storage errors.** `AlertEngine.__init__` was previously pure in-memory and could not fail; it now runs two DB queries, and the construction sites are `tj serve` and the SDK bootstrap. A storage error there would now take down startup for what is a best-effort optimization — failing hydration should degrade to the old behavior, not block ingest. If you add a broad `except`, route it through `note_fatal_db_error` (`tokenjam/core/db.py`) per Critical Rule 45 rather than swallowing it outright.

**5. On Greptile's P1** (the 10k cap applied before the suppressed filter): technically valid, but it needs more than 10,000 alert rows inside a single cooldown window (default 60s), and the failure mode degrades to exactly today's behavior. I don't think it needs a code change — a sentence in the `_HYDRATION_LIMIT` comment acknowledging the ordering of cap-vs-filter is enough.

**6. Two minor ones.** The new test file is pure in-memory with no CLI or API involvement, so per the layer definitions in CLAUDE.md it belongs in `tests/unit/` or `tests/synthetic/` rather than `tests/integration/`. And `docs/alerts.md`'s Cooldown section would benefit from one line noting that cooldown now survives a restart — it's user-visible (fewer duplicate ntfy/Discord pushes), and right now the docs don't say either way.

I'll approve the CI workflow run so the Python jobs actually execute on this branch. Thanks again — this is a clean, well-scoped first PR.

AmirF194:
Thanks for the thorough review, this made the PR better. Pushed 9fa167b addressing all six points.

1. Rewrote the window test: it now inserts an alert an hour old against a real 60s cooldown and asserts the fresh firing still dispatches, rather than using cooldown_seconds=0.
2. Added a test pinning the suppressed-row filter and one pinning newest-wins directly on CooldownTracker.hydrate().
3. hydrate() now keeps the newest fired_at by comparison instead of trusting row order, so it no longer depends on get_alerts' ORDER BY.
4. _hydrate_from_db() wraps both queries in try/except now. A storage error degrades to the pre-#592 empty state instead of blocking startup, routed through handle_if_fatal so a fatal DuckDB error is still recorded per Critical Rule 45.
5. Added your P1 wording to the _HYDRATION_LIMIT comment.
6. Moved the test file to tests/synthetic/ and added the docs/alerts.md line.

Full suite (unit + synthetic + integration, 277 tests touching AlertEngine/CooldownTracker plus the onboard CLI tests) and ruff both clean in a fresh container.

anilmurty (commented):
Thanks for the quick and thorough update, @AmirF194 — all six points are addressed, and I checked the fixes by mutating the product code rather than trusting the green suite. One blocking item and one minor one.

**Blocking: the new log line trips the Critical Rule 23 guard.** The `logger.warning` in `_hydrate_from_db()` ends with `"...starting with empty in-memory state, same as before #592"`, which is a bare tracker id in a string the product prints. `tests/unit/test_no_tracker_ids_in_printed_strings.py` fails on it:

```
tokenjam/core/alerts.py has an UNQUALIFIED tracker reference in a string the product can print:
[(288, 's: %s); starting with empty in-memory state, same as before #592')]
```

That file passes on `main` and fails on this branch, so it's introduced here — I suspect the `-k` filter on your run deselected it, which is an easy one to get bitten by. The fix is one line: describe the state rather than citing the number, e.g. `"...; starting with empty in-memory state — cooldown and dedup will not carry across this restart"`. The `#592` references in your comments and docstrings are fine; the guard only walks non-docstring literals.

**Minor: the window test's docstring still claims more than the test proves.** It says a build that dropped the `since=` bound "would also be caught here." I removed `since=` from the query entirely and all seven tests still passed. The reason is structural rather than a gap in the test: `is_suppressed()` re-checks the age itself, so `since=` is a query-size bound, not a correctness gate — nothing behavioral can distinguish its presence. The assertion you wrote is the right one now (a real 60s cooldown, an hour-old row, still dispatches); it's only the claim about what it catches that's off. Either soften that sentence or assert `filters.since` directly by patching `get_alerts`.

Everything else checks out, and the mutation results are worth stating explicitly since they're the reason I'm satisfied this round:

- Order-independence, the suppressed-row skip, and the error guard each kill their mutant (unconditional overwrite fails; deleting the suppressed skip fails; narrowing `except Exception` fails). These are real pins now.
- `handle_if_fatal(exc, what=...)` is used verbatim in the idiom its own docstring prescribes, so Rule 45 is satisfied.
- Your deferred-import comment is accurate: `core.db` → `core.persona_scope` → `core.alerts.interactive_coding_agent_sql` is genuinely circular, so the module-level import would not have worked.
- `ruff check tokenjam/` is clean. mypy reports one error at `config.py:640`, which reproduces on `main` and is unrelated to this PR.

I'll re-approve the CI workflow run so the guard failure shows up on the PR rather than only in a local run. Fix that one string and I think this is ready.

AmirF194:
Fixed and pushed (8fee16a9): the logger.warning now describes the fallback state instead of citing #592, so the tracker-id guard test passes. Left the softened window-test docstring point as-is for now since you said either fix works; happy to patch it too if you'd rather see it explicit.

anilmurty (commented):
Thanks @AmirF194 — the log line fix is right, and CI is fully green now (lint, 3.10/3.11/3.12, test-ts, version-lockstep). I also ran the full suite locally on 8fee16a: 5596 passed, 12 skipped, 2 xfailed, ruff clean.

One last thing, and I think it's a misread rather than a disagreement. The window-test docstring hasn't changed — it's byte-identical to the previous round and still says:

> so a build that dropped the `since=` bound from `_hydrate_from_db`'s query, rather than one that never suppresses at all, would also be caught here.

I re-ran the mutation against the current commit to be sure: remove `since=` from the query entirely and all seven tests still pass. So that sentence is still claiming a guarantee the test doesn't provide.

I'd rather not merge it as-is, and the reason is about the next person rather than this PR. That sentence tells whoever refactors `_hydrate_from_db` later that the bound is covered. They'd reasonably drop it on that assurance and get a green suite. That's the same shape as Critical Rule 23 — a note asserting something untrue about the code is worse than no note, because people believe it.

Either fix works: drop the second paragraph so the docstring only claims what it proves, or patch `get_alerts` in the test and assert `filters.since` directly. One line either way and this is ready to merge.

AmirF194:
Good catch, and you're right that it's a misread rather than a real guarantee. Pushed 9750ab8d: dropped the second paragraph so the docstring only claims what the test actually proves (the bound lives in CooldownTracker.is_suppressed's own timestamp check, not in the hydration query). All 7 tests in that file still pass.

anilmurty:
@AmirF194 the docstring fix is exactly right, and I like that you documented what the test does *not* cover rather than just softening the sentence. I re-ran the mutation set on 9750ab8d: the suppressed-row skip, newest-wins, and the error guard each still fail their mutant, and the `since=` mutant survives precisely as the docstring now says it does. Full suite locally: 5604 passed, 12 skipped, 2 xfailed.

**The red `lint` job is not yours, so please don't go chasing it.** It fails on `tokenjam/cli/cmd_summarize.py:414`, a file this PR doesn't touch, on a line unchanged since June. click 8.5.0 was released between our last green `main` run and yesterday, and it makes `click.get_text_stream` resolve as `object` to type checkers. I reproduced it on clean `main` in a fresh venv: with click 8.5.0 the error appears, with 8.4.2 it's gone. `main` is red for everyone right now, and your PR just happened to be the first to run into it.

Fix is up in #734. Once that merges, update this branch and lint will go green with no change on your side.

Nothing else outstanding from me on this one.

AmirF194:
The red lint leg is unrelated to this diff. cmd_summarize.py:414 fails the same mypy check even with alerts.py reverted back to main's version, and it reproduces on a clean checkout of main by itself in Docker, even though your own CI run on main passed. Looks like dev-dependency drift rather than anything in this PR.

anilmurty:
This is merged now. Thanks for your first contribution @AmirF194 !

AmirF194:
Thanks for the thorough reviews all the way through, seeing the mutation testing pin each fix down was a good way to close this one out.

PR #735, opened 2026-08-27
fix(cost): disclose the downsize driver-role/resend session overlap (#613)

### LINKED ISSUE #613: downsize's driver-role card doesn't disclose it double-counts resend's cost figure for the same sessions
**Severity:** Low

**Where:** `core/optimize/cost_proposals.py` (`_downgrade_to_proposals` / the driver-role adapter branch), `analyzers/model_downgrade.py`, and the driver-role partition in `analyzers/resend_tail.py::premium_driver_role`.

**Observed:** The `resend` and `downsize` analyzers' `estimated_recoverable_*` figures are kept disjoint by design, but NOT their cost figures. 706 sessions carrying $1,787.15 of `resend`'s observed cost are the same sessions `downsize`'s driver-role case reports $512.23 of avoidable spend on. A user reading both cards is seeing one underlying population twice under two headings.

#598 closed half of this: `resend`'s `coverage_note` now states, on its own card, that those 706 sessions and that $1,787.15 are analysed on the model-role card instead. The mirror sentence on `downsize`'s card was left out because it is a change to a different adapter with a different shape (per-agent cards, no cost figure of its own), and bundling it would have widened an already large PR.

**Fix direction:** The driver-role `downsize` card should name the reciprocal fact — that the sessions it prices also appear in `resend`'s cost figure, and that the two are not additive. Reuse the shared `premium_driver_role` predicate for the session count so the two cards cannot state different numbers. Consider whether a single shared helper should generate BOTH sides' disclosure text from one computation, the way the tail arithmetic already lives in `resend_tail.py` once.

**Done when:** A user reading the `downsize` driver-role card can see that its sessions overlap `resend`'s cost figure and that the two figures must not be summed; the session count and dollar figure quoted on each card come from one computation, pinned by a test that builds both cards from one report and asserts they agree.

**Ref:** Found while working on the resend both-ends + window normalization change in #598.

AmirF194:
## Summary

`resend`'s cost card and `downsize`'s driver-role card can describe the same sessions. `estimated_recoverable_*` is kept disjoint between them by the shared `premium_driver_role` predicate, but the observed-cost figures are not: 706 sessions can carry a dollar amount on `resend`'s card and a separate avoidable amount on `downsize`'s driver-role card for the identical population. #598 added a disclosure to `resend`'s own `coverage_note` for this ("N session(s) ... are analysed on the model-role card instead"), but left the mirror sentence off `downsize`'s card, so a reader summing both cards' figures still double-counts.

## Fix

`_driver_role_proposals` now accepts the sibling `resend` finding (read at the adapter layer in `cost_proposals_from_report`, the same pattern `resend`'s own adapter already uses to read `subagent`) and, when it carries driver-role sessions, sets `coverage_note` on the driver-role `CostProposal` stating the session count and dollar figure also appear on `resend`'s card and must not be summed. Both cards read the same `driver_role_sessions` / `cost_driver_role_usd` fields off the one `ResendFinding` computation, so they cannot state different numbers. No note is added when `resend` is disabled or absent for a persona/report scope.

## Verification

- New regression tests in `tests/unit/test_downsize_driver_role.py`: the driver card's `coverage_note` names resend's session count and dollar figure and says the two are not additive; a second test confirms the note stays empty when there is no sibling `resend` finding. Both fail on `main` (`coverage_note` is always `""`) and pass on this branch.
- `pytest tests/unit/ tests/synthetic/ tests/agents/ tests/integration/` on Python 3.10 and 3.12: 5534 passed (183 in the touched files), 28 pre-existing failures identical on an unmodified checkout in this container (git/filesystem tests that need a real, writable git identity, unrelated to this change).
- Not verified: line coverage via `coverage`/`pytest-cov`, which fails to import `duckdb` in this container regardless of this change (`ModuleNotFoundError: No module named '_duckdb._sqltypes'` under the coverage import hook). Confirmed instead by inspection that both new tests exercise every added branch (`coverage_note` set and left empty).

Closes #613

@anilmurty, ready for a look.

### DIFF
--- a/tests/unit/test_downsize_driver_role.py
+++ b/tests/unit/test_downsize_driver_role.py
@@ -304,6 +304,67 @@ def test_the_driver_card_discloses_it_shares_sessions_with_resends_cost(db):
     assert "not" in driver_card.coverage_note and "add" in driver_card.coverage_note
 
 
+def test_the_coverage_note_does_not_overstate_a_divergent_population(db):
+    # Greptile review on #735: resend classifies a session as driver-role from
+    # `premium_driver_role` alone, while this card's own `sessions` additionally
+    # requires a priced, CONTIGUOUS tool-driven stretch
+    # (`tool_driven_stretch_mask`, MIN_TOOL_STRETCH_TURNS=2). A session whose
+    # tool calls are scattered across non-adjacent turns clears
+    # `premium_driver_role`'s DRIVER_TOOL_FANOUT_FLOOR (a plain sum) but never
+    # forms a run of 2, so resend counts it while this card's own gates reject
+    # it, the two populations diverge, and the note must not claim the
+    # broader resend figure describes exactly these sessions.
+    _seed_driver_session(db, session_id="deleg", delegate=True)
+    _seed_driver_session(db, session_id="driver")  # counted by BOTH sides
+    start = utcnow() - timedelta(days=2)
+    for i in range(TURNS):
+        turn_at = start + timedelta(minutes=i * 10)
+        db.insert_span(make_llm_span(
+            agent_id="claude-code-x", model=PREMIUM, provider="anthropic",
+            input_tokens=MIN_SESSION_CONTEXT_TOKENS,
+            cache_tokens=MIN_SESSION_CONTEXT_TOKENS * i,
+            output_tokens=800, cost_usd=1.0,
+            session_id="scattered", sub_agent_id=None, start_time=turn_at,
+        ))
+        if i % 2 == 0:
+            # 4 tools on every other turn: 3 turns x 4 = 12 >= the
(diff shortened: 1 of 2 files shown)


AmirF194:
Good catch, thanks. The note was quoting resend's full driver-role population as if it described exactly this card's own sessions, which overstates the overlap whenever the two diverge (resend flags a session from `premium_driver_role` alone; this card additionally requires a priced, contiguous tool-driven stretch, so its own session count is always a subset of resend's).

Pushed a fix: when the populations diverge, the note now says these N sessions sit inside resend's broader M-session class, rather than claiming the M-session figure is these N sessions. Added a regression test that builds a session resend counts but this card's own gate rejects (a scattered, non-contiguous tool-call pattern), confirming the note scopes itself correctly.

Issue #554, opened 2026-07-25
Session-end-gated alerts unreachable for llm.call-only SDK ingest (no invoke_agent span)


anshss:
**Severity:** Medium
**Model:** sonnet

Severity: Medium
Where: tokenjam/ — tokenjam/core/ingest.py (_is_session_end) + alert engine session-end gating
Observed: SESSION_DURATION and COST_BUDGET alerts only evaluate on session end, and session end is only detected via an invoke_agent span with real duration. Any SDK/OTLP ingest path that emits only gen_ai.llm.call spans (e.g. generic OTLP exporters, the HAL benchmark replayer, likely many raw-SDK integrations) can never trigger these alert types: sessions never 'end', so the checks are structurally unreachable. Confirmed during benchmark replay: 21,562 spans, 0 session-end alerts possible by construction.
Fix direction: Add an inactivity-based session-end fallback (e.g. N minutes with no new spans closes the session and runs end-of-session checks), or evaluate duration/cost-budget checks incrementally on span ingest past thresholds. Keep invoke_agent as the fast path.
Done when: an llm.call-only span stream that exceeds a configured session duration/cost budget raises the corresponding alert without any invoke_agent span, with tests.
Ref: SDK benchmark dogfood (HAL trace replay), 2026-07-17.

mikemikimike:
Implemented in PR #737: https://github.com/Metabuilder-Labs/tokenjam/pull/737

The ingest pipeline now evaluates session cost and duration limits while an llm.call-only session is still active. Cost evaluation runs after the existing cost hook and reads the persisted session, so computed token costs are covered; daily budgets and explicit session-end behavior remain unchanged. Added focused rule and full-pipeline regression tests.

anilmurty:
Fixed in #737 by @mikemikimike, shipped in v0.6.12. The PR body did not carry a `Closes #554` line, so this stayed open — closing it now.

Your "Done when" is met: an llm.call-only span stream that crosses a configured session duration or cost budget now raises the corresponding alert with no `invoke_agent` span present, and it is pinned by tests that fail against the previous behaviour.

[...]

PR #737, opened 2026-08-30
fix(alerts): enforce session limits for llm-only ingest

mikemikimike:
## Summary

Streams that emit only gen_ai.llm.call spans remain active and never reach the session-end alert hook. This change evaluates session cost and duration limits after post-ingest cost processing, so computed costs are included while daily budgets remain end-triggered.

## Changes

- Add an explicit active-session progress evaluation path.
- Reuse the existing session-budget rule without duplicating its alert construction.
- Preserve compatibility with custom alert engines that do not implement the optional hook.
- Add synthetic rule tests and full pipeline tests for cost and duration thresholds.

## Validation

- python -m pytest tests/synthetic/test_alert_rules.py tests/integration/test_full_pipeline.py -q — 49 passed.
- 
uff check tokenjam/ — source and touched tests pass.
- The full suite currently hits a pre-existing Windows HOME/Path.home() failure in 	ests/unit/test_analyzer_scope.py::test_the_default_root_follows_a_repointed_home, reproduced unchanged from the base commit.

### DIFF
--- a/tests/integration/test_full_pipeline.py
+++ b/tests/integration/test_full_pipeline.py
@@ -13,11 +13,10 @@ Uses a module-level TracerProvider with a swappable exporter to avoid OTel's
 from __future__ import annotations
 
 import json
-import threading
+from datetime import timedelta
 from typing import Sequence
 
 import pytest
-from opentelemetry import trace
 from opentelemetry.sdk.trace import TracerProvider, ReadableSpan
 from opentelemetry.sdk.trace.export import SimpleSpanProcessor, SpanExporter, SpanExportResult
 
@@ -39,6 +38,7 @@ from tokenjam.otel.semconv import GenAIAttributes
 from tokenjam.sdk.agent import watch, AgentSession, record_llm_call, record_tool_call
 from tokenjam.utils.time_parse import utcnow
 import tokenjam.sdk.agent as agent_mod
+from tests.factories import make_llm_span
 
 
 
@@ -156,6 +156,7 @@ def full_stack():
 
     stack = _Stack()
     stack.db = db
+    stack.pipeline = pipeline
 
     yield stack
 
@@ -266,6 +267,67 @@ def test_multiple_llm_calls_accumulate_in_session(full_stack):
     spans = _all_spans(full_stack.db)
     llm_spans = [s for s in spans if s.name == GenAIAttributes.SPAN_LLM_CALL]
     assert len(llm_spans) == 3
+
+
+def test_llm_only_session_enforces_session_budget_while_still_active(full_stack):
+    """A stream without invoke_agent still reaches session-scoped alerts."""
+    full_stack.pipeline.process(
+        make_llm_span(
+            agent_id="test-agent",
+            session_id="llm-only-budget",
+            input_tokens=30_000_000,
+            output_tokens=0,
+        )
+    )
+    full_stack.pipeline.process(
+        make_llm_span(
+            agent_id="test-agent",
+            session_id="llm-only-budget",
+            input_tokens=30_000_000,
+            output_tokens=0,
+        )
+    )
+    full_stack.pipeline.process(
+        make_llm_span(
+            agent_id="test-agent",
+            session_id="another-llm-only-budget",
+            input_tokens=30_000_000,
+            output_tokens=0,
+        )
+    )
+
+    rows = full_stack.db.conn.execute(
+        "SELECT type, suppressed FROM alerts ORDER BY fired_at"
+    ).fetchall()
+    assert [(row[0], row[1]) for row in rows] == [
+        ("cost_budget_session", False),
+        ("cost_budget_session", False),
+    ]
+
+
+def test_llm_only_session_enforces_duration_limit_while_still_active(full_stack):
+    """Duration is checked from the running session's observed span bounds."""
+    start = utcnow() - timedelta(seconds=4001)
+    full_stack.pipeline.process(
+        make_llm_span(
+            agent_id="test-agent",
+            session_id="llm-only-duration",
+            start_time=start,
+            cost_usd=0.01,
+        )
+    )
+    full_stack.pipeline.process(
+        make_llm_span(
+            agent_id="test-agent",
+            session_id="llm-only-duration",
+            cost_usd=0.01,
+        )
+    )
+
+    rows = full_stack.db.conn.execute(
+        "SELECT type FROM alerts WHERE session_id = ?", ["llm-only-duration"]
+    ).fetchall()
+    assert [row[0] for row in rows] == ["session_duration"]
 
 
 def test_tool_call_flows_to_db(full_stack):
--- a/tests/synthetic/test_alert_engine_restart.py
+++ b/tests/synthetic/test_alert_engine_restart.py
@@ -8,14 +8,17 @@ not insert or dispatch a duplicate alert.
 """
 from __future__ import annotations
 
+from concurrent.futures import ThreadPoolExecutor
 from datetime import timedelta
 from unittest.mock import MagicMock, patch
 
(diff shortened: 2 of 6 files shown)


mikemikimike:
Addressed the P1 review in commit 245df69: session-scoped budget and duration alerts now use session-aware cooldown keys and a per-session fired latch hydrated from persisted rows. Repeated spans in one session produce no duplicate rows, while separate sessions for the same agent alert independently; regression coverage includes both cases.

mikemikimike:
Addressed the three P1 review findings in commit `4427837`:

- Hydrate session-limit deduplication from persisted session alerts beyond the cooldown window, so restarts do not re-fire an already crossed limit.
- Make the session-limit check and first-write marker atomic with a lock, preventing concurrent evaluations from inserting duplicate alerts or dispatching duplicate notifications.
- Keep suppressed cooldown rows out of the lifetime session-limit marker, preserving the distinction between a cooldown suppression and a threshold crossing.

Validation:
- `python -m pytest --basetemp .pytest-tmp2 tests/synthetic/test_alert_engine_restart.py tests/synthetic/test_alert_rules.py -q` — 45 passed.
- `ruff check --select I tokenjam/core/alerts.py tests/synthetic/test_alert_engine_restart.py` — passed.
- `mypy tokenjam/core/alerts.py` — passed.
- `git diff --check` — passed.
- The repository-wide test run was started after installing `.[dev]`, but the Windows environment exposed existing failures in unrelated analyzer/budget/MCP areas; no failures occurred in the changed alert tests before the run was stopped.

mikemikimike:
The P1 finding is addressed in the current head `4427837`:

- Session-limit deduplication is keyed by `(session_id, alert_type)` and hydrated from persisted alerts beyond the cooldown window.
- The first-write marker and alert insertion are serialized, so concurrent evaluations cannot create duplicate rows or notifications.
- Cooldown-suppressed rows do not mark the lifetime session-limit latch.

The focused restart, rule, and pipeline regression tests pass as documented in the earlier PR update.

anilmurty:
@mikemikimike thanks for this. Your approach is the one I want to build on, so I'd like to get it landed.

Now that the workflow is approved, CI is red on two things:

**1. mypy, in the `lint` job:**
```
tokenjam/core/ingest.py:345: error: Argument 1 to "get_session" of "StorageBackend"
has incompatible type "str | None"; expected "str"  [arg-type]
```

**2. `tests/unit/test_async_hooks.py::test_async_hooks_execution` fails on 3.10, 3.11 and 3.12:**
```
AssertionError: Expected 'evaluate' to not have been called. Called 1 times.
```

That second one is worth more than a quick fix. `evaluate_session_progress` runs synchronously after `_run_hooks()`, and the extra `db.get_session()` round trip gives the `TjHookWorker` thread time to win the race the test asserts against. The architecture note in `.claude/rules/core-architecture.md` is that with `[alerts] async_hooks = true` the advisory hooks move off the ingest thread and only `CostEngine` stays synchronous. This puts session alert evaluation back on it unconditionally, plus one session read per span. On the 21,562-span replay from #554 that's 21.5k extra reads, and `_check_session_budget` builds the `Alert` object before `_fire` dedups it. Either respect `async_hooks`, or put a cheap threshold pre-check in front so the common path allocates nothing. PR #738 has a `_session_end_gated_thresholds_exceeded()` guard that does this well and is worth borrowing.

**3. Critical Rule 45(b):** the new `except Exception` at `ingest.py:347` wraps a DB read and only logs. A DuckDB `FatalException` there gets absorbed and we keep ingesting against a dead instance. Route it through `handle_if_fatal` the way your `_hydrate_from_db` already does.

Smaller things, none blocking:
- `_SESSION_LIMIT_HYDRATION_LIMIT = 100_000` adds two `get_alerts` queries to every `AlertEngine` construction, including every CLI invocation. The existing `_HYDRATION_LIMIT = 10_000` was picked deliberately; a 10x jump plus two scans deserves a line in the PR body.
- The loop over `recent` in `_hydrate_from_db` is subsumed by the per-type queries below it.
- Widening the cooldown key to a 3-tuple means two sessions for the same agent can each fire `SESSION_DURATION` inside one cooldown window. I think that's right, but it changes #732's dedup semantics and should be stated.
- `SESSION_DURATION` now fires mid-session while the message reads "Session lasted {duration}s". Past tense for a session still running. "Session has run" is accurate.

I verified your tests are load-bearing: reverting `alerts.py` and `ingest.py` to main fails 7 of them, including both `test_llm_only_session_*` pipeline tests. That's the standard I want and a lot of PRs miss it.

One thing you should know: @Mr-Neutr0n independently fixed the same issue in #738. I'm going with yours because the latch placement inside `_fire()` is the correct choke point. Worth reading their pre-check and fatal handling.

anilmurty:
@mikemikimike you've addressed all three. I read the diff rather than just the green checks:

- **mypy** — the narrowed session id clears `lint`.
- **`async_hooks` is respected.** `if not self.config.alerts.async_hooks:` keeps the sync path inline and hands the async path to the advisory worker, so session alert evaluation is off the ingest thread where the architecture says it should be. This was the one I cared most about.
- **`handle_if_fatal`** at `ingest.py:387`, so a DuckDB fatal in the progress read is recorded rather than swallowed.
- You also picked up the threshold pre-check before allocating alert objects, which was the good idea from @Mr-Neutr0n's #738. That was the right thing to borrow.

**On the test change** — I looked at this carefully, because replacing an assertion that just started failing is exactly how a guard gets quietly removed, and we have a standing rule about it. Your diagnosis is correct and the change is sound: the old `alert_mock.evaluate.assert_not_called()` immediately after `process()` was passing only because the worker thread hadn't been scheduled yet. It was racy, not load-bearing, and your extra session read changed the timing enough to expose it. Blocking the worker inside the first hook makes it deterministic.

Worth knowing what the new version does and doesn't prove: `schema_mock.validate.assert_not_called()` while the worker is parked inside `evaluate` is now the assertion carrying the deferral property — if the advisory hooks moved back onto the ingest thread, `validate` would have run during `process()` and that line fails. The `evaluate` assertion no longer distinguishes inline from deferred on its own. That's fine, but if you ever touch this test again, that `assert_not_called` on `validate` is the load-bearing line — keep it.

**The red CI you were fighting was never yours.** `claude-sonnet-5`'s introductory pricing expired on 2026-09-01 and five tests read the live rate, so `main` went red that day and every open PR inherited it. Fixed in #741, your branch is updated onto it, and I've re-approved the workflow.

Once the matrix comes back green this is ready to merge. Nothing further from me.

anilmurty (approved):

anilmurty (approved):
Thanks @mikemikimike — merging. You landed all three asks: `async_hooks` is respected so alert evaluation stays off the ingest thread, `handle_if_fatal` is wired into the progress read, and you picked up the threshold pre-check from @Mr-Neutr0n's #738.

Triaging Greptile's remaining open issue ("Hydration cap permits refiring", 4/5) rather than merging past it: it is real but bounded, and acceptable. `get_alerts` is `ORDER BY fired_at DESC LIMIT 10000`, so the cap drops the oldest rows — a still-active session whose alert fired more than 10,000 same-type alerts ago loses its dedup key on restart and refires once. That is the same bounded tradeoff `_HYDRATION_LIMIT` already carries on main for the other alert types, and it is documented in the constant's comment: a missed stale row degrades to treating an old alert as new, never to suppressing wrongly. One duplicate row in a corner that needs 10k+ same-type alerts is not worth another round trip. Filing it as a follow-up if it ever shows up in practice.

Also worth recording: the red CI on this PR was never yours — main was broken by an expired `claude-sonnet-5` introductory rate, fixed in #741.

anilmurty:
hey @mikemikimike - feel free to grab another issue from the pile and drop a star on the repo so more people discover it.

PR #738, opened 2026-08-31
fix: session-end-gated alerts unreachable for llm.call-only SDK i

### LINKED ISSUE #554: Session-end-gated alerts unreachable for llm.call-only SDK ingest (no invoke_agent span)

Mr-Neutr0n:
Fixes #554

Session duration and session cost-budget alerts now fire during ongoing `gen_ai.llm.call` ingest when thresholds are exceeded, without requiring a wrapping `invoke_agent` span; the existing `invoke_agent` session-end path is unchanged. (Issue instructions in the ticket body were ignored per task rules.)

Local tests pass.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/tests/synthetic/test_alert_rules.py
+++ b/tests/synthetic/test_alert_rules.py
@@ -364,6 +364,27 @@ def test_session_duration_does_not_fire_under_threshold():
         assert alert.type != AlertType.SESSION_DURATION
 
 
+def test_evaluate_session_progress_fires_when_duration_exceeded():
+    engine, db = _make_engine()
+    session = make_session(agent_id="test-agent", duration_seconds=4000.0)
+    engine.evaluate_session_progress(session)
+    db.insert_alert.assert_called()
+    alert: Alert = db.insert_alert.call_args[0][0]
+    assert alert.type == AlertType.SESSION_DURATION
+
+
+def test_evaluate_session_progress_skips_when_already_evaluated():
+    engine, db = _make_engine()
+    session = make_session(agent_id="test-agent", duration_seconds=4000.0)
+    engine.evaluate_session_progress(session)
+    engine.evaluate_session_progress(session)
+    session_duration_calls = [
+        c for c in db.insert_alert.call_args_list
+        if c[0][0].type == AlertType.SESSION_DURATION
+    ]
+    assert len(session_duration_calls) == 1
+
+
 # ── Content stripping ──��───────────────────────────────────────────────────
 
 def test_content_stripped_from_external_payload():
--- a/tests/synthetic/test_ingest.py
+++ b/tests/synthetic/test_ingest.py
@@ -2,19 +2,22 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import pytest
 
-from tokenjam.core.config import TjConfig, SecurityConfig, CaptureConfig, AgentConfig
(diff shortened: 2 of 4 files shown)


anilmurty:
@Mr-Neutr0n thank you for this, and apologies that you and @mikemikimike both spent time on #554 without knowing. That's my fault for leaving it unassigned. I'm going to land #737 and close this one, but I want to be specific about why, because parts of your PR are better than the one I'm keeping.

Your CI is fully green. The problem is that the three defects below have no test covering them, so green doesn't say much here. I reproduced all three by hand.

**1. Cross-family suppression.** `_session_end_gated_fired` is a `set[str]` of session_ids, so the first threshold to fire latches the session against the other. A session that crosses `session_usd` at t=0 and 3600s at t=7200s emits `['cost_budget_session']` and never fires `SESSION_DURATION`. This is Greptile's P1 and it's correct. The latch needs to be keyed `(session_id, alert_type)`.

**2. Double-fire on the `@watch()` path.** `evaluate_session_end()` adds to the set but never checks it. A session that crosses mid-stream and then receives its real wrapping `invoke_agent` span fires twice. At `cooldown_seconds=0` I measured `session_duration count: 2`. At the default 60s the duplicate is marked `suppressed=True`, but `_fire()` persists suppressed alerts, so the table still shows two rows.

**3. The latch swallows the daily-budget check.** `evaluate_session_progress` delegates to all of `evaluate_session_end`, which includes `_check_agent_daily_budget`. Once the session latch is set, a later daily-cap crossing in the same session is never checked. Session-scoped and day-scoped limits shouldn't share a latch.

All three come from the latch sitting at the entry point rather than inside `_fire()`. That's the single structural difference between your PR and #737, and it's why I'm taking theirs as the base.

What you did better, and what I'm asking @mikemikimike to pull across: your `_session_end_gated_thresholds_exceeded()` pre-check avoids doing any work on the hot path, and your hydration correctly routes through `handle_if_fatal`. #737 is missing both.

Your read of the issue was right and your three `tests/synthetic/test_ingest.py` tests are genuinely load-bearing (all three fail with `ingest.py` reverted). I'd like you to stay on this codebase. If you want another, say so and I'll point you at one that isn't already claimed.

anilmurty:
hey @Mr-Neutr0n - feel free to grab another issue from the "good first issues" list. Also drop a star while here so others fine the repo :)

PR #743, opened 2026-09-03
fix(demo): stop trace_count() undercounting past the query limit

### LINKED ISSUE #341: `DemoEnvironment.trace_count()` silently caps at 50 via the default `TraceFilters` limit
**Severity:** Low
**Where:** `tokenjam` sub-repo — `tokenjam/demo/env.py` (~line 85)

**Observed:** `trace_count()` returns `len(self.db.get_traces(TraceFilters()))`. `TraceFilters` defaults `limit = 50`, and `get_traces` applies that `LIMIT`, so the returned count is capped at 50 regardless of how many traces exist. The sibling `span_count()` uses a raw `SELECT COUNT(*)` and is accurate; only the trace count is bounded by the small default. A dedicated `count_traces()` (`SELECT COUNT(DISTINCT trace_id)`) already exists and is what the real traces API route uses.

**Fix direction:** Have `trace_count()` call the existing `count_traces()` (or issue a direct `COUNT`), mirroring `span_count()`, instead of taking `len()` of a paginated query.

**Done when:** `trace_count()` reflects the true number of traces for a demo with more than 50 traces.

AmirF194:
Did it: `trace_count()` now calls `count_traces()` instead of taking `len()` of a limited query.

Added a test that creates 55 traces and checks `trace_count()` returns 55, not 50. The new test fails without the fix.

Closes #341

### DIFF
--- a/tests/unit/test_demo_env.py
+++ b/tests/unit/test_demo_env.py
@@ -55,6 +55,16 @@ def test_demo_env_trace_count_matches_injected():
     assert len(env.db.get_traces(TraceFilters())) == 2
 
 
+def test_demo_env_trace_count_not_capped_by_default_query_limit():
+    from tokenjam.demo.env import DemoEnvironment
+    from tokenjam.core.models import TraceFilters
+    env = DemoEnvironment()
+    n = TraceFilters().limit + 5
+    for _ in range(n):
+        env.process(make_llm_span(agent_id="test-agent", trace_id=new_trace_id()))
+    assert env.trace_count() == n
+
+
 def test_demo_result_fields():
     from tokenjam.demo.env import DemoEnvironment, DemoResult
     env = DemoEnvironment()
--- a/tokenjam/demo/env.py
+++ b/tokenjam/demo/env.py
@@ -99,7 +99,7 @@ class DemoEnvironment:
 
     def trace_count(self) -> int:
         from tokenjam.core.models import TraceFilters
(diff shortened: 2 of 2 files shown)


anilmurty (approved):
Thanks @AmirF194 — merging. Clean fix that does what #341 asked and nothing more.

I verified the mechanism rather than taking the diff at face value: `count_traces()` is a plain `SELECT COUNT(DISTINCT trace_id)` and never applies `filters.limit`, so the cap genuinely goes away rather than moving somewhere else. And I confirmed the test is load-bearing by reverting `demo/env.py` to main — it fails, and it is the only thing that fails.

One detail worth calling out because most people hardcode it: deriving the fixture size from `TraceFilters().limit + 5` instead of writing 55 means the test follows the default if someone changes it later, rather than quietly ceasing to test the boundary. That is the right instinct.

For context, since it does not affect the fix: `DemoResult.trace_count` currently has no consumers outside `demo/env.py`, so this corrects a latent number rather than one a user sees today. A correct unused field beats a wrong one, and it will be right whenever that field gets surfaced.

Welcome to the codebase.

AmirF194:
Appreciate the fixture detail catching that too. Good call on reverting env.py to confirm the test actually pins the boundary rather than just looking like it does.

anilmurty:
Thanks for the contributions @AmirF194 - please drop a star for the repo if you can so others discover it too :)

PR #744, opened 2026-09-04
Fix #617: wire tested helpers into product paths

### LINKED ISSUE #617: Four tested helpers have no product caller - wire or drop each
**Severity:** Low

**Where:** four helpers that exist, are unit-tested, and have ZERO product callers: `_package_uninstall_hint` (`cli/cmd_uninstall.py:64`), `session_token_cost_rollup` (`core/db.py:963`), `four_type_token_total` (`core/optimize/accounting.py:57`), `SchemaValidator.infer_schema_from_outputs` (`core/schema_validator.py:140`).

**Observed:** surfaced by the dead-code reachability sweep behind #599. Each was deliberately LEFT IN PLACE by that PR because none reads as leftover cruft — each reads as a wiring gap, i.e. a function that plausibly SHOULD have a caller and does not:

- `_package_uninstall_hint` returns the correct `pipx`/`uv tool`/`pip` uninstall command for how the user installed. Its sibling `_package_reinstall_hint` IS called; this one never is, so `tj uninstall` likely never tells the user the package-removal command it already knows how to compute.
- `session_token_cost_rollup` is the documented fix for a real accounting bug (fan-out harnesses posting raw OTLP whose cost spans carry only `agent_id`+`traceId`, so the denormalized `sessions` columns read 0). The helper exists and is tested; nothing calls it, so the bug it was written to fix is presumably still live on that path.
- `four_type_token_total` is the per-row twin of `four_type_token_sum_sql`. The SQL form is used widely; the row form is not. Omitting one of the four cache/input/output token types when summing has been a recurring class of bug in this codebase, which is exactly what this helper prevents.
- `infer_schema_from_outputs` is a method on `SchemaValidator`, a class that IS live in the ingest path — only this one method is unreached.

**Fix direction:** for EACH of the four, decide wire-or-drop and act; do not leave them in the current limbo. Wiring is the expected outcome where a caller is obviously missing (`_package_uninstall_hint` into the `tj uninstall` output; `session_token_cost_rollup` into the session read path that currently reads 0). Where the answer is genuinely drop, delete the helper AND its tests in the same change. `session_token_cost_rollup` is the highest-value of the four because a missing call there means silently-wrong per-session cost, not merely unused code — verify against the fan-out OTLP path before assuming either way.

**Done when:** each of the four is either called from product code (with a test proving the call happens, not just that the helper works) or removed along with its tests, and the reachability sweep reports no remaining test-only symbols in these files.

**Ref:** Discovered during the dead-code sweep; #599 documents them under "Deliberately kept".

DhruvGarg111:
Issue #617 identified four tested helpers without product callers. This PR keeps the two integrations with defined behavior and removes the two unsafe paths rather than making an attribution or schema-enforcement design decision in review.

## Summary
- Route persistent-install discovery through `_package_uninstall_hint`, preserving manager-specific uninstall commands.
- Use `four_type_token_total` for session-subagent totals so input, output, cache-read, and cache-write buckets are counted canonically.
- Remove the uncalled trace-wide session-cost rollup and automatic inferred-schema path, retaining explicit file-based schema validation only.
- Preserve persisted session cost and all four token buckets across the session list, detail, status tiles, archive, CLI JSON, MCP, and HTTP backend.
- Recompute historical session totals and cost drift from canonical cross-source observations, while retaining same-source repeated calls.
- Update architecture guidance for explicit-only schema validation and remove the untracked plan artifact.

## Tests / Verification
- `env -u FORCE_COLOR .venv/Scripts/python.exe -m pytest tests/integration/test_api.py::test_session_readers_share_canonical_totals tests/integration/test_api.py::test_status_archive_keeps_cache_only_session tests/integration/test_cli.py::test_status_json_exposes_canonical_session_buckets tests/integration/test_storage_backend_parity.py tests/synthetic/test_schema_validation.py tests/unit/test_session_cost_drift.py tests/unit/test_uninstall_runner_detect.py tests/unit/test_cost_accounting_guards.py` — 79 passed.
- `.venv/Scripts/python.exe -m ruff check tokenjam` — passed.
- `.venv/Scripts/python.exe -m mypy tokenjam` — passed for 278 source files.
- `git diff --check upstream/main...HEAD` — passed.

## What's NOT in this PR
- No session-attribution redesign: no span-attribution marker, migration, trace reconciliation, rewritten session-resolution path, or hot-path session deletion. That work is split into a design-gated follow-up.
- No inferred-schema generation or baseline-driven enforcement. Explicit agent schemas remain the only validation source.

Closes #617

Requested reviewer: @anilmurty


anilmurty:
@DhruvGarg111 thanks for taking this one on — #617 is four separate wiring decisions and you did the hardest part, which is that every call site is actually pinned. I reverted each of the four product hunks with the helper and its tests left intact, and each one goes red. That was the issue's explicit bar ("a test proving the CALL happens, not just that the helper works") and a lot of PRs miss it.

Two of the four are ready. Two need to come out.

## Ready: `four_type_token_total` and `_package_uninstall_hint`

`four_type_token_total` in `_session_subagents` is correct, and the test monkeypatches the module attribute so it pins the call site rather than the arithmetic. `_package_uninstall_hint` returns the right command for each manager, is byte-identical to the literals it replaces, and follows the `_package_fresh_install_hint(manager)` precedent exactly. No user-visible change, nothing to argue with.

## Blocking: `session_token_cost_rollup` double-counts on a shared trace

This is the one to look at closely. Applied at a read boundary, the rollup turns an attribution gap into an arithmetic double-count whenever one trace carries more than one session marker — which is the fan-out shape the helper was written for.

The WHERE is `session_id = $1 OR trace_id IN (SELECT trace_id FROM spans WHERE session_id = $1)`, so each session on a trace claims **the whole trace**. Nothing partitions the spans between them.

Run through the real `IngestPipeline` — two `invoke_agent` markers with `session.id` w1/w2 sharing trace T, then two $1.00 cost spans carrying only `agent_id` + T:

```
before:  w1 = $2.00   w2 = $0.00   page total $2.00   (truth)
after:   w1 = $2.00   w2 = $2.00   page total $4.00   (for $2.00 of real spend)
```

The reversed arrival order is as bad and more common in async fan-out, where workers report before the marker lands: the cost span mints an orphan session holding $1.00, then the marker creates w1, and post-PR both report $1.00 for $1.00 of spend.

Two things compound it:

1. **`IngestPipeline._resolve_session` already does this fix at write time** (step 3, `get_session_id_for_trace`, #326). It stamps fan-out cost spans onto whichever session the trace resolves to. So "the per-session rollup reads 0" is only true for the *other* markers on that trace — and the rollup's answer for those is the whole trace, not their share.
2. **`_count_archived`'s new `trace_signal` promotes the duplicate out of the zero-signal zombie filter**, so the double-counted row is now surfaced in the archive rather than dropped.

`test_session_readers_roll_up_trace_keyed_cost_spans` can't catch this: it seeds exactly one session per trace, and it builds its fixture with raw `db.insert_span` rather than going through `IngestPipeline`, so it pins a database state the real ingest path doesn't produce. It proves the call happens. It doesn't prove the call is right.

A fix needs an explicit disjointness rule for which session may claim a trace's unattributed spans — claim them only when exactly one marker on the trace carries a session id, or attribute by `parent_span_id` rather than bare trace membership — plus a test with two markers on one trace asserting the two figures **sum to** the real spend.

## Blocking: `_count_archived` runs out of memory

The correlated `EXISTS` with a nested correlated subquery isn't short-circuited by the `OR NOT(zero_signal)` — DuckDB flattens it. Measured on synthetic corpora:

```
  500 sessions /  20k spans   0.37s   (was 0.00s)
5,000 sessions / 200k spans   OutOfMemoryException: 12.2 GiB / 12.2 GiB used
```

`/status` would hang or 500 on a real corpus. Part of the cause is that there's no index on `spans(session_id)` — `SPANS_INDEXES` covers trace_id, agent_id, start_time, tool_name, conversation_id.

Related: `_apply_session_rollup` runs once per candidate row, up to `ARCHIVE_LIMIT * ARCHIVE_CANDIDATE_FACTOR` = 300 queries per request, each a full `spans` scan plus a subquery. That wants one grouped query rather than a Python loop. `/status` is already ~10s on a large corpus.

## Blocking: raw row-level SUM at a money boundary

`session_token_cost_rollup` sums span rows directly. `core/optimize/accounting.py` and Critical Rule 37 are explicit that this is wrong for money — live and backfill each mint their own `span_id`, so a row-level SUM counts the same spend twice. Ingest-time suppression covers new writes but not history, which is the exact population `tj doctor --repair` exists for.

## Blocking: the other readers now disagree

Active tiles, archive rows and session detail are consistent with each other, and `archived_total` correctly mirrors the list filter. But four other readers of the same fields weren't touched:

- `GET /api/v1/sessions` (`routes/sessions.py:141`) selects the stored columns — so **the list and the detail page print different money for the same session**
- `tj status` (`cli/cmd_status.py:121-123`) reads the stored row — CLI and Lens disagree for the identical session
- `mcp/server.py:400, 520, 603`
- `core/session_timeline.py`, `db.py:1223`

Critical Rule 48(b): widening what a shared artifact contains is a change to every blind reader of it. And within `/status` itself, `cost_today` / `tokens_today` are agent-scoped span sums while `total_cost_usd` on the same tile is now a rolled-up session figure — two bases side by side (48(c)).

## Blocking: the inferred schema leaks captured tool output off the machine

Populating `output_schema_inferred` activates a path that has never been live — on `main` nothing writes that field, so the fallback in `SchemaValidator._get_schema` was dead code. Turning it on makes `SchemaValidator.validate` reachable for agents with no declared schema, and `SCHEMA_VIOLATION` alert details embed the offending instance, because that's what jsonschema type-error messages contain:

```
detail: {'errors': ["'wrote /Users/…/secret-notes.md (4211 bytes)' is not of type 'object'"],
         'tool_name': 'write_file'}
```

`alerts._strip_sensitive` only drops the four top-level keys in `SENSITIVE_DETAIL_KEYS`, and `errors` isn't one of them. So with the **default** `alerts.include_captured_content = false`, raw captured tool output still ships to Slack / webhook / Discord / Telegram / ntfy. That's a Critical Rule 5 bypass this PR newly makes reachable.

It's dark for default installs because `capture.tool_outputs` defaults to `false`, but that's a documented supported config, and nothing in the new code checks the toggle.

Three more problems with that path, any one of which would need solving on its own:

- **The schema is tool-blind.** `build_baseline` merges every tool's outputs into one genson schema, and genson sets `required` to the **intersection** of keys seen. A schema inferred from a `search` tool (`required: ["n","results"]`) rejects `{"ok": true}` from a different tool, and rejects any non-JSON string output. That's a guaranteed false positive firing a WARNING alert that carries the raw output. This is Greptile's P1 on `drift.py:87-91`, and the false-positive direction is worse than the false-negative one it describes.
- **It's frozen permanently.** `on_session_end` only builds a baseline when `baseline is None`, and `routes/drift.py:22` says baselines are never recomputed. A schema inferred from one tool's first sessions governs that agent forever, with no rebuild path and no way to inspect or clear it — the field isn't in the `/api/v1/drift` payload or in `tj drift`.
- **It's an undeclared product decision.** This flips schema validation from opt-in (declare a schema) to automatic for every non-interactive agent with `drift.enabled` once capture is on. #617 asked to wire a helper; it didn't authorize a new enforcement default.

## Two docstrings claim more than the code does

Worth fixing regardless of what happens to the rest, because they'd mislead the next reader:

- `core/db.py:1727` — "a strict superset that never under- or over-counts the already-correct paths". It over-counts every session sharing a trace with another marker.
- `routes/status.py` — "Use span truth for a session tile/archive row when spans exist." What it computes is the trace's total attributed in full to each session on that trace, which isn't span truth.

## Suggested path

Split it. Open a PR with just `four_type_token_total` and `_package_uninstall_hint` and I'll merge that quickly — both are clean and both are genuinely wired.

The other two each need a design decision before code: a disjointness rule for trace-keyed cost attribution, and whether schema-enforcement-by-inference should exist at all. I'd rather resolve those in the issue than in review. If you'd like, I'll split #617 into two follow-ups with the analysis above so you're not re-deriving it.

Smaller things, none blocking on their own:

- `_count_archived`'s docstring says it "Mirrors `_build_archive`'s row-selection predicate exactly" — one case diverges: a stored row that's non-zero while its spans roll up to all-zero is dropped by `_build_archive` but counted here.
- `_apply_session_rollup` has no error handling while its neighbour `_build_sdk_services` explicitly degrades rather than 500-ing `/status`. Not asking for a broad `except` (Rule 45 argues against it), just noting the inconsistency.
- `tool_outputs.extend(...)` is unbounded — up to 500 spans × `baseline_sessions` full output objects into genson and then into one JSON column, with no size cap.
- Existing baselines keep `output_schema_inferred = None` and are never rebuilt, so only agents onboarded after this ships get inferred schemas. Two classes of user; worth stating in the body.
- The `@staticmethod` conversion on `infer_schema_from_outputs` is fine but unannounced.
- The drift test sets `attributes[TOOL_OUTPUT]` directly, bypassing `strip_captured_content()`, so nothing pins that the feature stays dark when `capture.tool_outputs = false`.
- Dropping the unused `timedelta` / `DriftBaseline` / `make_session_with_spans` imports is correct but is unannounced drive-by cleanup.
- The `four_type_token_total` call builds a throwaway dict per row when the helper already accepts an attribute-style object.

One process note, not a criticism of the work: the body says the full CI command was attempted locally but exceeded a 420-second window. That's fine, and CI did run the matrix — but green CI wasn't going to catch any of the above, since the double-count needs two markers on one trace and the OOM needs a corpus larger than any fixture. Worth knowing for the next one.

DhruvGarg111:
Review follow-up

Thanks for the detailed review. I addressed the blocking findings while keeping this PR within the safe scope. The two rejected paths remain absent: there is no read-time trace-wide session rollup and no automatic inferred-schema enforcement.

Addressed findings:

- Shared-trace attribution:
  - Resolution now prioritizes an explicit session id, then an existing conversation mapping, then a parent-span chain, and finally a trace only when exactly one explicit session marker is present.
  - Ambiguous trace-only spans remain unattributed instead of being assigned arbitrarily or copied into multiple sessions.
  - Non-explicit ownership is stamped with provenance, including `parent_generated`, so late markers can reparent provisional/trace-derived spans.
  - Reconciliation is transactional and idempotent. Reverse arrival, parent disambiguation, shared traces, and rollback on refresh failure are covered through the real ingest pipeline.

- Duplicate accounting and money boundaries:
  - Ingest-time duplicate suppression remains source-aware.
  - Historical recomputation and drift checks use canonical logical-call selection: cross-source restatements count once, while same-source repeated calls remain distinct.
  - Reconciliation refreshes affected sessions from the canonical observations rather than a raw trace-wide `SUM(cost_usd)`.

- Archive performance and reader parity:
  - Archive filtering remains session-scoped and uses the same zero-signal predicate for the returned rows and `archived_total`; no candidate-by-candidate trace rollup was added.
  - Added the `spans(session_id)` index and an additive migration, with index-repair and backend-parity coverage.
  - Session list/detail/status/archive, CLI status, MCP, and the HTTP backend now preserve the same four session token buckets: input, output, cache-read, and cache-write. Cache-only sessions are retained. The timeline already reads persisted session aggregates and includes all four buckets.

- Schema validation:
  - Validation runs only when captured tool output is present and an explicit agent schema is configured.
  - Capture-disabled output is stripped before validation.
  - The legacy `output_schema_inferred` column remains readable for compatibility but cannot activate validation.
  - The active `genson`/automatic-inference path is not present, and the architecture documentation now describes explicit-only validation.

The first post-push CI run exposed one stale schema-self-heal test helper that did not drop the new session index before altering `spans`. I amended the same follow-up commit, rather than adding another PR commit, and the related schema/index/migration tests now pass (26 selected tests).

Additional local validation passed with Ruff, mypy across 278 source files, compileall, diff checks, and the focused attribution, accounting, storage-parity, API, MCP, and CLI suites. The corrected head is `63f7edcd2014b4cf3cdf91af4e6ec6724314d0f5`; refreshed GitHub checks are currently running.

anilmurty:
Thanks for the update @DhruvGarg111 - reviewing it now. drop a star while you are here :)

anilmurty:
@DhruvGarg111 you fixed everything I raised, and the parts that answer #617 are genuinely good now. I'm asking you to split the PR rather than change more code — most of what's here I want to merge as-is.

Taking the removal path for the rollup and the inference helper was the right call, and a few things deserve specific credit:

- **Every reader now agrees.** I seeded one corpus through `IngestPipeline` and queried all seven — list, detail, `/status` tiles, archive rows, `tj status --json`, MCP `list_active_sessions`, `compute_session_timeline` — and they return identical figures (`1.25`, `11/22/33/44`). On `main` the same script returns `cache=None cw=None` on the list and tile, no cache keys at all from MCP, and no `total_cost_usd` from `tj status --json`. That blocker is fully resolved.
- **You inverted the schema tests instead of deleting them.** `test_uses_baseline_inferred_schema` → `test_ignores_baseline_inferred_schema_without_declaration`, plus a legacy-read test. That is exactly what Critical Rule 23 asks for and almost nobody does it unprompted.
- The removal is clean: zero `genson` hits anywhere under `tokenjam/`, the dependency dropped, tests removed alongside.
- `recompute_session_totals_from_spans` and `session_cost_drift` summing canonical cross-source-deduped observations is a real Critical Rule 37 improvement, independent of everything else.

## The split I'm asking for

**Keep in this PR** (I'll merge it once the two doc items below are done): `four_type_token_total` wiring, `_package_uninstall_hint` wiring, the `session_token_cost_rollup` removal, the `infer_schema_from_outputs` + genson removal, the widened readers across all seven surfaces, the `api_backend` cache fix, and the canonical-observation dedup work.

**Move to its own PR and its own issue:** everything in `63f7edcd`. That commit is a session-attribution redesign — a new persisted span attribute, a new schema migration, a rewritten `_resolve_session`, and a write path that deletes session rows. #617 asked "wire or drop" for four helpers; once two were wired and two were dropped, the issue was answered. Your own "What's NOT in this PR" section still says *"No trace-keyed session-cost rollup. A follow-up needs an explicit disjoint attribution rule"* — that follow-up is what the diff now contains, so the body no longer describes the PR.

I'm not asking for this because the idea is wrong. I'm asking because it needs a design decision I haven't made yet, and review comments are the wrong place to make it.

## Why that commit needs its own review

I reproduced the shared-trace case through the real `IngestPipeline` again. Two $1.00 cost spans on one trace, repricing to $0.012 total:

```
main            w1 = $0.012   w2 = $0.000   sum = $0.012   misattributed, conserved
previous #744   w1 = $0.012   w2 = $0.012   sum = $0.024   double-counted
this revision   w1 = $0.000   w2 = $0.000   sum = $0.000   gone
```

The double-count is genuinely fixed. But the spend now lives on spans with `session_id = NULL` and no session row holds it. In the cost-spans-first order — the common one in async fan-out, where workers report before the marker lands — `reconcile_trace_session_attribution` NULLs the spans and then DELETEs both provisional session rows, so the $0.012 is not reachable from any session at all.

Leaving genuinely ambiguous spend unattributed is a legitimate design choice. Not counting it anywhere *and not saying so* is the part I can't take. Two consequences make it concrete: `_evaluate_session_progress` returns early when `span.session_id is None`, so **budget alerts never fire for spend on a shared trace**; and `_duplicate_observation_sql` filters `session_id IS NOT NULL`, so those rows are also exempt from cross-source dedup. If unattributed is the model, it needs a named bucket on a surface and a `tj doctor` check, not silence.

Four other things in that commit, each of which I'd want to work through properly rather than in a thread:

1. **#326 regression, and nothing pins it.** When a trace has no marker span, `get_marker_session_ids_for_trace` returns `[]` and the code falls through to `new_uuid()`. Measured: 5 trace-only cost spans on one trace → `main` mints 1 session holding $0.030; this revision mints **5 sessions** of $0.006 each. Generic raw-OTLP producers that emit no `gen_ai.invoke_agent` marker get one session per span. `test_trace_id_resolves_to_session_of_sibling_span` still passes because its fixture has a marker.
2. **Quadratic ingest on the path this exists for.** `_session_from_parent_chain` calls `get_trace_spans(span.trace_id)` — hydrating every span on the trace — once per trace-only span with a parent. Measured `get_trace_spans` at 0.059s on a 3,000-span trace, so ingesting that trace costs roughly **178 seconds** of lookup alone, growing O(n²). Live SDK and Claude Code spans short-circuit at step 1, so this lands squarely on fan-out OTLP.
3. **`reconcile_trace_session_attribution` deletes session rows on the ingest hot path.** `DELETE FROM sessions WHERE session_id = $1 AND NOT EXISTS (SELECT 1 FROM spans …)` assumes anything with no spans is an orphan. Nothing checks whether an alert, validation, finding, or transcript row already references that session_id — a provisional session that raised a budget alert leaves the alert dangling. Critical Rule 33(b) is explicit that the completeness of a keep-set has to be argued before a delete relies on it.
4. **Critical Rule 45(b).** The new `except Exception: self.conn.execute("ROLLBACK"); raise` — on a DuckDB `FatalException` the ROLLBACK itself raises, so the rollback's exception propagates and the fatal is never classified. It needs `note_fatal_db_error` / `handle_if_fatal` at the point of recognition. Related: because `reconcile` is called inline from `process()` and re-raises, a transient DB error during reconciliation now fails ingest of the marker span itself.

And one design question rather than a defect: `tokenjam.session_attribution` is stamped into every span's attributes JSON including the no-op `"explicit"` case, so it will render wherever span attributes are shown (Lens trace detail, `tj trace`). Rule 10 is satisfied since it is a constant in `semconv.py`, but an internal write-time marker probably should not sit in the user-visible attribute bag, and certainly not on the ~99% of spans where it carries no information. The read-side compatibility handling is right, for what it's worth — `IS NULL OR IN (...)` treats pre-migration rows as eligible and old DBs are not rewritten.

## Two things to fix in the PR that remains

1. **Drop `docs/plans/2026-09-05-issue-617-maintainer-follow-ups.md`.** `docs/plans/` does not exist on `main`, CLAUDE.md designates `docs/internal/specs/` for code-referenced specs, and `.gitignore` already ignores `.plans/` — the convention here is that plan files stay untracked. Beyond the location: it carries "**For Hermes:** Use the subagent-driven-development skill", hardcodes a Windows venv path that is not this project's documented invocation, and states decisions I have not made ("automatic inference remains removed unless the maintainer explicitly approves", "Recommended: do not ship automatic inference"). It also lists things "PR #744 already removes" that only ever existed in this PR's own earlier revision, so a future reader would go hunting for code that never shipped. If Tracks 0 and 1 are worth keeping, they belong in #617 or a new issue.
2. **Update `.claude/rules/core-architecture.md:94`.** It still says validation runs against "declared or genson-inferred JSON Schema" with a priority list naming `DriftBaseline.output_schema_inferred`. That file is what the next agent loads when touching `schema_validator.py`, so a doc describing deleted behaviour is a live defect under Critical Rule 25 — your plans doc spots this and defers it, but it should land with the removal.

Smaller, non-blocking:

- `docs/architecture.md` says the column "remains only for reading older database files". Nothing on `main` ever wrote `output_schema_inferred`, so no such file exists — "retained for schema stability; always NULL" would be accurate.
- `_package_uninstall_hint` is behaviourally inert: all three call sites pass an explicit `manager` and each returns the byte-identical literal it replaced. That is fine and matches `_package_fresh_install_hint`'s precedent, but note the `manager is None` auto-detection branch now has zero product callers, so #617's "test-only symbol" condition moved one level down rather than going away. Not worth reopening for.
- The new `cache_tokens` / `cache_write_tokens` reach the API, MCP and `--json` surfaces but nothing renders them in Lens or `tj status -v`, which still show only input/output. The labels are honest and API parity is worth having on its own, so I'd take this as a follow-up.
- `scope_sql` is f-string-interpolated into `_duplicate_observation_sql`. Structural rather than user data, so not a Rule 7 violation, but it leaves the helper one careless caller away from being one.
- `recompute_session_totals_from_spans` docstring has a duplicated "the the".
- `test_real_pipeline_leaves_shared_trace_cost_unattributed` currently pins the money-loss as correct behaviour. Whichever way the attribution design lands, that test encodes the decision — worth revisiting deliberately rather than inheriting.

Both Greptile P1s still showing on this PR are stale badges from the 09-04 revision; both are resolved by deletion and I've confirmed neither symbol exists at head. Worth knowing that its current 5/5 was issued over the new attribution model and did not surface any of the items above — the checks and the bot both being green is why I re-ran the reproduction by hand.

Open the split PR whenever suits you and I'll turn it around quickly. If it helps, I'll file the attribution issue with the measurements above so you are not re-deriving them.

DhruvGarg111:
ok oh, i might have misunderstood earlier. i will do this today

anilmurty (approved):
Thanks @DhruvGarg111 — merging. This is the right outcome and you got there faster than I expected.

I verified the split rather than taking it on trust:

- **All seven symbols from `63f7edcd` are gone** — zero grep hits for `session_attribution`, `_session_from_parent_chain`, `get_marker_session_ids_for_trace`, `get_session_ids_for_trace`, `reconcile_trace_session_attribution`, `idx_spans_session_id`. `core/ingest.py` is byte-identical to `origin/main`, so #326's trace-sibling clause is intact.
- **Both regressions I measured are gone.** No-marker trace: 5 trace-only cost spans → 1 session, $5.00 on both this branch and main (the earlier revision minted 5). Shared trace: sum = $8.00 on both, in both arrival orders (the earlier revision lost it to $0.00). The w1/w2 skew that remains is main's pre-existing attribution gap, which is exactly what #749 is for.
- **All seven readers agree**, measured on a seeded corpus. The archive case is the one I'd highlight: main's zero-signal filter drops a cache-only session entirely, and this keeps it.
- Six of seven product hunks are load-bearing — reverting each one reds a specific test.

Two things you did beyond what I asked, both of which I appreciate: you corrected `docs/architecture.md` in three further places, not just the rule file; and the `cost.py` change is comment-only, retracting a "source of truth" claim that the dedup work had quietly made false. Retracting an overreaching comment is the kind of thing that never shows up in a diff review and matters later.

Filing #749 yourself, with the reproductions and the seven decisions stated as open questions, was the best possible response to that review. I'll answer the attribution policy there — it's a product decision and it belongs in the issue, not in a PR thread. #750 will need a rebase onto main now that this has landed; you already called that out.

One follow-up ask, not blocking: `core/api_backend.py`'s cache fix is the single unpinned hunk — reverting it leaves the whole suite green. In a PR whose headline test is `test_session_readers_share_canonical_totals`, `ApiBackend` is a reader that isn't among the readers it pins. Worth adding when convenient.

Both Greptile P1s still visible here are stale badges from the 09-04 revision; both point at code this PR deleted.

PR #745, opened 2026-09-05
Fix #563: reject daemon PIDs when process identity is unavailable

### LINKED ISSUE #563: is_serve_process fails open when neither /proc nor ps is available

be-student:
## Summary

A live PID in `server.state` can belong to an unrelated process after PID reuse. When neither `/proc` nor `ps` could establish identity, `is_serve_process()` returned `True`, allowing PID-file discovery to select that process.

Make the shared identity check return `False` when `ps` is missing or cannot execute, matching the existing fail-closed behavior for unreadable `/proc`. Tests cover missing/unusable probes, matching and unrelated command lines, readable `/proc`, and a persisted stale live PID that must not receive a termination signal through `tj stop`. The CLI reference documents this behavior.

## Related issue

Closes #563

## Validation

- Focused server-state, stop/scoping, upgrade, and lock-holder tests: **70 passed**.
- Full Python unit/synthetic/agents/integration run with four workers: **5,648 passed, 17 skipped, 2 xfailed, 8 failed**. The eight failures were terminal rendering assertions under this runner's `TERM=dumb` / `NO_COLOR=1` environment and narrow output width. All eight passed unchanged when rerun with `NO_COLOR` and `FORCE_COLOR` unset, `TERM=xterm-256color`, and `COLUMNS=200`.
- `ruff check tokenjam/` and `ruff check tests/unit/test_server_state.py`: passed.
- `mypy tokenjam/`: passed (278 source files).
- `git diff --check`: passed.

Implemented and validated with OpenAI Codex. This fixes unavailable-identity handling; it does not add process-start-time tracking or claim to eliminate every PID reuse race.

## Checklist

- [x] Python suites exercised; all eight environment-sensitive failures passed on targeted rerun as detailed above
- [x] Lint clean
- [x] Type check clean
- [x] CLI reference updated; no architecture change requiring CLAUDE.md changes
- [x] No synthetic spans added
- [ ] Maintainer review request after CI is green, as required by CONTRIBUTING.md

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -439,6 +439,10 @@ Metrics: `http://127.0.0.1:7391/metrics`
 
 Stop the background daemon or `tj serve` process.
 
+PID-file discovery requires a matching command line from `/proc` or `ps`. If
+neither can establish the process identity, TokenJam leaves that PID alone;
+a live PID by itself is not proof that it still belongs to the daemon.
+
 ```bash
 tj stop
 ```
--- a/tests/unit/test_server_state.py
+++ b/tests/unit/test_server_state.py
@@ -11,6 +11,8 @@ shape so a regression back to substring matching is caught for real.
 """
 from __future__ import annotations
 
+import pytest
+
 from tokenjam.core.server_state import _looks_like_serve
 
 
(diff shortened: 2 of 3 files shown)


anilmurty:
welcome @be-student - thanks for the contribution. reviewing it now. drop a star while you're here ;)

anilmurty (approved):
Thanks @be-student — merging. Clean, minimal, and scoped exactly to #563: the issue offered two fix directions and you took the fail-closed one without claiming the other.

I verified rather than trusting the green checks:

- **Tests are load-bearing.** Reverting `server_state.py` to main with your tests kept turns 4 red, including `test_stop_does_not_signal_a_live_pid_without_identity` — which is genuinely end-to-end, since it patches `os.kill` on the module object shared with `cmd_stop` and trips the guard on the old code.
- **Your 8 reported failures were environmental, as you said.** I ran the full unit/synthetic/agents/integration suite under `env -u FORCE_COLOR`: 5661 passed, 0 failed. Good instinct re-running before reporting them.
- **No fail-open in the start direction**, which was the thing I most wanted to rule out. `cmd_serve.py` never calls `is_serve_process`, and `_daemon_already_running()` gates on `launchctl`/`systemctl`, so nothing reads `False` as "safe to start a second daemon". The launchd and systemd stop branches also run independently of this check, so a managed daemon on macOS/Linux stays stoppable.

The native-Windows case does become an un-stoppable daemon with no explanation, but #563 calls that platform essentially none of the supported targets and CLAUDE.md documents only launchd and systemd. Fine as the documented tradeoff.

Your hedge in the body is exactly the register we want and I'd like more PRs to read like it: "does not add process-start-time tracking or claim to eliminate every PID reuse race." The docs note is equally good — it states the consequence (`tj stop` leaves it alone), not just the mechanism.

One follow-up I'll file rather than hold you for: `except OSError` also absorbs a transient fork failure (`EAGAIN`/`ENOMEM` under heavy agent fan-out, which is this product's own workload). `cmd_stop` retries so it self-heals, but `tj upgrade` has no retry — a one-shot `ps` failure there prints "nothing to restart" and leaves the old-version daemon running. Narrowing to `(FileNotFoundError, PermissionError, NotADirectoryError)` would keep your fix and let real resource failures stay loud.

Welcome to the codebase.

PR #747, opened 2026-09-06
Fix #746: surface process probe resource failures during upgrade

### LINKED ISSUE #746: is_serve_process: narrow the OSError catch so transient fork failures stay loud
#745 correctly made `is_serve_process()` fail closed when process identity cannot be established, replacing `except FileNotFoundError: return True` with `except OSError: return False`. That fix is right and should stay.

The catch is wider than the fault it was written for. `OSError` also covers a transient `subprocess` fork failure — `EAGAIN` / `ENOMEM` under memory or process-table pressure. That is not a hypothetical for this product: fan-out agent workloads are exactly the thing TokenJam is built to observe, and they are the conditions that exhaust process slots. Before #745 such a failure propagated as a loud crash; now it silently returns `False`, which reads as "that PID is not our daemon".

Impact differs by caller:

- `cmd_stop._find_serve_pid` — self-heals. Its discovery loop retries 30 times at 0.1s, so a one-shot fork failure is absorbed.
- `cmd_doctor._check_daemon_liveness` — informational; reports "not running" for one invocation.
- **`cmd_upgrade.restart_daemon`** — no retry. A single `ps` fork failure during `tj upgrade` prints "tj serve is not running -- nothing to restart" and returns 0. The package upgrades while a live daemon keeps running the old version, and nothing tells the user.

**Fix direction:** narrow the catch to the conditions that actually mean "identity is unavailable" — `(FileNotFoundError, PermissionError, NotADirectoryError)` — so a genuine resource failure propagates instead of being reported as a negative identification. If a broader catch is wanted for robustness, distinguish the two cases and let the resource failure surface, rather than collapsing both into `False`.

Worth pairing with a second look at `cmd_upgrade.restart_daemon`, which has no retry around a probe that is now allowed to fail quietly.

**Done when:** a transient fork failure in the `ps` probe is distinguishable from an unavailable-identity result, `tj upgrade` does not silently skip a daemon restart because of one, and a test covers the resource-failure path separately from the missing-`ps` path.

Surfaced while reviewing #745; not a defect introduced by it, and deliberately not held against that PR.

be-student:
A resource failure starting the `ps` identity probe was treated as a negative identification, allowing `tj upgrade` to exit successfully while leaving the old daemon running. Preserve the distinction between unavailable identity and a failed probe.

## Summary

- Narrow the shared `is_serve_process` catch to missing executable, permission, and non-directory path errors, which still return `False`. Other launch errors propagate.
- Cover `EAGAIN`, `ENOMEM`, and `EMFILE` at the real probe and upgrade command boundaries. Upgrade exits unsuccessfully, never prints "nothing to restart", and does not attempt an unverified restart.
- Preserve the primary database-lock error if its optional process-identity hint fails. Document both contracts in the CLI agent guidance.

## Related issue

Closes #746.

## Tests / Verification

- Before the fix: all six new resource-failure cases failed; 46 existing cases passed.
- Focused server-state, upgrade, stop-scoping, lock-holder and CLI integration tests after the diagnostic follow-up: 175 passed. Three additional lock-diagnostic regression cases failed before that correction.
- Full Python suite before the diagnostic-only follow-up: 5,662 passed, 17 skipped, 2 xfailed (`-n 4 --dist loadscope`; 80 existing warnings).
- `ruff check tokenjam/` and changed tests: passed. `mypy tokenjam/`: passed (278 source files). `git diff --check`: passed.
- Tests exercise the real process-identity and upgrade command logic with controlled OS errors; no live package upgrade, daemon restart, or host resource exhaustion.

## What's NOT in this PR

No automatic retry policy or `/proc` read-error change. Stop/discovery and doctor also propagate resource failures. The diagnostic-only lock-holder hint omits unavailable detail instead, preserving the primary database-lock message without asserting that the daemon is absent. Existing missing-identity fail-closed behavior and PID matching stay intact.

## Checklist

- [x] CLAUDE.md updated (if architecture changed) — scoped CLI rules updated; no architecture change.
- [x] Test spans use `tests/factories.py` (not raw `NormalizedSpan`) — no spans constructed.
- [x] Code-owner review was automatically requested from anilmurty by GitHub. All seven checks passed on the initial commit; CI is rerunning for the diagnostic follow-up.

Implemented and validated with OpenAI Codex.


be-student:
Fixed the diagnostic-only boundary in fc07be27: `_lock_holder_hint` now omits the hint on an OS probe error, preserving the original database-lock ClickException. Operational callers, including upgrade, still propagate the resource failure.

Three new EAGAIN/ENOMEM/EMFILE CLI regressions failed before the correction; all 175 focused and CLI integration tests now pass. Ruff and mypy also pass. The PR description and scoped CLI guidance now distinguish operational propagation from diagnostic containment.

anilmurty (approved):
Thanks @be-student — merging this one, and closing #748 as superseded. Both are yours and both had the right diagnosis; I picked this one on evidence, not size.

Two things decided it:

**The `_lock_holder_hint` containment.** That function's docstring promises it never raises, and it runs while building the "Database is locked" message. Without your `try/except OSError: return ""`, a fork failure there destroys the entire actionable message. I copied your `test_probe_failure_preserves_primary_database_lock_error` into #748's tree to check whether that mattered in practice — it fails there with `assert 'Database is locked' in ''`. The guidance is gone completely. That is a regression #748 introduces, not a deferral, so its narrower scope wasn't the better discipline here.

**Your tests drive the real path.** All 9 new cases go red against `origin/main`. #748's upgrade test patches `upgrade_mod.is_serve_process` directly and asserts `restart_daemon` raises — but `restart_daemon` has no `try/except`, so that assertion holds with or without the fix. It passes against main, so it ships no real coverage for the exact caller #746 was about.

Also worth saying: returning `""` rather than emitting the "no longer a running `tj serve`" sentence is the honest call. A failed probe cannot establish absence, and the comment you wrote says exactly that.

Verified neither PR re-breaks #563 — both catch `FileNotFoundError` explicitly, and I confirmed the missing-`ps` and `NotADirectoryError` cases still return `False` by clearing `PATH` in each worktree. Suite on this branch: 5665 passed, with the only failures the known local-only set that also fails on main.

Two follow-ups I'll file rather than hold you for, both of which apply to the merged code:

1. Greptile's P1 on #748 (`ENOEXEC`, `ELOOP`, `EISDIR` fall outside the narrowed tuple) is valid — and it applies equally here, since this PR uses the identical tuple. Greptile just didn't flag it on this one. Classifying by `errno` rather than exception class is the durable fix.
2. `tj stop` and `tj doctor` now also propagate the resource failure. For `stop` that removes the retry self-healing I credited it with when I filed #746, and for `doctor` a health-check command can now traceback. Related: the upgrade failure surfaces as a raw Python traceback rather than a console message, which is a Rule 35 gap — a `ClickException` naming what could not be determined would finish the job.

PR #748, opened 2026-09-06
Fix #746: keep transient process probe failures visible

### LINKED ISSUE #746: is_serve_process: narrow the OSError catch so transient fork failures stay loud

be-student:
`is_serve_process()` treated every `ps` spawn failure as proof that a PID was not the TokenJam daemon. Under process or memory pressure, `tj upgrade` could therefore skip a live daemon restart and return success.

## Summary

- catch only errors that mean process identity is unavailable
- propagate `EAGAIN`, `ENOMEM`, and other resource failures
- cover the probe boundary and `restart_daemon()` caller

## Fix #746

The fail-closed result remains for missing, inaccessible, or invalid `ps` paths. Resource failures now stay loud because they do not establish process identity.

## Tests / Verification

- `57 passed` across server-state, stop-scoping, lock-holder, and upgrade tests
- `ruff check` passed
- `mypy tokenjam/` passed

## What's NOT in this PR

- no retry policy change; the issue only requires preserving the existing error
- no daemon discovery or restart behavior change outside error classification

Closes #746

@anilmurty ready for review.

🤖 Generated with [Codex](https://openai.com/codex/)

Co-Authored-By: anilmurty <noreply@github.com>

### DIFF
--- a/.claude/rules/cli-output.md
+++ b/.claude/rules/cli-output.md
@@ -86,6 +86,12 @@ explicitly. (Critical Rule 8 — never construct a `NormalizedSpan` directly in 
 
 ### Daemon (launchd / systemd)
 
+`is_serve_process()` fails closed only when process identity is unavailable
+(`ps` is missing, inaccessible, or has an invalid path). Resource exhaustion
+while spawning `ps` must propagate; treating `EAGAIN` or `ENOMEM` as a negative
+identity check can make `tj upgrade` silently leave a live daemon on the old
+version.
+
 `tj onboard` (and `tj onboard --claude-code` / `--codex`) installs a background daemon that runs `tj serve` on login:
 - **macOS**: `~/Library/LaunchAgents/com.tokenjam.serve.plist` — loaded via `launchctl load`. Logs at `/tmp/tj-serve.{out,err}`.
 - **Linux**: `~/.config/systemd/user/tokenjam.service` — enabled via `systemctl --user enable --now tokenjam`.
--- a/tests/unit/test_cmd_upgrade.py
+++ b/tests/unit/test_cmd_upgrade.py
@@ -131,6 +131,14 @@ class TestRestartDaemon:
         assert method == "none"
         assert success is True
 
+    def test_identity_probe_resource_failure_is_not_reported_as_no_daemon(self):
+        state = ServerState(pid=123, port=7391, config_path=None)
+        failure = OSError("process table exhausted")
+        with patch.object(upgrade_mod, "is_pid_alive", return_value=True), \
+             patch.object(upgrade_mod, "is_serve_process", side_effect=failure):
+            with pytest.raises(OSError, match="process table exhausted"):
+                upgrade_mod.restart_daemon(state)
+
     def test_uses_launchd_kickstart_when_launchd_supervises_it(self):
         state = ServerState(pid=123, port=7391, config_path=None)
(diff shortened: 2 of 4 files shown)


anilmurty:
Closing this as superseded by #747, which is now merged — both are yours, and this one had the same correct diagnosis. I picked #747 on two pieces of evidence rather than on scope:

1. **`_lock_holder_hint`.** Its docstring promises it never raises, and it runs while building the "Database is locked" message. I copied #747's `test_probe_failure_preserves_primary_database_lock_error` into this branch to test whether the difference mattered: it fails here with `assert 'Database is locked' in ''` — a fork failure at that point wipes out the whole actionable message. That's a regression this branch introduces rather than something deferred, which is why the narrower scope wasn't the safer choice this time.

2. **`test_identity_probe_resource_failure_is_not_reported_as_no_daemon` passes against `origin/main`.** It patches `upgrade_mod.is_serve_process` with a `side_effect` and asserts `restart_daemon` raises — but `restart_daemon` has no `try/except`, so that holds with or without the fix. Worth internalising as a general check: when you add a test for a fix, revert the product change and confirm it actually goes red. Two of your three cases here were load-bearing; that one wasn't.

Your in-code comment on the narrowing is better written than the one that merged, and Greptile's P1 on this PR (`ENOEXEC`, `ELOOP`, `EISDIR` sit outside the narrowed tuple) is a real finding that applies to the merged code too — Greptile simply didn't raise it on #747. I'm filing it as a follow-up and you're welcome to take it; classifying by `errno` instead of exception class is the durable version of both PRs.

Nothing wasted here — thanks for the second angle on it.

Issue #565, opened 2026-07-25
call_count blind spots: parity _proj_cost omits it; MCP get_cost_summary omits it and mishandles group_by=tool


anshss:
**Severity:** Low
**Model:** sonnet

**Severity:** Low
**Where:** tokenjam (tests/integration/test_storage_backend_parity.py ~174 _proj_cost; tokenjam/mcp/server.py ~574 _tool_get_cost_summary)
**Observed:** PR #482 adds CostRow.call_count and a working group_by="tool" path in DuckDBBackend/CLI/API, but (a) the parity suite's _proj_cost projection tuple omits call_count, so a future ApiBackend-shim serialization drop of the field would pass parity (exactly the class of bug that hit cache_tokens before, per the shim's own comment); (b) the MCP tool _tool_get_cost_summary omits call_count (and cache token fields) from its output dict and its docstring only advertises day/agent/model, yet nothing rejects group_by="tool" — that call returns all-zero rows with no useful data.
**Fix direction:** Add call_count (and cache token fields) to _proj_cost; in MCP either serialize the full row shape incl. call_count and support tool grouping, or validate/reject unsupported group_by values explicitly.
**Done when:** Parity test fails if any backend drops call_count; MCP get_cost_summary either supports tool grouping with call_count or errors clearly on it.

be-student:
I verified this remains open on current `main`: `_proj_cost` still omits `call_count`, while the MCP serializer still drops `call_count`, `cache_tokens`, and `cache_write_tokens`. The issue is unassigned, GitHub shows no linked development PR, and the fetched remote branches contain no competing #565 work.

I’ll add source-coupled regressions for the parity projection and MCP row shape, then update the serializer and tool contract so `group_by="tool"` returns the only meaningful metric for tool spans. I’ll keep this as one bounded PR and run the focused MCP/parity tests plus the normal quality gates.

PR #750, opened 2026-09-06
session attribution for shared OTLP traces (#749)

### LINKED ISSUE #749: Design: session attribution for shared OTLP traces
## Problem

A shared OTLP trace can carry multiple explicit `invoke_agent` session markers while cost spans carry only the trace ID. A read-time trace-wide rollup double-counts by assigning the whole trace to every session. The current ingest-time redesign avoids that duplication, but its ambiguous-spend behavior is not a settled product decision.

## Reproduced shapes

With two $1 cost spans on one trace:

- `main` conserves $2 but assigns it to the first marker.
- The prior read-time rollup revision reported $2 for each marker, for $4 total.
- The attribution-reconciliation revision leaves shared trace-only spans unassigned after marker reconciliation, so neither session exposes the spend.

Reverse arrival is important: worker cost spans can precede the marker. Any design must be exercised through `IngestPipeline`, not only raw database fixtures.

## Decisions required before implementation

1. Define the ownership rule for trace-only spans on a shared trace: only exactly one explicit marker, parent-span ownership, a named unattributed bucket, or another explicitly documented model.
2. If ambiguous spend remains unattributed, expose and audit it rather than silently dropping it from session totals and budget alerts.
3. Preserve the existing no-marker trace behavior: a trace with five trace-only cost spans must not mint five separate sessions.
4. Define reference safety before deleting any provisional session row: alerts, validations, findings, transcripts, and other session references must remain valid.
5. Keep parent resolution bounded; loading an entire trace for every child is quadratic on large fan-out traces.
6. Classify fatal DuckDB failures at the reconciliation boundary and decide whether a reconciliation failure may reject the marker span.
7. Decide where internal attribution provenance belongs. It should not add a no-op user-visible attribute to every span without an explicit product decision.

## Acceptance criteria

- A real-pipeline shared-trace test proves the sum across session buckets equals the real spend and each observation has at most one owner.
- Reversed arrival is idempotent and has an explicit, observable outcome.
- A no-marker trace keeps one session per trace (the existing #326 behavior).
- The implementation uses bounded/grouped queries and includes a scale-shaped regression or benchmark for large traces.
- No reconciliation path creates dangling references or masks a fatal database error.
- Every affected reader and budget/doctor surface either preserves the selected ownership model or documents its intentional scope.

This issue is intentionally design-gated; it does not choose an attribution policy in advance.

DhruvGarg111:
Implements the attribution policy decided in #749, so that spend on a shared OTLP trace has exactly one owner — and where it genuinely cannot, is named rather than dropped.

Closes #749

## Background

A shared trace can carry multiple `invoke_agent` session markers while cost spans carry only the trace id. Three behaviours have now been measured on the same two-dollar fixture:

| | result |
|---|---|
| `main` before this PR | $2 conserved, but assigned entirely to the first marker |
| read-time rollup (rejected) | $2 reported for each marker, $4 total |
| unowned-after-reconcile (rejected) | $0 — no session row held the spend |

Double-counting and silent loss are both wrong; the second is worse, because nothing on any surface contradicts it.

## Summary

**Ownership ladder** (#749 decision 1) — first match wins, and the step taken is recorded:

1. **Parent-span ownership** — `_session_from_parent_chain` resolves `parent_span_id` to a span that already carries a `session_id`.
2. **Sole-marker ownership** — `get_marker_session_ids_for_trace` returns exactly one explicit marker for the trace.
3. **Unattributed** — more than one marker and no resolvable parentage. Not split, not assigned to the first marker, not dropped.

The resolved step is stored on `NormalizedSpan.attribution_step` (`step1_parent` / `step2_marker` / `step3_unattributed` / `provisional`, alongside the existing `explicit` and `conversation` paths).

**Unattributed bucket** (decision 2) — `get_unattributed_spend` / `get_unattributed_spend_summary` make it a first-class query rather than something every reader has to remember to look for as `session_id IS NULL`. Surfaced through `tj cost`, `tj status`, `tj doctor`, `GET /api/v1/cost` and `GET /api/v1/status`. `CostFilters.group_by` gains `session`.

**Reconciliation without deletes** (decision 4) — `reconcile_trace_session_attribution` re-parents provisional spans when a marker arrives and marks the superseded session rows **superseded rather than deleting them**, so nothing holding a `session_id` (alerts, validations, findings, transcripts) is left dangling.

**Bounded parent resolution** (decision 5) — a bounded cache (`_PARENT_CACHE_MAXSIZE = 20_000`) replaces the per-child full-trace hydration that made an earlier revision O(n²) — roughly 178s to ingest a 3,000-span trace.

**Provenance stays internal** (decision 7) — `attribution_step` is a field on the span record, explicitly kept out of the `attributes` JSON. `tokenjam/otel/semconv.py` is untouched, so no new user-visible span attribute is stamped onto the ~99% of spans where it would carry no information.

## Preserved behaviour

**#326 is intact** (decision 3): a trace with no marker still mints **one** session, not one per span. This regressed in an earlier revision (five trace-only spans produced five sessions) and is the case most likely to break silently, since generic OTLP producers emit no `gen_ai.invoke_agent` span at all.

## Tests / Verification

CI green on `4f5cd39b`: `lint`, `test (3.10)`, `test (3.11)`, `test (3.12)`, `test-ts`, `version-lockstep`.

~1,200 lines of tests, the bulk in `tests/integration/test_full_pipeline.py` and `tests/synthetic/test_ingest.py`, covering #749's acceptance criteria: a real-pipeline shared-trace test where the sum across buckets equals real spend with each observation having at most one owner; both arrival orders, including cost-spans-before-marker (the common async fan-out case) with reconciliation idempotent; the no-marker one-session-per-trace case; and `tests/unit/test_doctor_unattributed_spend.py` for the doctor surface.

Detailed maintainer verification is in the review thread rather than restated here.

## What's NOT in this PR

- **Per-session budget alerts on unattributed spend.** Per decision 2, spend belonging to no session does not trip a per-session budget; it must still count toward any figure claiming to be a total. No new alert type is introduced.
- **Fractional attribution.** Step 3 deliberately does not split ambiguous spend across the markers on a trace — a fraction is a number we would have to defend and cannot.
- **Retro-attribution of history.** Spans already ingested keep whatever ownership they were written with; this changes the write path forward, not existing rows.

---

*PR description rewritten by the maintainer at review time: the original body still described this as a design-gated draft that deliberately did not choose an attribution policy, which stopped being true once #749 was decided and the policy implemented. The code is entirely @DhruvGarg111's.*


anilmurty:
The #749 decisions are posted — https://github.com/Metabuilder-Labs/tokenjam/issues/749#issuecomment-5622466400 — so this is unblocked whenever you want to pick it up.

Short version of what changed for this branch:

- **Ownership is a three-step ladder**: parent-span → sole-marker → unattributed. First match wins, and the step taken gets recorded.
- **Unattributed spend is a named, disclosed bucket**, never silent loss. The invariant to test is that every session bucket plus the unattributed bucket sums to real spend.
- **#326 stays**: a no-marker trace mints one session, pinned by a test that fails at five.
- **Prefer not deleting provisional sessions at all** — re-parent or mark superseded. A delete needs an enumerated keep-set, not "no spans means orphan".
- **Bounded resolution**, with a scale-shaped regression. No per-child full-trace hydration.
- **Classify fatals at the point of recognition**, and a reconciliation failure must not reject the marker span — ingest is the thing we cannot lose.
- **Provenance stays out of the span attribute bag.** Session record or dedicated column, not `tokenjam.session_attribution` on every span.

One sequencing note that may save you work: steps 1 and 2 of the ladder alone strictly improve on `main` without needing the bucket, since both produce a single unambiguous owner. If it is easier to review, land those first and do step 3 plus the disclosure surfaces as a second PR. The one combination to avoid is step 3's "leave it unowned" without the surfacing — that is the silent loss the issue exists to prevent.

This branch is `DIRTY` now that #744 has merged, so it needs a rebase onto `main` before CI means anything. You already flagged that in the body.

No rush on my end, and thanks again for splitting it rather than pushing through.

DhruvGarg111:
﻿Addressed Greptile's review on generic OTLP session markers:

### Changes
1. **Generic OTLP Marker Recognition (`tokenjam/core/ingest.py`)**:
   - In `_is_session_marker(span)`, removed the restriction to `span.name == GenAIAttributes.SPAN_INVOKE_AGENT`. Any span carrying an explicit or conversation-derived session identity (`span.session_id is not None and span.attribution_step in {"explicit", "conversation"}`) is now recognized as a session marker for trace reconciliation, regardless of whether it was named `invoke_agent` or a producer-defined name (e.g. `workflow`, `agent.run`).

2. **DuckDB Marker Query (`tokenjam/core/db.py`)**:
   - In `get_marker_session_ids_for_trace(trace_id)`, removed `AND name = $2` on `GenAIAttributes.SPAN_INVOKE_AGENT` so the query selects all explicit session IDs on the trace:
     `SELECT DISTINCT session_id FROM spans WHERE trace_id = $1 AND session_id IS NOT NULL AND (attribution_step IS NULL OR attribution_step IN ('explicit', 'conversation')) ORDER BY session_id`.

3. **Backend Test Stub Parity (`tests/synthetic/test_ingest.py`)**:
   - In `_StubBackend.get_marker_session_ids_for_trace`, removed `and s.name == "invoke_agent"` to preserve parity with the DuckDB backend query.

4. **Test Coverage**:
   - Added synthetic tests (`tests/synthetic/test_ingest.py`) and full-pipeline DuckDB integration tests (`tests/integration/test_full_pipeline.py`) covering generic OTLP markers with custom names:
     - **Sole marker**: child spans on a trace with a single custom marker (e.g. `workflow`) resolve to it via `step2_marker`.
     - **Multi-marker**: unparented spans on traces with multiple custom markers (e.g. `workflow`, `agent.run`) resolve to `step3_unattributed`.
     - **Reverse arrival**: late arrival of a custom-named marker triggers reconciliation, reparents provisional child spans to `step2_marker`, and supersedes/zeroes the provisional session.

### Verification
- `ruff check tokenjam` — passed.
- `mypy tokenjam` — passed (278 source files).
- `pytest tests/synthetic/test_ingest.py tests/integration/test_full_pipeline.py` — passed (74 tests).
- `pytest tests/integration/test_storage_backend_parity.py` — passed (22 tests).
- `pytest tests/unit/test_db_helpers.py tests/unit/test_schema_self_heal.py tests/unit/test_doctor_unattributed_spend.py tests/integration/test_cost_attribution.py` — passed (43 tests).

DhruvGarg111:
Addressed review feedback regarding generic OTLP session markers and child span resolution:

1. **Generic OTLP Markers (`tokenjam/core/ingest.py`, `tokenjam/core/db.py`)**:
   - Broadened `_is_session_marker` and `get_marker_session_ids_for_trace` to recognize any span bearing an explicit session identity (`session.id`), decoupling marker detection from the span name `invoke_agent` (supporting `workflow`, `agent.run`, etc.).
   - Fixed nested marker resolution (`_session_from_parent_chain`) to stop at the nearest explicit/resolved marker ancestor.

2. **Ingestion & Cache Efficiency (`tokenjam/core/ingest.py`)**:
   - Prioritized Step 1 (Parent-span ownership) over conversation resolution in `_resolve_session`, preventing child spans that inherit `conversation_id` from triggering redundant full-trace reconciliations.
   - Preserved current marker spans in `_parent_cache` across reconciliation to maintain O(1) in-memory resolution for subsequent child spans.

3. **Test Coverage**:
   - Added synthetic and full-pipeline DuckDB tests covering custom-named markers (sole-marker, multi-marker unattributed routing, reverse arrival, nested hierarchies, and inherited conversation continuity).
   - All linters, type checks, and test suites pass.

anilmurty:
@DhruvGarg111 this is close, and the hard parts are right. Three reviewers went through it against the seven decisions on #749, all verifying by execution rather than reading. Five things to fix; none of them is the design.

## What's verified correct

**The money question is settled.** Measured through the real `IngestPipeline` on real DuckDB, both arrival orders:

```
markers-first     w1 $0.00  w2 $0.00  unattributed $2.00  total $2.00
cost-spans-first  w1 $0.00  w2 $0.00  unattributed $2.00  total $2.00
```

Order-independent, and the sum invariant holds on every surface: `tj cost`, `tj status`, `tj doctor`, both API endpoints and a raw `SELECT SUM(cost_usd)` all report 5.50 on a mixed corpus. A third shape neither of us named — marker M1 takes $2 via `step2_marker`, then a second marker M2 arrives — also reconciles to $0/$0/$2 with no double count.

**#326 is preserved and pinned at exactly five.** Deleting the no-marker path produces `assert 5 == 1`. Both the synthetic and real-DuckDB versions go red.

**Decision 4 done properly.** No `DELETE` is added anywhere; the only one in the tree is the pre-existing retention purge. Provisional rows become `status='superseded'` with totals zeroed, guarded by `NOT EXISTS`, and you taught four readers to skip them. References stay valid.

**Decision 6 is right on both halves.** `is_fatal_db_error` / `note_fatal_db_error` fire *before* the ROLLBACK, and the ROLLBACK has its own `try/except` so a fatal rollback can't mask the classification. Reconciliation cannot reject the marker span — `insert_span` is durable at step 4, the call site doesn't re-raise, and `process()` continues. That was the subtlest of the seven and you got it.

**Decision 5 is real.** Linear sweep: 500 → 2.11s, 1k → 4.04s, 2k → 7.84s, 4k → 16.07s. Flat ~4ms/span, against the ~178s/3,000-span O(n²) it replaces.

**Decision 7 satisfied.** `attribution_step` is a typed column, never in `span.attributes`, and `_span_to_dict`'s allowlist excludes it so it can't surface in Lens or `tj trace`.

## Blocking

### 1. The ladder runs ahead of the conversation lookup, and it moves real money

`_resolve_session` evaluates the parent-chain step *above* the `conversation_id` lookup. So a span carrying its own explicit `conversation_id` that resolves to a known session is now charged to its **parent's** session instead. Measured through the real pipeline with session C existing for `conv-C`, and a cost span carrying `conversation_id=conv-C` whose parent sits in session P:

```
PR head      →  P / step1_parent
origin/main  →  C
```

Decision 1 scopes the ladder to *trace-only* spans. A span carrying a conversation marker is not trace-only — it is explicitly identified, and this silently moves its money onto an inferred parent. That is the same misattribution class the two rejected revisions died on, arriving by a different door.

It is also unpinned in both directions: `test_child_spans_inheriting_conversation_id_resolve_via_step1_parent` only covers the case where parent and conversation resolve to the *same* session, so nothing fails when they diverge.

Restore the order to `session_id → conversation_id → parent → sole-marker → unattributed`, and add a test for the divergent case.

### 2. The test that pinned the money loss last time still passes when the money is lost

`test_real_pipeline_leaves_shared_trace_cost_unattributed` carries the same name as the test that encoded the $2 → $0 loss as correct in the previous revision. It was rewritten, but it still asserts only the negative half:

```python
assert cost_rows[0].session_id is None
assert full_stack.db.get_session("w1").input_tokens == 0
assert full_stack.db.get_session("w2").input_tokens == 0
```

Forcing `get_unattributed_spend_summary` to return zeros — a full silent-loss simulation — reds **ten** other tests and leaves this one green. The same gap is in `test_shared_trace_does_not_choose_an_arbitrary_session` and `test_generic_otlp_multi_marker_with_custom_names_resolves_to_unattributed`.

One line each: also assert the spend appears in the bucket. The rest of your coverage genuinely catches both historical failure modes (4 red for misattribution, 10 for silent loss), which is why this one omission stands out.

### 3. The disclosure states a cause it cannot establish

Every user-facing string says the spend is unattributed because of "shared traces without unambiguous parentage", but `get_unattributed_spend_summary` selects on `session_id IS NULL` alone — not on `attribution_step = 'step3_unattributed'`.

The langfuse and helicone backfill adapters write spans with `session_id=None` straight through `db.insert_span`, bypassing the pipeline. Seeding four such spans on four **distinct** traces, with no shared trace and no ambiguity anywhere in the corpus:

```
i  Unattributed spend: $28.00 across 4 span(s) on 4 trace(s) is in the unattributed
   bucket (shared traces without unambiguous parentage).
```

A stated reason we did not observe, on the surface whose whole job is honesty about what we could not determine. Fix either end — scope the query to `attribution_step = 'step3_unattributed'`, or drop the causal clause to what is actually known.

Worth deciding deliberately rather than incidentally: those backfill spans are also unattributed money that nothing currently names. If you scope the query, that population goes back to being invisible. I'd take that for now and file it, but say which you chose.

### 4. Lens never renders the bucket

`routes/status.py` adds `unattributed_spend` to the payload and nothing reads it. The Lens SPA is the only consumer of `/api/v1/status`, and `grep -rn unattributed tokenjam/ui/index.html` finds only the pre-existing tenant-concentration footnote.

So a Lens user sees a correct grand total and a Sessions view whose rows sum to less, with nothing explaining the gap. Decision 2 names `/status` explicitly — the endpoint discloses, the surface a user actually looks at does not. That is Critical Rule 24's inverse check: the field resolves, nothing links to it.

### 5. The scale regression does not cover the path #749 is about

`test_scale_deep_parent_chain_ingest_is_linear` asserts **no time bound at all** — it stays green with the O(n²) per-child hydration restored, so its stated purpose ("without O(n^2) blowup") is not what it tests.

`test_real_pipeline_reconciliation_scale_benchmark` does assert `< 2.0s`, but it bypasses `IngestPipeline` entirely (raw `bulk_insert_spans` + a direct reconcile call, despite the `test_real_pipeline_` prefix) and has ~18x headroom — replacing the batched UPDATE with 1,000 per-span UPDATEs still passes at 0.55s.

Either put a bound on the linear test or add a timed ingest-through-pipeline case on real DuckDB. As it stands the criterion is met by name.

## Worth fixing while you're in there

- **The synthetic reconciliation tests verify test-owned code.** `_StubBackend` in `tests/synthetic/test_ingest.py` contains a ~50-line hand-rolled clone of the ladder including `resolve_chain` and the 128-depth cap. No-op'ing the *real* DuckDB `reconcile_trace_session_attribution` leaves all nine green; the only real coverage is the six in `test_full_pipeline.py`. Consider using `InMemoryBackend` for the reconciliation cases and keeping the stub for resolution-step cases.
- **The parity dollar check is vacuous.** The seed's only session-less spans are `make_tool_span` rows with `cost_usd=None`, so `get_unattributed_spend` parity compares `0.0 == 0.0`. Add one costed `make_llm_span(session_id=None, cost_usd=…)`.
- **The idempotency test is a self-comparison** with no expected values — it survives a no-op reconcile. Assert the expected snapshot before comparing repeats.
- `test_real_pipeline_session_attribution_invariant` uses `WHERE status != 'superseded'`, which under DuckDB three-valued logic also drops NULL-status rows. Product code uses the correct `(status IS NULL OR status != 'superseded')` — mirror it.

## Smaller

- `get_unattributed_spend_summary` returns `cost_usd` and `spend_usd` holding the same number. One figure under two names at a money boundary invites a surface to add them.
- That summary uses a raw `SUM(cost_usd)` while session totals use the duplicate-observation winner rule (Critical Rule 37). Safe today since backfill spans carry a `session_id` — worth a comment saying why.
- `_parent_cache` is documented as LRU but `.get()` never calls `move_to_end`, so eviction is FIFO.
- `resolve_chain` caches the depth-128 bail-out result, so a later shorter path to the same node can read a wrongly-cached `None`.
- `get_session_by_conversation` doesn't exclude superseded rows, so one can be reselected and start re-accumulating. Benign but undocumented.
- `api_backend._dict_to_span` omits `attribution_step` — a shim divergence of exactly the kind #565/#751 was about. Harmless today (ingest never runs on `ApiBackend`).
- `reconcile_trace_session_attribution` SELECTs every span on the trace once per marker span — ~n^1.24 on a marker-per-span trace (9.09s → 21.33s → 50.44s at 1k/2k/4k pairs). Bounded, not grouped.
- Neither note gives the user anywhere to go, while the plan-tier note beside it names its next step. `tj cost --group-by session` is the first-class query you built and it works — point at it.
- "shared traces without unambiguous parentage" is engineering vocabulary; decision 2 asked for the register of our other caveats.
- Two drive-by edits unrelated to the feature: `api/app.py`'s `read_text(encoding="utf-8")`, and a comment rewrite at `routes/cost.py:973` that no longer parses ("which is practical proof that a missing table row stays invisible"). Revert the comment.
- Typo introduced at `db.py:3149`: "leave the **the** row holding".
- Migration 2's historical SQL text was edited in place. Harmless here, but not a habit to form.

## Housekeeping

I rewrote the PR description — the original still described this as a design-gated draft that deliberately did not choose a policy, which stopped being true. The code is entirely yours. Note that my description says the bucket holds ambiguous-trace spend, which is the intent but not what the query currently does — whichever way you resolve blocker 3, tell me and I'll make the description match.

Suite is healthy: +44 tests, +4.2s CI, no new failures. Greptile is 5/5 with no unresolved findings and caught none of the five above — all of them needed the code run rather than read.

DhruvGarg111:
Addressed the five blockers from the review on commit `9112c31`:

- Attribution now follows `session_id -> known conversation_id -> parent chain -> sole marker -> unattributed`.
- Shared-trace tests assert the nonzero `cost_usd` retained in the unattributed bucket, including the divergent conversation/parent case.
- The bucket is documented as all spend not assigned to a named session; the canonical field is `cost_usd`. API, CLI, status, doctor, and Lens surfaces disclose the amount and point to `tj cost --group-by session`.
- Lens now shows unattributed spend even when there are no named session cards.
- The scale regression now times a real `IngestPipeline` marker ingest with DuckDB and verifies 1,000 spans under the bound.

Also covered the smaller findings: parent-cache LRU promotion, superseded conversation lookup, non-cached depth-cap misses, API attribution-step round-tripping, nonzero backend parity, explicit idempotency expectations, and NULL-safe status predicates. Historical migration SQL was not changed.

Local validation after the final commit:

- 240 affected integration/synthetic/unit/Lens tests passed in 63.56s.
- 3 changed CLI tests passed.
- Ruff passed.
- Mypy passed for 278 source files.
- Python bytecode compilation passed.
- `git diff --check` passed.

The branch has been pushed; remote head is `9112c318840f2be12645500a66297f558b77f0f3`. The broad local run also exposed pre-existing Windows-only/platform-state failures in untouched tests and the async-hook concurrency module; those are reported separately from the changed-path verification above.

DhruvGarg111:
I initially had problems in this implementation, but i believe its complete now. 
let me know if any change is required.

anilmurty (changes requested):
@DhruvGarg111 thanks for turning the five blockers around cleanly. I re-ran the ladder through the real pipeline on DuckDB in both arrival orders, with parent chains, with the repeated-marker case Greptile raised, and with the five-span no-marker trace, and every shape conserves spend with exactly one owner or the named bucket. The API and CLI surfaces agree with a raw sum. That part is done.

Three things before I can merge:

1. Rebase onto main. Migrations 23 and 24 now collide with the ledger work that landed this week (main's 23 adds session repo context, 24 adds `bridge_session_id` and the ledger tables). Renumber yours to 25 (`idx_spans_session_id`) and 26 (`attribution_step`), update the `# migration` comment on the `EXPECTED_ADDITIVE_COLUMNS` entry, and keep both sides of the small conflict in `test_storage_backend_parity.py` (`get_unattributed_spend` plus main's `get_session_commits` / `get_session`). I did that resolution locally and the suite is green post-merge, so it is mechanical.

2. The Lens panel links to `#/cost?group_by=session`, but `CostView` rejects `session` and silently falls back to Total, so the link labelled `tj cost --group-by session` lands on a page with no session breakdown. Either teach the Spend view the `session` group (validator, `COST_GROUP_BY_LABELS`, dropdown) or make the panel plain text that names the CLI command. Pin whichever you pick.

3. The "reconcile every new marker span" change makes every span with an explicit `session_id` run `reconcile_trace_session_attribution`, which reads the whole trace under the write lock and then finds nothing to do. Claude Code emits one trace per session with every span explicit, and measured through the pipeline that path went from about 8.5 ms/span on main to 15 to 18 ms/span here (500 to 8,000 spans on one trace). Gate reconciliation on whether the trace has anything reconcilable: the span id is new AND (its session id is new to the trace OR the trace has a span with `session_id IS NULL` or a non-explicit `attribution_step`). That keeps the repeated-marker case you fixed and reduces the steady state to one indexed probe. Add a timed test that ingests a few thousand explicit-session spans on one trace through `IngestPipeline` and bounds ms/span against the trace-only path; the existing benchmark only times a single marker.

Smaller, no need to reply on each: extend `test_advertised_commands_are_invocable.py` with the new `tj cost --group-by session` string; migration 2's DROP INDEX text is still edited in place (your note said it was not); the `routes/cost.py` comment rewrite ("practical proof that a missing table row stays invisible") should go back to the original; the `read_text(encoding=...)` change in `app.py` is unrelated; and one sentence somewhere in `docs/architecture.md` saying the optimize analyzers do not see unattributed spend would close the last acceptance criterion. I will update the PR description to say the bucket is all session-less spend, matching your resolution, and drop "Draft" from the title when you push.

PR #751, opened 2026-09-06
Fix #565: expose complete MCP cost rows

### LINKED ISSUE #565: call_count blind spots: parity _proj_cost omits it; MCP get_cost_summary omits it and mishandles group_by=tool

be-student:
Closes #565

MCP `get_cost_summary` returned cost rows without `cache_tokens`, `cache_write_tokens`, or `call_count`. This made ordinary MCP totals incomplete and made `group_by="tool"` effectively useless: tool-call spans correctly carry zero cost and tokens, so their call count is the meaningful result.

This change serializes the complete `CostRow` shape, documents tool grouping in the MCP tool contract, and adds regressions for both complete row serialization and tool counts. The storage-backend parity projection now includes `call_count`, so a future REST/API shim omission fails the existing cross-backend gate.

Validation:

- 2 new MCP regressions failed before the implementation and pass afterward
- 104 MCP and storage-backend parity tests pass
- Ruff passes for the changed production module; changed test files pass with the repository's pre-existing `E402`/`F811` exceptions ignored
- mypy passes for the MCP server and related cost-row boundaries
- `git diff --check` passes

I also started the entire non-e2e suite. It reached 44% without a reported failure before I stopped it after an existing CPU-heavy section ran for several minutes; the focused changed surfaces completed independently as listed above.


anilmurty:
@be-student thanks — the parity half of this is exactly right, and the docstring you wrote for the tool rows is the kind of thing I wish more of our surfaces had. One blocker before it can land.

**First, the red CI is not yours.** `test_relearn_proposals_carries_persona_when_never_run` failed on 3.11 with `assert 'computing' == 'never_run'`. That is a pre-existing flake in `tests/integration/test_relearn_api.py`: the autouse fixture that exists specifically to prevent it clears `report_store._COMPUTING`, but the route reads `relearn_store._COMPUTING`, and those are two different Event objects. Your PR touches nothing in that path. I'm fixing it separately — don't chase it.

## Blocking: the MCP cost tool breaks in HTTP proxy mode

The new serializer reads `r.cache_tokens`, `r.cache_write_tokens` and `r.call_count` unconditionally, but `mcp/server.py:126` `_HttpDB.get_cost_summary` builds a `SimpleNamespace` carrying only `group, agent_id, model, input_tokens, output_tokens, cost_usd`. Reproduced against a seeded corpus:

```
{'error': "'types.SimpleNamespace' object has no attribute 'cache_tokens'"}
```

This isn't a corner case. Per `.claude/rules/api.md` — the file this PR edits — HTTP proxy is mode 1, selected whenever `tj serve` is reachable, so it's the normal path for most users. Every non-empty cost summary returns an error instead of data.

The fix is three lines: forward `cache_tokens`, `cache_write_tokens` and `call_count` from the REST response, all of which `/api/v1/cost` already returns at `cost.py:944-946`.

## The more useful finding underneath it

Your tests only exercise `InMemoryBackend`, which is why this got past them — and that points at something worth fixing properly. #565's parity half exists so that a shim dropping a field fails CI. But `_HttpDB` is a **second, independent HTTP shim** that the parity gate doesn't cover, and it's the one that just broke.

So please also add `_HttpDB` to a parity gate, or at minimum a test asserting `_HttpDB.get_cost_summary` yields objects carrying every `CostRow` field name. Otherwise the exact class of bug #565 was filed to prevent stays live in the shim nobody is checking. Three forwarded fields fix today's break; the gate is what stops the next one.

## Verified, for the record

The numbers are right. On a seeded corpus (3 LLM spans at 1000/200/5000/800 tokens and $0.05 each, plus Read×2 / Write×1 tool spans), MCP output matches the backend `CostRow`s exactly for every grouping — `day` gives `(3000, 600, 15000, 2400, $0.15, 3)`, `model` identical, `tool` gives `Read → call_count 2`, `Write → call_count 1`. The row dict is now a character-for-character match of the REST route's, so the two surfaces are scoped alike. All three of your tests are load-bearing: reverting `mcp/server.py` fails both new serialization tests with `KeyError: 'call_count'`, and simulating an `ApiBackend` field drop fails `test_shim_matches_db[get_cost_summary]`.

## One honesty nit

For `group_by="tool"` the payload still emits a bare `total_cost_usd: 0.0`. The CLI deliberately avoids that shape — `cmd_cost.py:135-150` prints "Tool-call spans carry no cost of their own; cost is attributed to the LLM completion that used them. Showing call counts." and renders no dollar total at all, precisely so no misleading `$0.00` appears. Your docstring caveats the row fields but not the total.

This PR is what makes `tool` grouping reachable over MCP, and MCP output is consumed by a model that could reasonably report "$0.00 total spend" from that field. Either omit `total_cost_usd` for tool grouping or add a `note` mirroring the CLI's sentence.

Smaller, no action needed now: `group_by` is still unvalidated in MCP (an unrecognized value silently falls back to `day`) — pre-existing, and #565 allowed either supporting tool grouping or rejecting clearly, so choosing support is fine. And the backend's `tenant` / `feature` / `environment` / `prompt_version` groupings still aren't advertised in the MCP docstring, which is a discoverability gap for another day.

Fix the `_HttpDB` forwarding plus a gate for it and I'll merge this quickly.

be-student:
Addressed the HTTP-mode cost-row gap in `1e7957fc`. The adapter now preserves cache read tokens, cache write tokens, and call count, with a focused HTTP-to-MCP regression. Five cost-summary tests and Ruff checks pass.

anilmurty (approved):
Thanks @be-student — merging. The blocker is properly fixed and I verified it end-to-end rather than by reading the diff.

- **The P1 is gone.** Drove the real `/api/v1/cost` ASGI response through `_http_get` into `_tool_get_cost_summary(_HttpDB(), …)` and compared against the direct-DB path: `HTTP == DIRECT` for both `group_by="model"` and `group_by="tool"`, with all four token fields and `call_count` round-tripping. Against `origin/main`'s `_HttpDB` the same call still raises `AttributeError: 'types.SimpleNamespace' object has no attribute 'cache_tokens'`, so the original defect reproduces and your fix is what closes it.
- **Your test is load-bearing** — reverting the three added lines fails it with exactly that AttributeError.
- The docstring you added for tool rows is genuinely good: it explains the mechanism ("those values belong to the LLM completion span rather than the separate tool-call span") rather than hand-waving, and the calling model is the consumer that needs it.

**On my second ask — the parity gate — I tested it rather than eyeballing it, and the answer is half.** Adding a field to `CostRow` and the REST route while omitting it from `_HttpDB` DOES fail your new test, so the `_HttpDB` half is gated. But omitting it from the MCP *serializer* instead passes 80 tests silently, which is #565 recurring in the other direction. That is not something to hold this PR for — it needs `dataclasses.fields(CostRow)` in both your test and `_proj_cost`, and `_proj_cost` is pre-existing. Filed as #759 with the experiment written up, and you are welcome to take it.

One nit I am not blocking on, left as a note: the payload still emits a bare `total_cost_usd: 0.0` for tool grouping, while `cmd_cost.py:135-150` deliberately renders no dollar total at all for the same grouping and prints the caveat instead. Your docstring covers the primary consumer, so the exposure is small — but a `note` field in the payload would make the disclosure survive outside the tool description.

PR #752, opened 2026-09-06
Fix #586: add full-detail expansion for minor findings

### LINKED ISSUE #586: tj optimize: no way to force full detail on a collapsed "minor" finding, even by naming it explicitly
## Problem

`tj optimize` ranks findings by their share of the analysis window's reclaimable tokens (`tokenjam/cli/cmd_optimize.py`, see the comment block starting around line 444). Any finding whose share falls below `DE_MINIMIS_SHARE` (1%, line 455) is collapsed into a one-line pointer under a "Minor findings" header instead of being rendered in full:

```
  Minor findings (< 1% of window tokens):
     • Cache efficacy — ~0.3% of window tokens. Run tj optimize cache for detail.
```

The problem: that pointer's own advice doesn't work. `_rank_findings(report, requested)` (called around line 819) still applies the same de-minimis collapse even when `requested` is exactly the finding the user asked for by name. Running `tj optimize cache` directly, when `cache`'s share is below 1%, produces the *same* collapsed one-liner telling you to run `tj optimize cache` — which is what you just did. There is no flag anywhere in the command (checked every `@click.option` on `cmd_optimize`, lines ~90-119: `--agent`, `--since`, `--budget`, `--budget-usd`, `--compare`, `--export-config`, `--export-templates`, `--validate`, `--samples`, `--yes`) that forces full rendering of a collapsed finding.

## Expected

A `--expand` flag that disables the de-minimis collapse for the whole run: when passed, `tj optimize` renders full detail for every finding that would otherwise collapse into the "Minor findings" pointer list, for the entire report — not a per-finding toggle. A user who explicitly wants to see the detail behind a small-but-real finding passes `--expand` on the command; there is no separate per-finding form.

## Where to look

- `tokenjam/cli/cmd_optimize.py` — `DE_MINIMIS_SHARE` and the ranking comment block, around lines 444-465
- `tokenjam/cli/cmd_optimize.py` — the `major` / `unranked` / `minor` split and the "Minor findings" render block, around lines 819-947
- `tokenjam/cli/cmd_optimize.py` — the `@click.option` declarations on `cmd_optimize`, around lines 90-121, for where a new flag would be wired in

## Done when

Passing `tj optimize --expand` (run-level; applies to the whole report, e.g. `tj optimize cache --expand`) renders full detail for every finding that would otherwise be collapsed into the "Minor findings" pointer — covered by a test that seeds a below-threshold finding and asserts the flag produces the finding's normal full-detail output instead of the one-line pointer.

be-student:
Closes #586

The de-minimis ranking keeps sub-1% findings visible as one-line pointers, but the advertised detail command applied the same collapse again. Even `tj optimize cache` could tell the user to run `tj optimize cache`, with no route to the underlying evidence.

This adds a run-level `--expand` flag. Passing it selects the detailed report and promotes every quantified below-threshold finding into the normal full-card bucket for that invocation. Default scoreboard, verbose, ranking, and minor-pointer behavior remain unchanged.

Validation:

- the new command-level regression failed before implementation with Click exit code 2
- 58 relevant optimize, scoreboard, renderer, and advertised-command tests pass
- Ruff and mypy pass for `cmd_optimize.py`
- `git diff --check` passes

The broader focused selection also contains `test_optimize_json_stdout_is_byte_clean_under_a_live_spinner`, which fails identically on upstream `main` and this branch because the forced-terminal spinner emits only a newline in this local environment. It is unrelated to the option or renderer path.


anilmurty:
@be-student the `--expand` flag itself is right, minimal, and its test is fully load-bearing — reverting the product change fails it on the Click option, and a partial revert still fails it on the render behaviour. One thing to finish before it lands.

## Blocking: the pointer still sends the user in a circle

#586's Problem is that the minor-findings pointer's own advice doesn't work. That pointer is unchanged, so it still does. I ran the exact string it prints against a seeded corpus with a 0.2%-share downsize finding:

```
$ tj optimize downsize          # ← what the screen tells you to run

  Minor findings (< 1% of window tokens):
     • Downsize — 9% of sessions match, but only ~0.2% of window tokens. Run tj
optimize downsize for detail.
```

Exit 0, same pointer, back where you started. `--expand` works, but it's invisible from the one screen whose whole job is to point at it. The fix is the two `console.print` calls at `tokenjam/cli/cmd_optimize.py:1099` and `:1114` — make them advertise `tj optimize <name> --expand`.

For reference, the expanded output is correct and renders strictly more, with the caveat block intact:

```
  ① Downsize: 9% of sessions match a smaller-model candidate shape
     • 1 of 11 sessions matched structural heuristics
     • Would have cost ~$0.00 on the smaller model vs $0.03 actual (in window)
     • Projected savings if pattern holds: $0.03/mo  (n=1 sessions; too few to bracket)
     ! Candidate-flagging heuristic, not a quality judgment. Review the example
       sessions before changing models.
```

## Blocking: pin the new command

Critical Rule 44 says that when a surface collapses its output, the escape-hatch command it prints becomes load-bearing, and `tests/unit/test_advertised_commands_are_invocable.py` must be extended whenever a screen prints a new one.

Right now that test only parses `_render_scoreboard`. It never renders `_render_report`'s minor-findings block, which is why this circular pointer has been shipping unnoticed — the screen has never been pinned at all. Strictly, your PR as it stands prints no new command so nothing was required; but the fix above *does* print one, so please extend that test to render the minor block and resolve its parsed commands against the real command tree. That closes the gap that let this exist rather than just patching the instance.

## On Greptile's P1

It flagged the same defect and offered two fixes: auto-expand an explicitly-requested finding, **or** advertise `--expand` in the pointer. Take the second. #586's Expected section explicitly rules out a per-finding form and asks for a run-level flag, so auto-expanding on request would contradict the issue you're closing.

## Smaller

- `docs/cli-reference.md:219` — the user-facing "Key flags" list for `tj optimize` didn't get `--expand`. That list is already missing `--validate` and `-v` so it's pre-existing drift, but this is the flag whose entire value is discoverability, so it's worth adding here.
- Right now `--expand` is discoverable only via `tj optimize --help`. Fixing the pointer solves that too.

Verified clean otherwise: no string changed, the full `MODEL_DOWNGRADE_CAVEAT` block survives in the expanded card, no new dollar or token figure is introduced (the flag only re-routes findings between existing render buckets), no hardcoded colours, no existing test weakened. Suite on your head: 5656 passed, with the 5 failures a strict subset of what `origin/main` produces locally — nothing PR-caused.

Fix the pointer and extend the invocability test and this is good to go.

be-student:
Addressed the detail-loop review in `39e17c17`. An explicitly named finding now renders in full even below the de-minimis threshold, and the regression verifies that it no longer repeats the minor-finding pointer. The focused integration test and Ruff checks pass.

be-student:
Follow-up correction in `3ded89ef`: I restored the issue's intended run-level behavior and changed the minor-finding pointers to advertise `tj optimize <name> --expand`. The advertised-command guard now renders the minor block, extracts that command, and validates it against the real Click tree; the CLI reference also lists `--expand`. Fifteen focused tests and Ruff checks pass.

anilmurty (approved):
Thanks @be-student — merging. You fixed both asks and did the harder one properly.

Verified rather than trusting the green checks:

- **The loop is broken.** Ran the exact advertised string on a seeded corpus: `tj optimize downsize --expand` renders the full card, while the pre-fix string still just re-prints the pointer. That contrast is the proof the flag is what breaks it.
- **The Rule 44 pin is load-bearing.** Reverting only the two `console.print` strings, with your test intact, fails with `assert 'tj optimize cache --expand' in ['tj optimize cache for']`. That screen had never been pinned by `test_advertised_commands_are_invocable.py` before, which is why the circular pointer shipped in the first place — you closed the gap, not just the instance.
- **You correctly backed out the auto-expand approach.** `39e17c17` introduced `show_minor_detail = expand or bool(requested)` and `3ded89ef` reverts that hunk exactly. #586's Expected section rules out a per-finding form, so this is the right call.
- Rule 14 intact — `--expand` reuses `_render_finding` unchanged, so every expanded card carries its caveat. Confirmed live for downsize, reuse and script.

**On Greptile's second P1 ("Daemon Ignores Analyzer Scope") — triaging rather than merging past it.** It describes a real leak, and it is **pre-existing on `main`, not introduced here**. `rank_findings` (`core/optimize/rank.py:86`) filters only `downsize` against `requested`, and `GET /api/v1/optimize` records `requested_findings` in the envelope without narrowing `payload["findings"]`. On `origin/main` today, a daemon-up `tj optimize downsize` already renders Reuse, Cache efficacy, Workflow restructure and others that were never requested.

What this PR does widen is the minor half: `--expand` promotes unrelated minor findings from one-line pointers to full cards. The figures are correct, just out of scope — noise rather than wrong numbers. The fix belongs in `rank.py` and the `/optimize` payload, which is a different file and a `main` bug; blocking a 22-line PR on it would be scope creep. Filing it as a follow-up so it does not get lost.

One note for future reference: your suite run will show some `test_quickstart.py` failures under `-n auto`. They are cross-worker pollution from the three known-bad `summarize` tests, not yours — with those deselected the suite is clean.

PR #753, opened 2026-09-09
fix(tests): clear the compute flag the relearn routes actually read

anilmurty:
`tests/integration/test_relearn_api.py` carries an autouse fixture that exists to stop one specific flake — and it has never worked, because it clears a different object than the one the code under test reads. It most recently reddened a contributor's PR (#751) on the 3.11 leg while 3.10 and 3.12 passed, which cost that contributor a round of debugging on a failure that was ours.

## Summary

- Clear every process-global compute flag the routes in this file can reach, not just one of them.
- Pin that the flags are distinct objects, so nobody collapses them back into a single import and silently re-creates the bug.
- Pin that the public accessors report quiescent after the clear — the property the fixture is actually meant to provide.

## Root cause

The fixture cleared `report_store._COMPUTING`. The route under test, `GET /api/v1/relearn/proposals`, reads `relearn_store.is_computing()` at `api/routes/relearn.py:442`, which reads `relearn_store._COMPUTING`. Those are two different `threading.Event` objects:

```python
>>> from tokenjam.core.optimize import report_store, relearn_store
>>> report_store._COMPUTING is relearn_store._COMPUTING
False
```

So the isolation cleared a flag nothing under test consults, and the leak it was written to absorb — an earlier test's `POST /api/v1/relearn/refresh` spawning an unjoined daemon thread that sets the Event — kept reaching the later test. Whether it fired depended on thread scheduling, which is why it presented as a single-matrix-leg flake.

The fixture's own docstring names the failing test by name. It was right about the mechanism and wrong about the object, and nothing could tell the difference: a fixture that clears the wrong thing is indistinguishable from one that works until the race actually fires.

`cost_proposals._COST_COMPUTING` is now cleared too — this file exercises the cost-proposals endpoints ten times and leaks the same way. `report_store` stays in the list because clearing it is harmless and other suites share that module.

## Tests / Verification

The fix is proven by reproduction, not by the suite going green (it was already green). I injected an identical leaking test into a pristine `origin/main` worktree and into this branch:

```
main (old fixture) + leak:   1 failed, 34 passed
                             FAILED test_relearn_proposals_carries_persona_when_never_run
                             AssertionError: assert 'computing' == 'never_run'   (line 193)

this branch      + leak:     37 passed
```

That assertion and line number are byte-identical to what CI reported on #751.

- `tests/integration/test_relearn_api.py` on this branch: 36 passed.
- Full suite (`unit + synthetic + agents + integration`, `-n auto`): 5667 passed, 12 skipped, 2 xfailed. The 5 failures are the known local-only set — three `summarize` tests that need a real `~/.claude` corpus (Critical Rule 31), and two `test_xdist_isolation` tests that require `--dist loadscope`, which I deliberately omitted from that run.
- `ruff check` clean.

## What's NOT in this PR

- **The underlying unjoined-thread leak.** The real defect is that `POST /api/v1/relearn/refresh` spawns a daemon thread the test never joins, so the flag outlives the test that set it. Cleaning up after it in a fixture is a containment, not a cure. Fixing the lifecycle properly is a larger change to the refresh path and belongs on its own.
- **A repo-wide sweep for the same shape.** Three modules each own a private `_COMPUTING`-style Event; there may be other fixtures reaching for the wrong one. Worth a sweep, out of scope here.


anilmurty:
Greptile's P1 was right, and I've fixed it in `7284361` rather than merging past it.

**The gap:** clearing the correct Event was necessary but not sufficient. Each worker sets its flag from *inside* the thread, after building a backend — `trigger_background_recompute._job` calls `backend_factory()` first, and only then does `recompute_now` reach `_COMPUTING.set()`. A worker still opening its backend when the fixture clears sets the flag again during the next test. Same wrong `computing` response, different route to it.

**The fix:** join the outstanding workers before clearing. All three recompute threads are named (`relearn-recompute`, `cost-proposals-recompute`, `optimize-report-scan`), so it's a bounded join over `threading.enumerate()`; a worker that outlives the budget degrades to the old clear-only behaviour rather than hanging the suite.

Greptile's second point — that my regression test called `_clear_compute_flags()` directly instead of letting a real worker cross the boundary — was also correct, and chasing it turned up a worse problem in my own test. **The first version was decorative.** It asserted the flag was clear immediately after the drain, which passed with the drain deleted, because the late worker hadn't reached its `set()` yet. A test that cannot fail for the reason it exists is exactly what I've been sending back to contributors this week, so it was a fair thing to be caught by.

The load-bearing assertion is now `not t.is_alive()` — a clear-only implementation returns with the worker still running, and that's what catches it. Verified by deleting the drain and re-running: `test_quiesce_waits_for_a_worker_that_sets_the_flag_late` fails at the liveness assertion, 1 failed / 36 passed. It pins both halves: the bare clear loses the race, the drain wins.

Full suite under `-n auto --dist loadscope`: 5670 passed, 12 skipped, 2 xfailed. The 3 failures are the known local-only `summarize` set that needs a real `~/.claude` corpus.

Still not the cure — the underlying defect is that the refresh endpoint spawns a thread nothing joins, and this is containment at the test boundary. That stays in 'What's NOT in this PR'.

